PR #10, opened 2019-09-02
highlight: replace libs with more powerful one

### LINKED ISSUE #6: Font in <pre>s is not monospaced
Screenshot 2019-08-30 at 21 20 05

Seems to me there's some issue with the CSS for the `pre` blocks where the monospace font is not properly applied.

vcapretz:
the new lib is based on highlight.js, which gives highlights for more languages
thus we can prettify the error stacktrace too 🎉

also makes snyk happy since this one doesn't contain vulnerability issues 🤷‍♀️

closes #6

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "build": "NODE_ENV=production webpack",
     "build:watch": "NODE_ENV=production webpack --watch",
     "start": "node example.js",
-    "start:dev": "yarn build:watch & yarn start",
+    "start:dev": "concurrently \"yarn build:watch\" \"yarn start\"",
     "start:example": "yarn build && yarn start"
   },
   "dependencies": {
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "build": "NODE_ENV=production webpack",
     "build:watch": "NODE_ENV=production webpack --watch",
     "start": "node example.js",
-    "start:dev": "yarn build:watch && yarn start",
+    "start:dev": "yarn build:watch & yarn start",
```
erikengervall: this'll start the process in the background and might prove hard to kill - is that what you want?
consider `concurrently` lib instead

vcapretz: https://github.com/vcapretz/bull-board/pull/10/commits/3d1a9c2c77e2583d922b1329daf344d71c9e13da

PR #18, opened 2019-09-21
Add 'Clean Jobs' to Delayed Jobs

### LINKED ISSUE #3: Add option to clean jobs
We can start with a button to clean all the jobs in the delayed status, can probably be expanded to all the statuses

alexleroyross:
There is now a button in the Delayed tab that cleans jobs.

It's difficult for me to test that this actually works.  Here's what I've been doing to test that this works:
1) Put the same Clean All button into failed jobs
2) Click it and verify that it moves the failed jobs into Waiting
This works.

Is there a simple way to create a test job that is Delayed, so I can test this more directly?

I've tried calling [moveToDelayed()](https://github.com/OptimalBits/bull/blob/develop/lib/job.js#L300) on newly created jobs, but that seems to move the jobs to Waiting rather than Delayed, so I can't test the Clean All button.  Or, perhaps it moves the jobs to the delayed status for a brief second, but they are immediately promoted to the waiting status for processing.  Either way, I can't test the Clean All button in the Delayed category.  Is there a way to force jobs to stay in the Delayed category?

Despite these difficulties with testing, I modeled this code verbatim after the 'Retry All' option that is available to Failed jobs.  The only difference is that the logic calls queue.clean() instead of job.retry().  Because the Clean All button moves jobs to Waiting when I place it in the Failed category, I believe it would work in the Delayed category.

I understand that this PR isn't ready to merge yet.  Any pointers would be appreciated.  Thank you very much.

closes #3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules/
 static/
 yarn-error.log
+*.rdb
--- a/ui/components/hooks/useStore.js
+++ b/ui/components/hooks/useStore.js
@@ -54,5 +54,17 @@ export default function useStore(basePath) {
       update,
     )
 
-  return { state, retryJob, retryAll, selectedStatuses, setSelectedStatuses }
+  const cleanAll = queueName => () =>
+    fetch(`${basePath}/queues/${queueName}/clean`, { method: 'put' }).then(
+      update,
+    )
+
+  return {
+    state,
+    retryJob,
+    retryAll,
+    cleanAll,
+    selectedStatuses,
+    setSelectedStatuses,
+  }
 }
--- a/index.js
+++ b/index.js
@@ -18,6 +18,7 @@ function UI() {
   router.get('/queues', require('./routes/queues'))
   router.put('/queues/:queueName/retry', require('./routes/retryAll'))
   router.put('/queues/:queueName/:id/retry', require('./routes/retryJob'))
+  router.put('/queues/:queueName/clean', require('./routes/cleanAll'))
   router.get('/', require('./routes/index'))
 
   app.use(bodyParser.json())
--- a/routes/cleanAll.js
+++ b/routes/cleanAll.js
@@ -1,0 +1,23 @@
+module.exports = async function handler(req, res) {
+  try {
+    const { queueName } = req.params
+    const { queues } = req.app.locals
+
+    const GRACE_TIME_MS = 5000
+
+    const queue = queues[queueName]
+    if (!queue) {
+      return res.status(404).send({ error: 'queue not found' })
+    }
+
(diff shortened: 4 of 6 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules/
 static/
-yarn-error.log
\ No newline at end of file
+yarn-error.log
+*.rdb
```
alexleroyross: Git kept trying to commit a dump.rdb file, so I added this to .gitignore.  I can remove it if it is a problem.

ui/components/hooks/useStore.js
```diff
@@ -54,5 +54,17 @@ export default function useStore(basePath) {
...
+    )
+
+  return {
+    state,
+    retryJob,
+    retryAll,
+    cleanAll,
+    selectedStatuses,
+    setSelectedStatuses,
+  }
```
alexleroyross: My code formatter changed this return statement from a single line to multiple lines.  I can change it back if this is an issue.

vcapretz: it was probably because a new item was added, so it's fine :D

.gitignore
vcapretz: no problems!

vcapretz:
hey @alexleroyross, thank's for the PR! 

I do understand testing wise we could do much better in the lib, it's something that needs to be done.

What I did to test a delayed queue was to change the `.add` method under `example.js` and add a delayed option as so: 
```js
example.add({ title: req.query.title }, { delay: 100000 })
``` 

this way the jobs are forced to go to the delayed status as soon as you add them, and it works as expected 😄 

so I would say we are ready to merge this one, what do you think?

alexleroyross:
> hey @alexleroyross, thank's for the PR!
> 
> I do understand testing wise we could do much better in the lib, it's something that needs to be done.
> 
> What I did to test a delayed queue was to change the `.add` method under `example.js` and add a delayed option as so:
> 
> ```js
> example.add({ title: req.query.title }, { delay: 100000 })
> ```
> 
> this way the jobs are forced to go to the delayed status as soon as you add them, and it works as expected 😄
> 
> so I would say we are ready to merge this one, what do you think?

Oh, I see!  I did not realize that supplying that 'delay' parameter was the same thing as moving a Queue to the 'delayed' status.  Thank you!

I tested it on my end, and it worked for me as well.  Awesome!  As far as I can tell, it looks like it is ready to merge, but it's up to you!  Thank you so much for your help.  I'll leave it up to you to close this and merge it.

vcapretz:
Awesome work @alexleroyross! I'll release a new version in npm now with this addition 🥇

PR #20, opened 2019-09-23
job-progress: fix percentage value when gt 100

### LINKED ISSUE #17: Request: Job progress count instead of percentage
Hello.

First of all, thank you for making a nice and clean interface for bull. I use it in my project that syncs data between databases. I do the work using batches, so I never know the max value of things that need processing in order to calculate the percentage. So if i do `job.progress(item.count)` where item.count is for example 4235 items, the frontend interface will go crazy because the green progressbar will be `style="width: 4235%"`.

What im proposing is a simple check, something along the lines of:
`if (job.progress > 100) // Don't render class="progress-bar", style="width: 4235%" and the % sign`

```
progress: ({ job }) => {
    switch (type(job.progress)) {
      case 'Object':
        return (
          <Highlight className="json">
            {JSON.stringify(job.progress, null, 2)}
          </Highlight>
        )
      case 'Number':
       if (job.progress > 100)  {
            return (
          <div className="progress-wrapper"> {job.progress}</div>
        )
        }
        return (
          <div className="progress-wrapper">
            <div
              className="progress-bar"
              style={{
                width: `${job.progress}%`,
              }}
            >
              {job.progress}
              %&nbsp;
            </div>
          </div>
        )
      default:
        return '--'
    }
  },
```

Thanks in advance

vcapretz:
jobs might have progress reports with values that are bigger than 100
this at least prevents the progress bar to break and instead show only the number

fixes #17

### DIFF
--- a/ui/components/Queue.js
+++ b/ui/components/Queue.js
@@ -166,6 +166,10 @@ const fieldComponents = {
           </Highlight>
         )
       case 'Number':
+        if (job.progress > 100) {
+          return <div className="progress-wrapper">{job.progress}</div>
+        }
+
         return (
           <div className="progress-wrapper">
             <div


ui/components/Queue.js
```diff
@@ -166,6 +166,10 @@ const fieldComponents = {
           </Highlight>
         )
       case 'Number':
+        if (job.progress > 100) {
+          return <div className="progress-wrapper">{job.progress}</div>
```
erikengervall: Instead of this early return, wouldn't it make more sense to limit the [width style](https://github.com/vcapretz/bull-board/blob/a66144a2c0a5b631233117b4eb2928f034c9c8d4/ui/components/Queue.js#L178) to 100 like: `width: ${job.progress > 100 ? 100 : job.progress}%`?

vcapretz: the catch is that we also don't render the "%" (percentage sign) in the text itself, so we would need to check the style AND the text 

the goal is that the number would not indicate percentage after is higher than 100, because if that happens it may indicate other use cases for reporting progress, as stated in the issue related to this PR

erikengervall: Hm, so an arbitrary progress that happens to be below 100 would automatically be considered a percentage? I’d recommend trying to structure this a bit more, or provide the user with options for what they’d like to have displayed.

erikengervall: I don’t mind this change as a temporary fix, but if so I’d suggest creating an issue for providing users with more options :)

vcapretz: it depends on the use case really, this is more on Bull than here on bull-board, since they provide a `.progress` method, users can call it with whatever parameters they want (as long as Bull allows), they recommend using it for percentages and specify the value should be between 0 and 100, but nothing blocks users from using this for other stuff 

https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#jobprogress

erikengervall: Ok!

PR #28, opened 2019-10-11
Add "add and setter" functions

sestolk:
Add ability to add already defined queues (using Bull)

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,6 +29,16 @@ function UI() {
 
 module.exports = {
   UI: UI(),
+  setQueues: (queues) => {
+    queues.forEach((item) => {
+      this.addQueue(item.queue);
+    })
+  },
+  addQueue: (queue) => {
+    queues[queue.name] = queue
+
+    return this;
+  },
   createQueues: redis => {
     return {
       add: (name, opts) => {
--- a/index.js
+++ b/index.js
@@ -29,15 +29,16 @@ function UI() {
 
 module.exports = {
   UI: UI(),
-  setQueues: (queues) => {
-    queues.forEach((item) => {
-      this.addQueue(item.queue);
-    })
-  },
-  addQueue: (queue) => {
-    queues[queue.name] = queue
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -29,6 +29,16 @@ function UI() {
 
 module.exports = {
   UI: UI(),
+  setQueues: (queues) => {
```
vcapretz: should we keep only this one method for adding queues already set in Bull?  
It could accept only arrays with one or more values and return the queues array with all the arrays set in bull-board.

What do you think? This is more of a suggestion, I wouldn't mind having the two methods but maybe having only one keeps the api cleaner

vcapretz:
amazing PR @sestolk! That is definitely gonna help a lot more people using the libraries without introducing a huge barrier! 

Thanks!

index.js
sestolk: Well I think it would be nice to be able to add extra queues when needed (like on-the-fly). But I can certainly live with only the setQueues that would check if `queues` is a single queue or an array and therefore have the same functionality as the two separate ones.

sestolk:
> 
> 
> amazing PR @sestolk! That is definitely gonna help a lot more people using the libraries without introducing a huge barrier!
> 
> Thanks!

Thanks! I was trying to get this to work in a project at work, but ran into this issue so I thought, why not make a PR for it. I am not really looking forward to maintain it myself, as it is just for a client of ours, but it was only missing this functionality (for now). It would be a big help if it gets accepted!

index.js
vcapretz: awesome, can you make these changes then?

so we would keep only the `setQueues` function, which would accept either one or more queues to be added and return all the queues we have 😄 

I can merge and release after you make these changes! Thanks again

sestolk:
Done, the merge request has been updated

vcapretz:
@sestolk published in npm v0.4.0 😄

PR #34, opened 2019-10-29
feat: use `proxyUrl` if provided

Eywek:
We want to be able to configure a different path from the one
mounted on Express router.
For example, if we use a reverse proxy that rewrite our path we
want to be able to load the UI.


vcapretz:
hey @Eywek thanks for the PR! I'll merge and deploy it later today and let you know 😃

Eywek:
hi 😁

vcapretz:
@Eywek just released in `0.5.0` 😄 

https://github.com/vcapretz/bull-board/releases/tag/v0.5.0
should be available in `npm` already, let me know how it works for you!

Issue #24, opened 2019-10-04
Add a Button to clear failed jobs


edy:
There are alredy two retry buttons. It would be nice to have some buttons to clean up all failed jobs and single jobs.

vcapretz:
hey @edy thank's for the issue! 

we already added a clean option for the delayed queue to start trying it out, I think it can be added to the failed one as well.
My only concern here would be how the UI would look like with a lot of buttons per job, so I might try it out and open a PR!

(feel free to also send PR's here if you feel like it 👍)

vcapretz:
fixed in #39

PR #39, opened 2019-11-17
Clean all for failed job

ex7r3me:
Now there's a 'clean all' button for failed tasks.
We may want to generalize clean functions by sending params in request body or even change the HTTP method to `DELETE`  with the resource ('delayed', 'failed') in the URL.

I changed `clean` names to `cleanDelayed` to prevent confusion, but I know this PR might not be ready to merge. I can change the code to your suggestion.

We can also add a button to clean a single job with maybe with a small trash icon or a more options button, what do you think? 

related to #24


vcapretz:
hey @ex7r3me I like the idea of allowing to clean the failed queue as well, thank you for this PR! 😄 

I would just keep one implementation for the `routes` and use a URL param for getting which queue is it, so in the UI we would call it like 

```
fetch(`${basePath}/queues/${queueName}/clean/failed`, { method: 'put' }).then(...
fetch(`${basePath}/queues/${queueName}/clean/delayed`, { method: 'put' }).then(...
fetch(`${basePath}/queues/${queueName}/clean/whateverHere`, { method: 'put' }).then(...
```

what do you think?

ex7r3me:
> hey @ex7r3me I like the idea of allowing to clean the failed queue as well, thank you for this PR! 
> 
> I would just keep one implementation for the `routes` and use a URL param for getting which queue is it, so in the UI we would call it like
> 
> ```
> fetch(`${basePath}/queues/${queueName}/clean/failed`, { method: 'put' }).then(...
> fetch(`${basePath}/queues/${queueName}/clean/delayed`, { method: 'put' }).then(...
> fetch(`${basePath}/queues/${queueName}/clean/whateverHere`, { method: 'put' }).then(...
> ```
> 
> what do you think?

sure! it's much nicer like this. I've just updated the PR with new routes, do we need any other changes?

vcapretz:
hey @ex7r3me sorry for the very long time to answer it 😅 I totally forgot this PR, my bad 

it looks very good now, thank you very much for this PR. I'm merging it and releasing it

PR #44, opened 2019-12-11
allow adding a route prefix to UI when using koa

ltjin:


vcapretz:
thanks for the PR @ltjin, and sorry for the long time to review it 😅

I was a bit worried about the breaking change in the UI export, which now should be called inside the application instead of bull-board calling it, but #49 is also adding breaking changes and we can have both in a release 😄 🎉 

also, appreciate a lot you keeping the docs updated!

vcapretz:
do you think it's worth creating an `example-koa.js` file to create a simple app using koa the way we do with express?

vcapretz:
hey, @ltjin I just merged a big PR which now causes some conflicts in this one, can you maybe solve them so I can merge? 

thank you again for the work!

ltjin:
I'm back to work now.
I found out that the 1.0.0-alpha.0 version using TS has been released and I will re-work on the new version.

gabrielstuff:
hello @ltjin did you had time to work on it ? thanks !

jvcjunior:
Any news on this one guys?

felixmosh:
bull-board `v3.1.0` has an official koa server adapter

PR #49, opened 2020-01-08
Add support for bullmq (bull v4) and some small improvements 

### LINKED ISSUE #32: It doesnt show the job name
Board shows the data, but not the job name, can we get it to show this?

### LINKED ISSUE #38: support for Bull 4
are you guys planning on supporting BullMQ 4 ?

Embraser01:
Hi,

I added support for bull v4 (#38). 

- Now, bull and bullmq are both optional peer dependencies and must be added by the user.
- The bull version is displayed on the dashboard
- For now, the `createQueues` function does not support bull 4

Unrelated (I can move it to another PR if you want), I added a button to toggle data instead of displaying it directly. In case of big data blob, the UI was freezing

closes #32 
closes #38

### DIFF
--- a/example.js
+++ b/example.js
@@ -1,25 +1,25 @@
-const { createQueues, setQueues, UI } = require('./')
+const { setQueues, UI } = require('./')
+const { Queue3 } = require('bullmq/dist/classes/compat')
+const Queue = require('bull')
 const app = require('express')()
 
 const sleep = t => new Promise(resolve => setTimeout(resolve, t * 1000))
 
 const redisOptions = {
-  redis: {
-    port: 6379,
-    host: 'localhost',
-    password: '',
-    tls: false,
-  },
+  port: 6379,
+  host: 'localhost',
+  password: '',
+  tls: false,
 }
 
+const createQueue = name => new Queue(name, { redis: redisOptions })
+const createQueue4 = name => new Queue3(name, { connection: redisOptions })
+
 const run = () => {
-  setQueues([/* Already defined (bull) queues */]);
-  // Or a single bull queue
-  setQueues(/* Already defined bull queue */);
+  const example = createQueue('ExampleBull')
+  const exampleMQ = createQueue4('ExampleBullMQ')
 
-  const queues = createQueues(redisOptions)
-
-  const example = queues.add('example')
+  setQueues([example, exampleMQ])
 
   example.process(async job => {
     for (let i = 0; i <= 100; i++) {
@@ -29,13 +29,22 @@ const run = () => {
     }
   })
 
+  exampleMQ.process(async job => {
+    for (let i = 0; i <= 100; i++) {
+      await sleep(Math.random())
+      await job.updateProgress(i)
+      if (Math.random() * 200 < 1) throw new Error(`Random error ${i}`)
+    }
+  })
+
   app.use('/add', (req, res) => {
     example.add({ title: req.query.title })
+    exampleMQ.add('Add', { title: req.query.title })
     res.json({ ok: true })
   })
 
   app.use('/ui', UI)
-  app.listen(3000, () => {
+  app.listen(3002, () => {
     console.log('Running on 3000...')
     console.log('For the UI, open http://localhost:3000/ui')
     console.log('Make sure Redis is running on port 6379 by default')
--- a/example.js
+++ b/example.js
@@ -1,6 +1,6 @@
 const { setQueues, UI } = require('./')
-const { Queue3 } = require('bullmq/dist/classes/compat')
-const Queue = require('bull')
+const { Queue3: QueueMQ } = require('bullmq/dist/classes/compat')
+const Queue3 = require('bull')
 const app = require('express')()
 
 const sleep = t => new Promise(resolve => setTimeout(resolve, t * 1000))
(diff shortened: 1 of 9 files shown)


index.js
```diff
@@ -29,18 +60,23 @@ function UI() {
...
-    bullQueues.forEach((item) => {
-      queues[item.name] = item;
+    bullQueues.forEach(item => {
+      queues[item.name] = item
     })
 
     return queues
   },
   createQueues: redis => {
+    if (isBullMQ) {
```
vcapretz: I've been actually considering removing the option to `createQueues` and keep only `setQueues`, what do you think? 

this way we can keep the lib easier to maintain! 👍

vcapretz:
that is amazing @Embraser01! thank you very much for this PR

I added only one minor comment, but other than that I think it's fine!

But I guess this would be enough for a 1.0 release, and for that I want to improve some other things in the lib, so I'll probably merge soon and hold the release to include more stuff

index.js
Embraser01: 👍 Make sense, I'll remove it

Embraser01: Just realize that it was the only place where we needed to `require('bull')`. It means that we can remove bull and bullmq from peerDependencies, right?

ericcarboni:
Unsolicited suggestion, but could you add the v4 docs (https://docs.bullmq.io) to the README in this PR? Might be helpful for new users who are starting with v4

index.js
vcapretz: yes! exactly, this way we don't need to control the lib version or anything, leaving it only up to the users and making sure we support the correct versions 😄

vcapretz:
no problems @ericcarboni! 

@Embraser01 can you add it please? at least some basic stuff, I'll make sure the docs are improved as much as possible before bumping to v1.0.0

index.js
Embraser01: After some testing, I saw that some functions have changed signature between Bull3 and BullMQ...

I can detect the version of a queue by checking (for example) if the `drain` method exists (BullMQ) or not (Bull3). This way, we can switch if needed...

For example, `queue.clean` has not the same signature:
```js
// With BullMQ
await queue.clean(GRACE_TIME_MS, LIMIT, queueStatus)
// With Bull3
await queue.clean(GRACE_TIME_MS, queueStatus)
```

What do you think?

ui/components/Queue.js
```diff
@@ -290,7 +295,10 @@ export default function Queue({
 }) {
   return (
     <section>
-      <h3>{queue.name}</h3>
+      <h3>
+        {queue.name}
+        {queue.version === 4 && <small> bullmq</small>}
```
Embraser01: Don't really know what to put there 🤔

index.js
vcapretz: yeah, that's tricky 😅 
I would say we can have some check when constructing the lib, and have that variable in the locals, I guess it will be a consequence to have some conditions in some places

ui/components/Queue.js
vcapretz: I'm fine with your current suggestion 👍

example.js
```diff
@@ -29,13 +29,22 @@ const run = () => {
...
+
   app.use('/add', (req, res) => {
     example.add({ title: req.query.title })
+    exampleMQ.add('Add', { title: req.query.title })
     res.json({ ok: true })
   })
 
   app.use('/ui', UI)
-  app.listen(3000, () => {
+  app.listen(3002, () => {
```
Embraser01:
```suggestion
  app.listen(3000, () => {
```

index.js
Embraser01: I added a map to store the version for each queue (this way a system can have both versions simultaneously). The function that check the bull version is [here](https://github.com/vcapretz/bull-board/pull/49/files#diff-168726dbe96b3ce427e7fedce31bb0bcR39)

example.js
```diff
@@ -1,25 +1,25 @@
...
+  password: '',
+  tls: false,
 }
 
-const run = () => {
-  setQueues([/* Already defined (bull) queues */]);
-  // Or a single bull queue
-  setQueues(/* Already defined bull queue */);
+const createQueue = name => new Queue(name, { redis: redisOptions })
+const createQueue4 = name => new Queue3(name, { connection: redisOptions })
```
vcapretz: this naming is causing me some confusion 😅 queue3 is the version 3 coming from bullmq, which is version 4?
maybe we can use names such as Queue3 and QueueMQ? it's a bit odd but it aligns with the lib name, what are your thoughts?

example.js
```diff
@@ -1,25 +1,25 @@
...
 
-  const queues = createQueues(redisOptions)
+const run = () => {
+  const example = createQueue('ExampleBull')
+  const exampleMQ = createQueue4('ExampleBullMQ')
```
vcapretz: like here for instance, it would be more clear with we had `exampleMQ = createQueueMQ`

vcapretz:
I think this PR is getting in great shape to be merged!

example.js
Embraser01: Yep, make sense! I will update it now

ui/components/Queue.js
```diff
@@ -154,6 +170,9 @@ const fieldComponents = {
       </div>
     )
   },
+  name: ({ job }) => {
+    return job.name
+  },
```
Embraser01: Note that I added a new column for the name.
With BullMQ, it is mandatory to set a name for the job...

Don't know if I should add it to Bull3 or not @vcapretz ?

vcapretz: yeah, if it works with Bull3 it's fine! 😄

Embraser01: Ok, I just handled the case in Bull3 where there was no name so that it doesn't show `__default__` but just `--`

vcapretz:
thanks for the awesome work @Embraser01! 

I have a `revamp` branch where I'm doing a lot of changes to be able to land a 1.0.0, this will be included there!

seawatts:
Did this get published to NPM?

vcapretz:
No, not yet 😅 I’m preparing a 1.0 launch, I’ll probably release a beta version so if you would like to try it out please follow the pr #51!

PR #50, opened 2020-01-14
feat: distance time for delayed jobs

ex7r3me:
I've added a time format for delayed jobs. so it'll show a more human-readable time format instead of the remaining millisecond. for example, it shows `about 1 hour` instead of `3600000`.

a little technical choice here is that I used `formatDistance` instead of `formatDistanceToNow` as it was already imported in the `Queue.js` file. we can also use `formatDistanceStrict` which would give more exact timing (for example `43 minutes`). which one do you think users would be happier with?

### DIFF
--- a/ui/components/Queue.js
+++ b/ui/components/Queue.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import { getYear, format, isToday, formatDistance } from 'date-fns'
+import { getYear, format, isToday, formatDistance, formatDistanceStrict } from 'date-fns'
 import { type } from 'ramda'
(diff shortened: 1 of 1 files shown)


vcapretz:
hey, @ex7r3me thank you for the PR! really appreciate it 

I think going with `formatDistanceStrict` might be preferable here, so users can have more insights on the delay time. 

After you update with this change I can merge the PR! 😄

ex7r3me:
@vcapretz nice! 
I tried it and yes it feels better, I've updated the PR with the new function

vcapretz:
thank you! just merged, gonna wait to release it

PR #64, opened 2020-01-25
Remove TODOs 

Yurickh:
AKA do something about the types

There are several `REVIEW` comments scattered around the changes, these are request for comments, since I'm new to the code base, it would be great to have some feedback on them :)~

Changes can be summarised as:

## Enable development environment
I noticed there were some misalignments on how the default setup is described in the README and the connections between the scripts. 
Also, the development build was breaking due to the fact that `job.progress` is not a function.

## Extraction of top-level types
There are some types that are prominent across the code base, so I extracted them to the top level `@types` file so we can have a single source of truth for how the data types are structured.
While I was at it, I created a `@types/api.ts` file that is supposed to contain the interfaces for the service endpoints (in the end it has only the get request) so they can be consumed client side as well.

## Removal of `TODO: typings`
By adding proper typings I managed to remove all `TODO`s across the codebase. Please check throughly if the types are correct and aligned with what you expect of them to receive.
In the process I noticed some misalignments (like some code expecting `job.delay` to have the delay amount, when in fact, it contains a boolean).


Yurickh:
As a sidenote, the branch name is due to the fact that I've opened it to fix an actual issue, but couldn't reproduce, so I went ahead and focused in the `TODO`s scattered around.

erikengervall:
> AKA do something about the types
> 
> There are several `REVIEW` comments scattered around the changes, these are request for comments, since I'm new to the code base, it would be great to have some feedback on them :)~
> 
> Changes can be summarised as:
> 
> ## Enable development environment
> I noticed there were some misalignments on how the default setup is described in the README and the connections between the scripts.
> Also, the development build was breaking due to the fact that `job.progress` is not a function.
> 
> ## Extraction of top-level types
> There are some types that are prominent across the code base, so I extracted them to the top level `@types` file so we can have a single source of truth for how the data types are structured.
> While I was at it, I created a `@types/api.ts` file that is supposed to contain the interfaces for the service endpoints (in the end it has only the get request) so they can be consumed client side as well.
> 
> ## Removal of `TODO: typings`
> By adding proper typings I managed to remove all `TODO`s across the codebase. Please check throughly if the types are correct and aligned with what you expect of them to receive.
> In the process I noticed some misalignments (like some code expecting `job.delay` to have the delay amount, when in fact, it contains a boolean).

I'll take a closer look the coming days, but from what I've seen so far the type improvements look awesome 👍

vcapretz:
Thank you very much @Yurickh! Amazing contribution, I'll also be reviewing it closely throughout the week!  ❤️

src/declarations.d.ts
```diff
@@ -1 +1,8 @@
-declare module 'redis-info'
+declare module 'redis-info' {
+  export const parse: (
+    rawRedisInfo: string,
+  ) => {
+    // This is not technically true, as metrics can also come as objects
```
erikengervall: it's something 🙆‍♂

src/routes/cleanAll.ts
```diff
@@ -1,9 +1,15 @@
 import { RequestHandler } from 'express'
 import { Queue as QueueMq } from 'bullmq'
 
-import { BullBoardQueues } from '../@types'
+import { BullBoardQueues } from '../@types/app'
 
-export const cleanAll: RequestHandler = async (req, res) => {
+type RequestParams = {
+  queueName: string
+  // REVIEW: please confirm we can't receive 'paused' nor 'latest' here
```
erikengervall: cc @vcapretz

src/routes/queues.ts
```diff
@@ -3,40 +3,35 @@ import { RequestHandler, Request } from 'express'
...
-  total_system_memory?: string
-  redis_version?: string
-  used_memory?: string
-  mem_fragmentation_ratio?: string
-  connected_clients?: string
-  blocked_clients?: string
-}
+import * as api from '../@types/api'
+import * as app from '../@types/app'
+// REVIEW: this is enough of an argument to move these constants to the top level, WDYT?
```
erikengervall: Sounds reasonable

As long as it doesn't introduce more layers in the compiled dist

src/ui/components/Queue.tsx
```diff
@@ -1,3 +1,6 @@
+// REVIEW: I feel like this file has become quite chockfull of responsabilities.
```
erikengervall: Very much agree.

I'd love to see this file modulized and properly unit tested, considering the wide responsibilities it has

src/ui/components/Queue.tsx
```diff
@@ -25,23 +28,39 @@ const formatDate = (ts: any) => {
...
-const Timestamp = ({ ts, prev }: { ts: any; prev?: any }) => {
+const Timestamp = ({
+  ts,
+  prev,
+}: {
+  ts: TimeStamp | null
+  prev?: TimeStamp | null
+}) => {
+  if (ts === null) {
+    // REVIEW: what should be the timestamp for null?
```
erikengervall: Unsure. Open for suggestions 🙂

src/ui/components/Queue.tsx
```diff
@@ -133,19 +154,19 @@ const fieldComponents = {
...
 
-  // FIXME: typings
-  attempts: ({ job }: { job: any }) => job.attempts,
+  attempts: ({ job }) => <>{job.attempts}</>,
 
-  // FIXME: typings
-  delay: ({ job }: { job: any }) =>
-    formatDistanceStrict(job.timestamp + job.delay, Date.now()),
+  delay: ({ job }) => (
+    // REVIEW: What should we do if timestamp comes as null?
```
erikengervall: Unsure. Open for suggestions 🙂

src/ui/components/Queue.tsx
```diff
@@ -255,7 +277,22 @@ const QueueActions = (props: {
...
+  selectStatus: (statuses: Record<string, Status>) => void
+  cleanAllDelayed: () => Promise<void>
+  cleanAllFailed: () => Promise<void>
+  retryAll: () => Promise<void>
+  retryJob: (job: AppJob) => () => Promise<void>
+}
+
+// REVIEW: this is in place so we can manipulate the return type
+//  of Object.keys which is string[] by default
+// We need to extend so babel doesn't think it's JSX
```
erikengervall: Looks good. Could keep either the entire comment (except for `REVIEW`) or just line 292 🙂

src/ui/components/RedisStats.tsx
```diff
@@ -40,10 +32,12 @@ const getMemoryUsage = (
       ).toFixed(2)}%`
     : formatBytes(parseInt(used_memory, 10))
 
-export const RedisStats = ({ stats }: { stats: ValidMetrics }) => {
+export const RedisStats = ({ stats }: { stats: Partial<ValidMetrics> }) => {
   const {
     redis_version,
-    used_memory,
+    // REVIEW: the API gives no guarantee these fields will be there.
+    //  Is it fine to use '0' as the fallback here?
```
erikengervall: Yeah, that's probably a good idea.

Ideally the API would ensure that a default is always returned, but that's a task for the future 🙂

src/ui/components/constants.ts
```diff
@@ -8,17 +8,25 @@ export const STATUSES = {
...
+  | 'data'
+  | 'id'
+  | 'name'
+  | 'opts'
+  | 'progress'
+  | 'timestamps'
+  | 'delay'
+  | 'failedReason'
+  | 'retry'
+  // REVIEW: this one is in none of the statuses, but we have a handler for it?
```
erikengervall: Unclear. Open for suggestions on what to do 🙂

src/ui/components/Queue.tsx
Yurickh: I think it's reasonable to tackle this on a follow-up PR.

src/ui/components/Queue.tsx
Yurickh: Yea, 292 is supposed to be a "permanent" comment here, hence the lack of indentation :p~

src/ui/components/constants.ts
Yurickh: I'm not sure if I'm missing anything in this one. As far as I understand, we handle the Fields from their mapping from `Status` inside `FIELDS`, but no `Status` is mapping to the `finish` field.
As far as I understand, the line 128 in `src/ui/components/Queue.tsx` is pretty much dead code. Or is it there for the eventful occasion where we'll want to add support to this field? 🤔

src/declarations.d.ts
Yurickh: I left this as it is because that's how we treat the return for this method currently. My fear of adding this type as it is, is that people will believe in it and treat object types wrongly. 😕 
I didn't want to make major changes to the codebase in this PR, rather just make it compatible with the typings.
Maybe it's a work for a follow-up? :D~

src/routes/queues.ts
Yurickh: WDYM? Is that something I have to worry about?
If compiled size is a focus point, it would be great to add something like [bundlesize](https://github.com/siddharthkp/bundlesize) to the CI tests so we can track it properly in PRs :)~

src/ui/components/Queue.tsx
Yurickh: I think it makes sense to understand when it comes as null and provide a matching message, or skip rendering altogether.

src/declarations.d.ts
vcapretz: yes! agreed, can you create a separate issue please @Yurickh?

src/routes/cleanAll.ts
vcapretz: nope, this is currently not implemented, you can only clean delayed and failed queues for now 😄

src/ui/components/Queue.tsx
vcapretz: definitely! let's not address this here!

src/ui/components/Queue.tsx
vcapretz: no delay means it will try to process the job as soon as possible, can be 0 here

src/ui/components/RedisStats.tsx
vcapretz: used memory = 0 can mean different stuff than "we couldn't retrieve it", maybe that should be more clear, at least a `-`

src/ui/components/constants.ts
vcapretz: no plans, for now, kill it

src/ui/components/RedisStats.tsx
Yurickh: I can change the code that shows up this number in the screen rather than adding the default case

Yurickh:
Ok, so based in the feedback I received from @erikengervall and @vcapretz I'm missing the following tasks:
- [ ] Kill the `finish` handler in the Queue component
- [ ] Remove `'0'` default use cases and update the handlers that rely it will be populated to handle the empty case as `-`
- [ ] Open an issue that describes the `constant` file uplifting
- [ ] Open an issue that describes adding the actual types for `redis-info` (maybe adding them to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped)? 🤯) and handling them properly in the codebase
- [ ] Remove the remaining `REVIEW` comments

I'll see these fixed as the week goes by :)~
Thanks a lot for the feedback folks 🙇

src/routes/queues.ts
erikengervall: I mean that if a TS project tries to resolve files that are outside of its root, it may add additional levels of folders in the compiled dist. Thus it may have an impact on paths within the dist (since we build both a client app and a server app and they rely on each others positions within dist) :)

Yurickh: Gotcha! I'll make sure to check that

src/ui/components/Queue.tsx
erikengervall: Ok!

erikengervall:
Merged master into this fella to get https://github.com/vcapretz/bull-board/pull/69 - it'll enable the pipeline for PRs from forks 🎉

Yurickh:
That's super cool! Thanks 🤗

src/routes/queues.ts
```diff
@@ -45,22 +39,26 @@ const getStats = async ({ queue }: BullBoardQueue): Promise<ValidMetrics> => {
...
+  const jobProps = job.toJSON()
+
+  return {
+    id: jobProps.id,
+    timestamp: jobProps.timestamp,
+    processedOn: jobProps.processedOn,
+    finishedOn: jobProps.finishedOn,
+    progress: jobProps.progress,
+    attempts: jobProps.attemptsMade,
+    delay: job.opts.delay,
```
Yurickh: This is a change from previous behaviour!
`delay` would be a boolean before, but would be used as a number in the components, so I'm changing it to use the actual delay instead of doing `+1` or `+0` depending of whether this is true or false 😅

src/declarations.d.ts
Yurickh: Will create an issue once https://github.com/DefinitelyTyped/DefinitelyTyped/pull/42186 is merged. If you could take an extra look there as well to guarantee I'm not messing it up, that would be extra good! 🙏

package.json
```diff
@@ -35,9 +35,9 @@
     "build:ui:watch": "webpack --watch",
     "build:routes": "yarn tsc",
     "build:clean": "rm -rf ./static ./dist",
-    "start:dev:docker": "docker-compose up",
+    "start:dev:docker": "docker-compose up -d",
```
erikengervall: ❤️

src/ui/components/App.tsx
```diff
@@ -32,13 +24,13 @@ export const App = ({ basePath }: { basePath: string }) => {
           'Loading...'
         ) : (
           <>
-            {state.data && state.data.stats ? (
+            {state.data?.stats ? (
```
erikengervall: 🤤

src/ui/components/App.tsx
```diff
@@ -32,13 +24,13 @@ export const App = ({ basePath }: { basePath: string }) => {
...
               <RedisStats stats={state.data.stats} />
             ) : (
               <>No stats to display </>
             )}
 
-            {state.data.queues.map((queue: Queueue) => (
+            {state.data?.queues.map(queue => (
```
erikengervall: Haha can't believe I left that `Queueue` in there 😅 Nice that you caught it 🎉

vcapretz:
great work! thank you very much @Yurickh

Yurickh:
Thanks a lot for all the feedback and support guys! 🙏 
I'll be creating the issues for uplifting constants and breaking down the Queue.ts file shortly in the future :)~

Issue #45, opened 2019-12-13
Standalone version


JonDum:
I was really excited about this project when I saw it, but I had trouble getting it to work. We have Bull queues running in production in a Docker Swarm with dozens of Fastify worker containers and several different redis endpoints. I tried using `fastify.use()` but it's not compatible I don't think. 

But this kinda sucks anyways — we would have to embed the bull-board and all its dependencies (express, react, react-dom, ramda, etc) directly into our server images. I was hoping this would be more of a standalone server that you can point to a redis process and give it a prefix and it'll find your queues for you (via `KEYS bull:*` every so often). That would mean just adding one more container instead of embedding an instance of it to every worker image.

I think you already have the basis of it here, an UI + Express server, what do you think of modifying it to work without any required integration into your node app that is running Bull?

JonDum:
Ok, so this was actually surprisingly easy.

Check out

https://github.com/JonDum/bull-board/blob/master/standalone.js

JonDum:
Now there's a docker image:

`docker run --rm -p 3000 jondum/bull-board:latest`

vcapretz:
hey @JonDum! that's really cool! 

I think this is a great addition to the project, would you mind sending a PR so we can add it here? or do you plan on maintaining your own version of the project? 

also, how would the process of updating the docker image work? 

thank you for taking the time to look into it and proposing a suggestion

tombh:
PR submitted #66

marabesi:
is the docker image from @JonDum the official one atm? Just wanted to run as a stand alone with docker and I found out that #66 has a few merge conflicts.

tombh:
Sorry, I just haven't been keeping up with master because I guess the maintainer just doesn't have the time to look into this branch

marabesi:
Hey @JonDum is there something that we can do to get @tombh work merged?

h3ku:
Is anyone still working on this?

PR #66, opened 2020-01-31
Standalone Docker image

tombh:
This builds on @JonDum's fork (mentioned in #45) to create a Docker image that automatically adds all queues prefixed with 'bull' or whatever value is specified in the `BULL_PREFIX` ENV var.

I've included it in the README, but I'll repeat the command here with my username as it's a nice oneliner to quickly try out and see what this PR is introducing: `docker run --net=host --rm tombh/bull-board`

I've also taken the liberty of including an update to this repo's Github Actions. It doesn't currently work as it needs someone (I'm assuming @vcapretz) to create a Docker Hub account and create a registry token. Once signed up that can be done here: https://hub.docker.com/settings/security

### DIFF
--- a/standalone.js
+++ b/standalone.js
@@ -1,0 +1,38 @@
+const { createQueues, setQueues, UI } = require('./')
+const Queue = require('bull')
+const redis = require('redis')
+const app = require('express')()
+
+const redisOptions = {
+  redis: {
+    port: process.env.REDIS_PORT || 6379,
+    host: process.env.REDIS_URL || 'localhost',
+    password: process.env.REDIS_PASSWORD || '',
+    tls: process.env.REDIS_USE_TLS || false,
+  },
+}
+
+const client = redis.createClient()
+const prefix = process.env.BULL_PREFIX || 'bull'
+
+function refreshQueues() {
+  console.log('Refreshing Queues')
+  client.KEYS(`${prefix}:*`, (err, keys) => {
+    keys = [
+      ...new Set(keys.map(key => key.replace(/^.+?:(.+?):.+?$/, '$1'))),
+    ].map(name => new Queue(name, redisOptions))
+    setQueues(keys)
+  })
+}
+
+setInterval(refreshQueues, process.env.REFRESH_INTERVAL || 10000)
+
+refreshQueues()
+
+app.use('/ui', UI)
+
+app.listen(process.env.PORT || 3000, () => {
+  console.log('Running on 3000...')
+  console.log('For the UI, open http://localhost:3000/ui')
+  console.log('Make sure Redis is running on port 6379 by default')
+})
--- a/standalone.js
+++ b/standalone.js
@@ -1,4 +1,4 @@
-const { createQueues, setQueues, UI } = require('./')
+const { createQueues, setQueues, UI } = require('./index')
 const Queue = require('bull')
 const redis = require('redis')
 const app = require('express')()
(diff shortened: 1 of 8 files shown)


tombh:
I think there's a problem with the `refreshQueues()` function generating new Redis client connections every time. The number of client connections reported in the UI seems to grow and grow. Or that could just be some problem with the way I'm using Bull queues in my actual app...

tombh:
Thanks for the original code @JonDum!

I can't think of a case where 
```js
let queue_names = []
...
if (!queue_names.includes(queue_name)) {
  ...
  queue_names.push(queue_name)
}
```
isn't strong enough.

Of course it is possible for a plain JS array to have duplicates, but there is nowhere in the code where a duplicate could be inserted, thus it's safe?

JonDum:
You're right there, should be good on duplicates. What if someone deletes/removes queues though? I think in my app some queues are transient so I had it done that way so they would be cleared out when they were no longer in redis.

tombh:
Ah, I hadn't thought of that, so yeah we definitely need a way to manage duplicates then.

So the thing is that Bull Board's `setQueues()` doesn't seem to sync state, it just adds any new queues it's given - therefore it doesn't delete queues it already has but isn't given. I haven't dug into Bull Board's code to look into it, maybe someone more familiar could comment on that behaviour?

Another thing to bear in mind is that the reason I changed your `refreshQueues()` is that it was leaking Redis connections. So it'll still need a little tweak anyway to only issue `new Queue()` once for each new queue.

thelinuxlich:
I tried to run this image but it is complaining about password(which I don't want to setup any)

standalone.js
```diff
@@ -0,0 +1,47 @@
+const { setQueues, router } = require('./dist/index')
+const Queue = require('bull')
+const redis = require('redis')
+const app = require('express')()
+
+const redisOptions = {
+  redis: {
+    port: process.env.REDIS_PORT || 6379,
+    host: process.env.REDIS_HOST || 'localhost',
+    password: process.env.REDIS_PASSWORD || '',
```
srosset81, line 10: This doesn't work if no password is set, it should be
```suggestion
    password: process.env.REDIS_PASSWORD || undefined,
```

tombh:
Thanks @thelinuxlich and @srosset81. I fixed it. And rebased against latest master. And pushed and built the latest image to Docker Hub as `tombh/bull-board`, same image name as before.

tombh:
Is the delay to merging this the failing CI?

tombh:
I fixed the package.json conflict and commented out the failing line in the build step to deploy the Docker image to the Docker Hub. I can't see the failure in the snyk build step as it's behind a login.

standalone.js
```diff
@@ -0,0 +1,47 @@
+const { setQueues, router } = require('./dist/index')
```
vcapretz, line 1: could you convert this file to `.ts` and remove the npm lock file too, please?

airhorns:
Would be super awesome to get this in @tombh !

airhorns:
Ah -- one thing this is missing would be detection of `bullmq` queues vs plain ole `bull` queues, and the wrapping in the appropriate adapter. Is anyone aware of a way to infer which kind of queue it is by looking at redis?

tombh:
I'm definitely planning to sort this out, just life and stuff, etc, you know how it is...

erwineverts:
I have created a separate repository for a docker image: https://gitlab.com/erwineverts/bull-board-docker

henrysachs:
any possibility getting back some track on this? Would love this. I currently try to integrate that into nest with nx but its compiling all sources to a single main.js. thats why all references to files dont work.

felixmosh:
> any possibility getting back some track on this? Would love this. I currently try to integrate that into nest with nx but its compiling all sources to a single main.js. thats why all references to files dont work.

I'm not sure I got the issue.

This issue is too old, what version do you use?

henrysachs:
> 

Hey there I tried to explain my Issue a bit more in detail in: #347

felixmosh:
Closing since it is out of sync with master branch

PR #73, opened 2020-02-13
Introduce integration testing

### LINKED ISSUE #2: Add integration tests
Integration tests, a suggestion is to use [Dockest](https://github.com/erikengervall/dockest) to spin up a redis instance so it would be possible to add/remove jobs in queues and test the code

erikengervall:
Introduces dockest as a means to perform integration testing.
Supertest will be used to send and validate requests and responses.
closes #2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 node_modules/
-./static
+static
 dist
 yarn-error.log
 *.rdb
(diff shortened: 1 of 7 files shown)


erikengervall:
Yo @vcapretz, take a look 🙏🏼

PR #77, opened 2020-02-20
Add `promote button` for delayed jobs

gohiei:


Ic3m4n34:
An even nicer improvement would be if we could select multiple/all jobs to promote :-)

PR #80, opened 2020-03-03
import paths from correct place

vcapretz:
fix bug with new alpha version


src/index.ts
```diff
@@ -23,9 +23,9 @@ const router = express()
 router.locals.bullBoardQueues = bullBoardQueues
 
 router.set('view engine', 'ejs')
-router.set('views', `${__dirname}/../src/ui`)
+router.set('views', `${__dirname}/dist/ui`)
```
erikengervall: Looks better :+1:

Issue #65, opened 2020-01-26
bull-board(v1.0.0-alpha.0) getting errors while testing the alpha, might be doing something wrong?


arkhamRejek:
```javascript
import {  router } from 'bull-board';

app.use('/ui', router)

```

trace:
`
 at Function.render (/Users/dev/platform/node_modules/express/lib/application.js:580:17)
    at ServerResponse.render (/Users/dev/platform/node_modules/express/lib/response.js:1012:7)
    at exports.entryPoint (/Users/dev/platform/node_modules/bull-board/src/routes/index.ts:8:7)
    at Layer.handle [as handle_request] (/Users/dev/platform/node_modules/express/lib/router/layer.js:95:5)
    at next (/Users/dev/platform/node_modules/express/lib/router/route.js:137:13)
    at Route.dispatch (/Users/dev/platform/node_modules/express/lib/router/route.js:112:3)
    at Layer.handle [as handle_request] (/Users/dev/platform/node_modules/express/lib/router/layer.js:95:5)
    at /Users/dev/platform/node_modules/express/lib/router/index.js:281:22
    at Function.process_params (/Users/dev/platform/node_modules/express/lib/router/index.js:335:12)
    at next (/Users/dev/platform/node_modules/express/lib/router/index.js:275:10)
`

philefstat:
Along with a similar trace I'm getting the error message `Error: Failed to lookup view "index" in views directory ...`

jamesholcomb:
Same error as @philefstat { Error: Failed to lookup view "index" in views directory "/Users/jamesholcomb/code/myapp/node_modules/bull-board/dist/../src/ui"

vcapretz:
Hey folks! Thank you for testing the alpha, sorry for the bugs tho 😅
Is it still happening in the v1.0.0-alpha.1 version?

jamesholcomb:
v1.0.0-alpha.0

arkhamRejek:
same issues with v1.0.0-alpha.1

Deadly0:
It is possible to use bull-board with bullmq right now? Have the same issue with alpha1. Is the stable version works with bullmq?

vcapretz:
we are going to take a look at why is this problem happening and make another alpha version available! sorry for the delay 😄 

@Deadly0 the latest stable 0.6 does not support bullmq unfortunately 👍

luastoned:
[...]

PR #81, opened 2020-03-06
Fix index.ejs not copied to dist folder

hieptuanle:
Fix issue #65 

This PR just copy `index.ejs` file from `src/ui` to `dist/ui`. I add a new library `cpx` because I don't want to write a bash script and mess with Windows users.


package.json
```diff
@@ -66,6 +67,7 @@
     "babel-preset-react-app": "^7.0.2",
     "bull": "^3.12.1",
     "bullmq": "^1.6.6",
+    "cpx": "^1.5.0",
```
erikengervall: Couldn't find any documentation that this indeed supports windows, have you tried it yourself?

Also the docs state

> Copy file globs, watching for changes.
> 
> This module provides a CLI tool like cp, but with watching.

We're not really interested in watching for changes - could this have unforeseen side-effects? Would perhaps be better with a more plain file-copy lib.

vcapretz:
hey @hieptuanle, thank you very much for taking the time to contribute to the project, really appreciate it 😄 

I just merged #82 and the same issue was solve there, so I'll close your PR

if you want to contribute more I would be very glad to review more PRs 👍

package.json
hieptuanle: Actually i did not test it on Windows. I will test it next time 👍

hieptuanle:
@vcapretz No problems! Thanks for your useful project.

PR #87, opened 2020-03-18
Wrap encodeURIComponent to queueName in store methods

worotyns:
When you have "/" characters in queue names like:
@domain/queueName

Then UI will prepare requests like:
http://localhost:3000/ui/queues/@func/queueName/clean/failed

and server will resolve this as "404".

This fix uses urlEncode on queuName before we send request.
http://localhost:3000/ui/queues/%40func%2FqueueName/clean/failed

and server resolve with "200" status.

PS. I was use **encodeURIComponent()** instead of **qs.encode()** couse of compiling problems - need webpack reconfigurations.

### DIFF
--- a/src/ui/components/hooks/useStore.ts
+++ b/src/ui/components/hooks/useStore.ts
@@ -1,6 +1,5 @@
 import { useEffect, useRef, useState } from 'react'
 import qs from 'querystring'
-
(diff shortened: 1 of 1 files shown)


worotyns:
Ok ;)

Issue #37, opened 2019-11-03
is there a pagination ?


abdolrhman:
does all the orders appears in one page, no pagaination ? 
i could implement one if no any

vcapretz:
hey @abdolrhman! thanks for opening the issue! 

we don't have pagination yet 😅 if you have some time to implement it please feel free to do so! let me know if you need some help

abdolrhman:
Thanks buddy,
I will give it a shot,
This package did really helped me alot, 
Would be also great if u could helped me to contribute in this great thing :))

vcapretz:
glad that this has helped you! 🤩

you can open a PR starting the changes and we can iterate on that! the UI part would be trickier than allowing pagination on the API itself, so I would say starting from the back-end could be easier!

abdolrhman:
i am trying to do it from the front end, by adding 
`  const [currnetPage, setCurrentPage] = useState(1);

  const [jobsPerPage, setJobsPerPage] = useState(10);`

in useStore file, 
so that for each click i set a new current page and jobs for that page,
but this requires me to change a little bit in the cycle that is already exist,,

abdolrhman:
also i have a poor knowledege in React or in front-end in general, but am trying my best :)

vcapretz:
hey @abdolrhman, sorry for the late response!

Regarding the pagination, the two states you mentioned look like a nice start!
We would then pass their values to the `update` method so we can read the new query parameters in the route and use then in the `getDataForQeues.js` file. 
Let me know if I can help with anything else 😄

matthiasfeist:
I think pagination would be super important! especially for larger queues!

gabrielstuff:
Pagination start to be a thing for us. We are wondering how we are going to add it.

rodrigosegatto:
If queue > 1000 registers loading all list in one page?

rivernews:
Hi @vcapretz I found this issue interesting and would love to work on it. Really nice package by the way! But I wonder which version tag should I work on. In my project I'm using bull board `0.6.0` w/o any issue. When I fork the repo I realize `master` is now in `v1.0.0-alpha.7` and contains a TypeScript rewrite.

[...]

PR #105, opened 2020-05-27
Add Pagination for Jobs List

rivernews:
Implements #37 

## This PR contains
- Prev / Next buttons for navigating pages
- Jump to page by number input, easier to inspect later on jobs when job list is large. (press enter to take effect - perhaps not so intuitive, but comparing to taking effect on every typing changes, this approach would save unnecessary request on redis. But any feedback are welcome)
- User input validation & error prevention
- Pagination boundary checks

## Screenshot

image

This PR is based on the javascript version (`0.6.0`) to avoid wait on the latest TypeScript rewrite ([see discussion](https://github.com/vcapretz/bull-board/issues/37#issuecomment-633896558)), therefore this PR can only be merged to version `0.6.0` and not the latest alpha version. But once the project moves forward probably will need to refactor this ticket into TypeScript as well.

In case anyone wants to try use it in your project you can install by `yarn install @shaungc/bull-board@0.6.0--debug-01`.

### DIFF
--- a/routes/getDataForQeues.js
+++ b/routes/getDataForQeues.js
@@ -51,8 +51,6 @@ module.exports = async function getDataForQeues({ queues, query = {} }) {
     return { stats: {}, queues: [] }
   }
 
-  console.log('query', query)
-
   const pairs = Object.entries(queues)
 
   const counts = await Promise.all(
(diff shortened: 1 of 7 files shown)


vcapretz:
awesome PR! can you please close it and re-open but pointing to `latest-stable` instead of master please? this way we can keep releasing stable versions before our alpha 👍 

thanks again :D

rivernews:
Ah I see, no prob.

PR #107, opened 2020-05-27
Add Pagination for Jobs List

rivernews:
Implements #37 

## This PR contains
- Prev / Next buttons for navigating pages
- Jump to page by number input, easier to inspect later on jobs when job list is large. (press enter to take effect - perhaps not so intuitive, but comparing to taking effect on every typing changes, this approach would save unnecessary request on redis. But any feedback are welcome)
- User input validation & error prevention
- Pagination boundary checks

## Screenshot

image

This PR is based on the javascript version (`0.6.0`) to avoid wait on the latest TypeScript rewrite ([see discussion](https://github.com/vcapretz/bull-board/issues/37#issuecomment-633896558)), therefore this PR can only be merged to version `0.6.0` and not the latest alpha version. But once the project moves forward probably will need to refactor this ticket into TypeScript as well.

In case anyone wants to try use it in your project you can install by `yarn install @shaungc/bull-board@0.6.0--debug-01`.

### DIFF
--- a/routes/pagination.js
+++ b/routes/pagination.js
@@ -1,6 +1,9 @@
+const DEFAULT_PAGINATION_SIZE = 10
+
 module.exports = function getPaginationParameters(query) {
+  // default pagination indices
   let start = 0
-  let end = 9
+  let end = DEFAULT_PAGINATION_SIZE - 1
   if (query.start && query.end) {
     const parsedStart = parseInt(query.start)
     const parsedEnd = parseInt(query.end)
--- a/routes/pagination.js
+++ b/routes/pagination.js
@@ -4,10 +4,10 @@ module.exports = function getPaginationParameters(query) {
   if (query.start && query.end) {
     const parsedStart = parseInt(query.start)
     const parsedEnd = parseInt(query.end)
-    if (parsedStart !== NaN) {
+    if (!Number.isNaN(parsedStart)) {
       start = parsedStart
     }
-    if (parsedEnd !== NaN) {
+    if (!Number.isNaN(parsedEnd)) {
       end = parsedEnd
     }
   }
--- a/routes/pagination.js
+++ b/routes/pagination.js
@@ -1,4 +1,4 @@
-export const getPaginationParameters = query => {
+module.exports = function getPaginationParameters(query) {
   let start = 0
   let end = 9
   if (query.start && query.end) {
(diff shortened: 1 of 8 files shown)


ui/components/Paginator.js
```diff
@@ -0,0 +1,91 @@
...
+    if (inputValue !== NaN) {
+      if (inputValue > 0 && inputValue <= totalPages) {
+        const startIndex = (inputValue - 1) * 10
+        const endIndex = startIndex + (Math.min(Math.max(totalJobs - 1, 0), 9))
+        setPagination({
+          start: (inputValue - 1) * 10,
+          end: endIndex,
+        })
+      } else {
+        alert(`Invalid page number. Please input a number between 1 and ${totalPages}`)
```
vcapretz, line 50: nice error handling here!

routes/pagination.js
```diff
@@ -0,0 +1,16 @@
+module.exports = function getPaginationParameters(query) {
+  let start = 0
+  let end = 9
+  if (query.start && query.end) {
+    const parsedStart = parseInt(query.start)
+    const parsedEnd = parseInt(query.end)
+    if (parsedStart !== NaN) {
```
vcapretz, line 7: I know this can have some quirks with strict equality, like `NaN !== NaN == true` 😅 
maybe changing to `Number.isNaN(parsedStart)` is a safer way to handle the failed conversion

vcapretz:
awesome PR! just added some comments, but overall it's really good!

rivernews:
> hey, awesome PR!
> 
> just one thing I noticed is that it seems we have fixed 10 jobs per page right? can we extract this to a const so it's more clear when making calculations line in `Paginator.js` that references the number 10 in different places?
> 
> it would be even better to add a query param that makes this customisable, let me know what you think :D

I made the page size configurable in UI, how's this look like:

image

*Diff changes for page size are all in frontend, handled state, validation and render there.

routes/pagination.js
rivernews: Nice catch! Fixed using `Number.isNaN()`.

ui/components/Paginator.js
rivernews: Felt the logic a bit convoluted there, but hope it can catch most edge cases!

vcapretz:
thank you for the changes! I'm going to release it in v0.9.0. please give it a try and let me know if you find any bugs 😄

rivernews:
Awesome, will do! Thanks for building and maintaining this wonderful project ✨

PR #139, opened 2020-09-22
add null check when job is undefined

### LINKED ISSUE #43: Cannot read property 'id' of null
Reported by sentry, I'm not sure on which specific conditions this happens:

`TypeError: Cannot read property 'id' of null
  File "/usr/src/app/node_modules/bull-board/routes/getDataForQeues.js", line 24, col 13, in formatJob
    id: job.id,
  ?, in Array.map
  File "/usr/src/app/node_modules/bull-board/routes/getDataForQeues.js", line 67, col 20, in null.<anonymous>
    jobs: jobs.map(formatJob),
  ?, in runMicrotasks
  File "internal/process/task_queues.js", line 93, col 5, in processTicksAndRejections
  ?, in async Promise.all
  File "/usr/src/app/node_modules/bull-board/routes/getDataForQeues.js", line 54, col 18, in async getDataForQeues
    const counts = await Promise.all(
  File "/usr/src/app/node_modules/bull-board/routes/queues.js", line 5, col 5, in async handler
    await getDataForQeues({`

hisabimbola:
This fixes a case where bull returns `null` for a job. I am not sure why bull does that but this PR adds a null check to prevent the app from crashing

fixes #43

### DIFF
--- a/routes/getDataForQeues.js
+++ b/routes/getDataForQeues.js
@@ -70,7 +70,7 @@ module.exports = async function getDataForQeues({ queues, query = {} }) {
       return {
         name,
         counts,
-        jobs: jobs.map(formatJob),
+        jobs: jobs.filter((job) => !!job).map(formatJob),
       }
     }),
   )


routes/getDataForQeues.js
```diff
@@ -70,7 +70,7 @@ module.exports = async function getDataForQeues({ queues, query = {} }) {
       return {
         name,
         counts,
-        jobs: jobs.map(formatJob),
+        jobs: jobs.filter((job) => !!job).map(formatJob),
```
erikengervall, line 73: A `reduce` instead of a `filter` -> `map` would reduce the iteration count by up to 2x 🙂 

I'd recommend going with that instead for this case!

erikengervall (changes requested):

lukepolo:
To avoid this error for now you can remove `removeOnComplete` ... gross but i only use bull board for dev so not a huge deal.

routes/getDataForQeues.js
robman87: return {
      name,
      counts,
      jobs: jobs
        .reduce((result, job) => {
          if (job) {
            result.push(formatJob(job))
          }
          return result
        }, [])
    }

felixmosh: I don't agree with you, this consumes 2x memory, it is better to use `filter` & `map` as it is more declarative

felixmosh (approved):

routes/getDataForQeues.js
felixmosh: I don't agree with you, this consumes 2x memory, it is better to use filter & map as it is more declarative

PR #160, opened 2020-11-24
fix: long job id overlaps job name, fixes #159

### LINKED ISSUE #159: Design bug: jobId overlap job name
I noticed a bug in the design (CSS I guess) of the bull-board ui. When the jobId is too long, it will overlap with the job name.

bulllboard_overlap

I suggest to use an ellipsis for the jobId for instance, since you can still view it clicking on the "options" tab. Or maybe changing the vertical alignement.
Cheers,
Paul

felixmosh:


vcapretz (approved):
Nice one!

PR #163, opened 2020-11-24
Quick glance at stats per queue

lukepolo:
The old version had a nice way to display stats quickly , I did a quick example of how it could look. If anyone thinks this has merit I can try to make it look better.

Quick Glance at Stats

### DIFF
--- a/src/ui/components/App.tsx
+++ b/src/ui/components/App.tsx
@@ -45,7 +45,7 @@ export const App = ({ basePath }: { basePath: string }) => {
           )}
         </div>
       </main>
-      <Menu queues={state.data?.queues.map(q => q.name)} />
+      <Menu queues={state.data?.queues} />
     </BrowserRouter>
   )
 }
--- a/src/ui/components/Menu/Menu.module.css
+++ b/src/ui/components/Menu/Menu.module.css
@@ -41,7 +41,6 @@
   color: inherit;
   text-decoration: none;
   display: block;
-  padding: 1rem 1.25rem;
   white-space: nowrap;
   transition: background-color 100ms ease-in;
   border-left: 3px solid transparent;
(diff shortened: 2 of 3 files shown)


Jivings:
This looks great, is it in a working state?

felixmosh:
In my opinion, it makes the UI cluttered and full of details. (Look at your screenshot, there many dots that has 0 jobs => redundant info).

I've thought about an `i` icon which will open on click / hover a popover with these details.
This way has 2 benefits,
1. The UI remains clean
2. You don't need to pass the stats of **all queues** on each request, it can be fetched only when the popover is open.

WDYT?

Jivings:
I need a way to quickly see failed/waiting/complete. The other numbers are not as important. Could we add these without adding too much clutter?

I don't think an **_i_** works as well for me. I need to see at a glance if any queue is smashing failed jobs.

felixmosh:
@Jivings how about opening a modal that will have stats of all the queues?

lukepolo:
I’d almost rather have another view like a dashboard type of thing that we could put up on a screen , not sure others use cases

felixmosh:
@lukepolo , an interesting idea!
Let's do it within a dashboard page

Jivings:
Yeah that would work for me too as long as I can get to the respective queue easily from the dashboard 👍

Deadly0:
I need this feature too. A dashboard page or an option to switch queue list to displaying queues status.

mannharleen:
Hello hello, is there feature being added soon?

Jivings:
I created something a bit simpler for this. The sidebar has a bar that represents the % of success/pending/failed jobs.

image

It means I can see if there are any queues that have serious problems that I need to take a closer look at. This works for me until we have a better solution.

PR #164, opened 2020-11-25
Fix crash when jobs are null #139

robman87:


vcapretz:
same thing here as #165 here 😛  if it is still a bug in v1 let us know, please! 

thanks once more 👍

PR #165, opened 2020-11-25
Fix crash when query is empty

robman87:
Get jobs for all statuses by default.


vcapretz:
hey @robman87, we released 1.0 as a stable version, which means we would encourage users to bump the version and we will remove the `latest-stable` soon, can you please try reproducing the error on v1 and let us know if it sill happens in there? 

if so please feel free to send a PR as well 😄 

thanks!

PR #175, opened 2020-11-26
Prevent getting jobs for undefined status

robman87:


felixmosh:
@robman87 what was the scenario that you've encountered with, that this PR fixes?

robman87:
Don't have a specific scenario. We wanted to use bull-board in production at work and it was crashing, traced the issue back to jobs becoming null for some reason. Had the same issue when using bull-arena package. Same fix there, filter jobs that are null. It's not a solution to the root cause but it prevents the script from crashing and jobs are displayed.

PR #176, opened 2020-11-26
Make sure only valid jobs are processed

robman87:
... by formatJob, prevents crash when jobs are null.


robman87:
Fixes #139 that is still left in master branch, latest-stable-branch is no longer maintained.

PR #177, opened 2020-11-27
refactor: use adapters as queue access

### LINKED ISSUE #174: Task: Remove bullMQ deps
Currently there are 2 check which are specific to bullMQ
1. https://github.com/vcapretz/bull-board/blob/master/src/index.ts#L43
2. https://github.com/vcapretz/bull-board/blob/master/src/routes/cleanAll.ts#L28

This usages are require us to put bullMQ in our deps even though it is not required.

My suggestion is to provide 2 wrappers one for each Q, which provides the **same** API.
`setQueues` will accept them.
Innerly each wrapper will make the "unique" calls.

felixmosh:
Makes the code cleaner and not relaying on queue types

closes #174


felixmosh:
@vcapretz WDYT?

vcapretz (approved):

vcapretz:
do we need to update the docs @felixmosh? if so, can you include it here in the same PR please?

felixmosh:
Should we release a minor / major?

PR #188, opened 2020-12-05
Maintenance & build optimization

felixmosh:
- [x] Bump deps to latest
- [x] Clean up deps
- [x] Add css minification on prod
- [x] Enable vendor chunk
- [x] Improves webpack's tree-shaking

PR #189, opened 2020-12-05
HMR for dev mode

felixmosh:
This PR adds:
1. HMR for styles
2. React fast refresh for react components
3. Open dev env on `yarn build:ui:watch`
4. Adds error overlay

In order to enable it, 
1. Move to babel as TS transpiler (instead of ts-loader)
2. Enable webpack-dev-server to handle HMR (proxies `/ui` to the origin server)

GZJzaILJay

**This PR relays on merge of** https://github.com/vcapretz/bull-board/pull/188

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,0 +1,11 @@
+module.exports = {
+  presets: [
+    '@babel/preset-env',
+    '@babel/preset-react',
+    '@babel/preset-typescript',
+  ],
+  plugins: [
+    '@babel/plugin-proposal-object-rest-spread',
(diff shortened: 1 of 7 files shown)


vcapretz:
do we still need this one even after #188? seems like they are also doing the same stuff (like some packages bumps), maybe we can merge that one and rebase it here before taking the time to review 👍

felixmosh:
We do need this pr since the #188 is only a deps bump, which this PR "continues".

vcapretz (approved):

Issue #114, opened 2020-06-29
pagination not in alpha


jadkins89:
First off, thank you for the time and energy you've devoted to this project. I have been using your alpha version to get some access to some very handy features (replaceQueues, clean all in completed, etc). I had to rollback to 0.9.0 though because pagination is necessary since I'm handling thousands of jobs sometimes. Not having clean all in completed though then becomes a real annoyance. Is there any chance we could see an update to alpha that brings in pagination?

vcapretz:
hey, thank you very much for your comment! 

I've been a bit busy but today I managed to review/merge some PRs and released a new alpha version. I'll try working this weekend on porting pagination to the alpha, you can follow the progress in this same issue!

sellomkantjwa:
Has pagination not made into 1.0.1 as well? Im only seeing the first 11 jobs.

felixmosh:
> Has pagination not made into 1.0.1 as well? Im only seeing the first 11 jobs.

You are right, I'v missed this feature, I will review on how it can be returned.

MatayoshiMariano:
@felixmosh any updates on this?

felixmosh:
Not yet, I'm working on #265, it consumes some time from me.

PR #191, opened 2020-12-06
Feature/paginate

s-r-x:
Pagination ^-^
#114 

- Pass _start/_end query params to adapters inside router
- minor state redesign for a new "page" variable and reset it on every "selectedStatuses" update
- view powered by react-paginate

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "query-string": "^6.13.7",
     "react": "17.0.1",
     "react-dom": "17.0.1",
+    "react-paginate": "6.3.2",
     "react-router-dom": "^5.2.0",
     "redis-info": "^3.0.8"
   },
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -49,11 +49,12 @@
     "date-fns": "2.10.0",
     "ejs": "3.0.1",
     "express": "4.17.1",
-    "highlight.js": "^10.4.1",
+    "highlight.js": "^9.18.3",
```
vcapretz, line 52: do this one needs to be rebased with master so we don't downgrade the version?

s-r-x: That specific commit(75ddcfd) can be removed actually, if highlight.js error was fixed on master

s-r-x: rebased to master

MatayoshiMariano:
@s-r-x why this was deleted? bull-board is not showing all jobs

PR #196, opened 2020-12-10
Add auth example

felixmosh:
This PR adds an example with passport auth.


examples/with-auth/index.js
```diff
@@ -0,0 +1,118 @@
...
+
+// Configure the local strategy for use by Passport.
+//
+// The local strategy require a `verify` function which receives the credentials
+// (`username` and `password`) submitted by the user.  The function must verify
+// that the password is correct and then invoke `cb` with a user object, which
+// will be set at `req.user` in route handlers after authentication.
+passport.use(new LocalStrategy(
+  function (username, password, cb) {
+    if (username === 'bull' && password === 'board') {
```
joaovbibiano, line 19: Should we put it insde the .env file? I think is the best solution at first glance

joaovbibiano (approved):

examples/with-auth/index.js
felixmosh: it is just an POC example, do you think that it is better to follow the best practices?

joaovbibiano: Hahah sorry @felixmosh I'm new to reviewing open source projects.

The main idea seems good to me, I think it's simple and I solve our problem.

felixmosh: I don't think that there is much "special" here, just a simple passport example (which is copied with the comments from passport's example :] )

joaovbibiano: Ok, good.

Now, we need other Collaborator approve this PR? So, we can finish the implementation? What is the worflow?

felixmosh:
@joaovbibiano can you approve this PR?

examples/with-auth/index.js
felixmosh: Just approve the PR it self, then I will be able to merge it, thanx!

joaovbibiano (approved):

joaovbibiano (approved):

joaovbibiano (approved):
This PR is directly simple, and solves the proposed problem

vcapretz (approved):

PR #204, opened 2020-12-21
Custom job data formatter

Jivings:
## Summary 

User can now specify AdapterOptions passed to the BullAdapter on creation. In this user can specify the dataFormatter, which provides an alternative way to parse job data to be displayed in the Jobs List UI.

## Rationale

I have a rather large payload that I send over my queue, and parsing that causes the UI to be very unresponsive. Using a custom data parser I can now instead show a subset of the full payload in the UI.

image

image

### DIFF
--- a/src/queueAdapters/bull.ts
+++ b/src/queueAdapters/bull.ts
@@ -1,17 +1,18 @@
-import { Job, Queue } from 'bull'
 import {
+  AdapterOptions,
   JobCleanStatus,
   JobCounts,
   JobStatus,
   QueueAdapter,
 } from '../@types/app'
+import { Job, Queue } from 'bull'
 
 export class BullAdapter implements QueueAdapter {
   public get client() {
     return Promise.resolve(this.queue.client)
   }
 
-  constructor(public queue: Queue) {}
+  constructor(public queue: Queue, public options: AdapterOptions) {}
 
   public getName(): string {
     return this.queue.name
--- a/src/queueAdapters/bull.ts
+++ b/src/queueAdapters/bull.ts
@@ -12,7 +12,7 @@ export class BullAdapter implements QueueAdapter {
     return Promise.resolve(this.queue.client)
   }
 
-  constructor(public queue: Queue, public options: AdapterOptions) {}
+  constructor(public queue: Queue, public options?: AdapterOptions) {}
 
   public getName(): string {
     return this.queue.name
--- a/src/routes/queues.ts
+++ b/src/routes/queues.ts
@@ -40,7 +40,7 @@ const getStats = async ({
   return validMetrics
 }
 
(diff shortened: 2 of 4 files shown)


src/queueAdapters/bull.ts
```diff
@@ -1,17 +1,18 @@
...
 } from '../@types/app'
+import { Job, Queue } from 'bull'
 
 export class BullAdapter implements QueueAdapter {
   public get client() {
     return Promise.resolve(this.queue.client)
   }
 
-  constructor(public queue: Queue) {}
+  constructor(public queue: Queue, public options: AdapterOptions) {}
```
felixmosh, line 15: Missing implementation on the `BullMQAdapter` as well

src/routes/queues.ts
```diff
@@ -88,11 +91,11 @@ const getDataForQueues = async (
       const status =
         query[name] === 'latest' ? statuses : (query[name] as JobStatus[])
       const jobs = await queue.getJobs(status, 0, 10)
-
+      const formatter = formatJob(queue.options?.dataFormatter)
```
felixmosh, line 94: I think that it is better to incapsulate the formatting part inside the adapters (inside `getJobs`), therefore, making `AdapterOptions` as `private` field.

felixmosh (changes requested):

felixmosh:
Thank you for this PR, it is a great idea to filter out some data fields :]

src/queueAdapters/bull.ts
Jivings: Yeah, I'm not using the other Adapter, so perhaps someone else can add & test that?

src/routes/queues.ts
Jivings: Yasss, maybe this is more correct. Will update 👍

src/queueAdapters/bull.ts
felixmosh: It has the same `getJob` implementation, with the same fields.

Jivings: Okay will update I just wont be able to test 🍕

src/routes/queues.ts
Jivings: I tried to do this but reached the extent of my typescript knowledge and am bumbling around trying to figure it out. Sorry no more time to spend on it now.

src/queueAdapters/bull.ts
Jivings: Done

felixmosh:
I think that it can be handled from a different angle, since the issue that you are trying to solve related to many cards with heavy data, I think that we should consider some virtual-scroller implementation.

Do you think this will solve your issue?

Jivings:
It will solve the heavy DOM issue, but I also want to not show personal data in the UI :)

PR #218, opened 2021-01-20
Close #217 Configure basePath doc

### LINKED ISSUE #217: How do I use basePath?
I would like to host the bull-board as an admin addon to one of my image services.

I works fine when I'm on localhost ie. http://localhost:3000/queuestatus.
But on my server the app is hosted under a sub path: https://<server_name>/<application_name>/queuestatus. 
I've seen a basePath variable in the code but no documentation on how to configure it.
Now the bull-board app fails to load the css and js files from the server because it tries to load them from https://<server_name>/queuestatus/<resources>...

I would like to see some documentation/example on how to use the basePath variable.

Thanks in advance.
/Andreas

softbrix:
Added documentation how to configure basePath in the Readme.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,25 +78,25 @@ That's it! Now you can access the `/admin/queues` route and you will be able to 
 
 ### Hosting router on a sub path
 
-If you host your express service on a sub path (ie. https://<server_name>/<base_path>/) then you can add the following code to provide the configuration to the bull-board router. In this example the sub path will be `taurus`.
+If you host your express service on a different path than root (/) ie. https://<server_name>/<sub_path>/, then you can add the following code to provide the configuration to the bull-board router. In this example the sub path will be `my-base-path`.
 
 ```js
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,6 +76,28 @@ app.use('/admin/queues', router)
...
+
+```js
+const { router } = require('bull-board');
+
+// ... express server configuration
+
+let basePath = 'taurus';
+
+app.use(
+  '/admin/queuestat',
```
felixmosh, line 91: Is it should be `/admin/queues`? 
I think that the example should be simplified,
Mount under `/my-base-path`, and set the `proxyUrl` to it.

felixmosh (changes requested):

felixmosh:
Thank you for updating this :]
Can you Mark as resolved, So I will able to merge it? :]

felixmosh (approved):

PR #220, opened 2021-02-01
Add read only mode, closes #219

### LINKED ISSUE #219: Feature Request: Read-only mode
I think it would be very beneficial to allow a user to configure the board to a read-only mode when configuring routes. In the UI, the buttons that affect the state could be grayed out.

My team is using bull board for observability. We would like to open up its usage to a select group of users, but we cannot allow them to modify the state of bull queues. This is why I suggest having an additional configuration object that could be passed into 'setQueues()'.

So something like the following with readonly being an optional field, defaulting to false.
```ts
setQueues( [ new BullAdapter(someQueue) ], { readonly : true } )
```

felixmosh:
This PR adds a `readOnlyMode` per queue.
In this mode, all queue & jobs related actions are removed from the UI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,25 @@ app.use('/admin/queues', router)
 
 That's it! Now you can access the `/admin/queues` route and you will be able to monitor everything that is happening in your queues 😁
 
+### Queue options
+1. `readOnlyMode` (default: `false`)
+Makes the UI as read only,hides all queue & jobs related actions
+   ```js
+    const Queue = require('bull')
+    const QueueMQ = require('bullmq')
+    const { setQueues, BullMQAdapter, BullAdapter } = require('bull-board')
+    
+    const someQueue = new Queue()
+    const someOtherQueue = new Queue()
+    const queueMQ = new QueueMQ()
+    
+    setQueues([
+      new BullAdapter(someQueue, { readOnlyMode: true }), // only this queue will be in read only mode
+      new BullAdapter(someOtherQueue),
+      new BullMQAdapter(queueMQ),
+    ]);
+   ```
+
 ### Hosting router on a sub path
 
 If you host your express service on a different path than root (/) ie. https://<server_name>/<sub_path>/, then you can add the following code to provide the configuration to the bull-board router. In this example the sub path will be `my-base-path`.
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ That's it! Now you can access the `/admin/queues` route and you will be able to 
 
 ### Queue options
 1. `readOnlyMode` (default: `false`)
-Makes the UI as read only,hides all queue & jobs related actions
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -76,6 +76,25 @@ app.use('/admin/queues', router)
...
+    const { setQueues, BullMQAdapter, BullAdapter } = require('bull-board')
+    
+    const someQueue = new Queue()
+    const someOtherQueue = new Queue()
+    const queueMQ = new QueueMQ()
+    
+    setQueues([
+      new BullAdapter(someQueue, { readOnlyMode: true }), // only this queue will be in read only mode
+      new BullAdapter(someOtherQueue),
+      new BullMQAdapter(queueMQ),
```
japrescott, line 94: make it clear that this new option also applied to `BullMQAdapter`

felixmosh: Any suggestion of how?

jeremysmith1: Something like this would show the full usage.
```ts
    const someQueue = new Queue()
    const readonlyQueue = new Queue()
    const queueMQ = new QueueMQ()
    const readonlyQueueMQ = new QueueMQ()
    
    setQueues([
      new BullAdapter(someQueue),
      new BullAdapter(readonlyQueue, { readOnlyMode: true }), // queue dashboard will be in read only mode
      new BullMQAdapter(queueMQ),
      new BullMQAdapter(readonlyQueueMQ, { readOnlyMode: true}), // queue dashboard will be in read only mode
```

README.md
```diff
@@ -76,6 +76,25 @@ app.use('/admin/queues', router)
 
 That's it! Now you can access the `/admin/queues` route and you will be able to monitor everything that is happening in your queues 😁
 
+### Queue options
+1. `readOnlyMode` (default: `false`)
```
jeremysmith1, line 80: @felixmosh I really like how you set it up so each queue can have different readonly settings. =)

src/@types/app.ts
```diff
@@ -17,6 +17,8 @@ export type JobCounts = Record<JobStatus, number>
 
 export interface QueueAdapter {
   readonly client: Promise<Redis.Redis>
+  readonly readOnlyMode: boolean
```
jeremysmith1, line 20: Would it make sense to mark this as an optional parameter like
```ts
  readonly readOnlyMode?: boolean
```

felixmosh: This is a class property definition, the value will be always boolean
https://github.com/vcapretz/bull-board/pull/220/files#diff-0acb621af47463a554bde3931cee27d24841eb1ceb1c3efc3e789ea2f4779045R18

jeremysmith1 (approved):

felixmosh:
@vcapretz can you review(and approve) it as well?

vcapretz (approved):

PR #221, opened 2021-02-01
Refactor error handlers in routes & ui

felixmosh:
This PR adds errorHandler for all routes, and handles error in the UI (popups a toast)
Changes made:
1. using `axios` in order to fetch from the server
2. add `react-toastify`
3. remove `query-string` - axios handles query params encoding
4. global error handler in routes, no need wrap each route handler with `try & catch`
5. global error handler of api requests

image


felixmosh:
@vcapretz can you review this pr please :]

vcapretz (approved):
really nice work!! thanks for that :D

PR #228, opened 2021-02-11
Display failed reason (if there is no stacktrace to show)

josmas:
This is reproducible with the example server, which at times makes jobs fail and stall (at random).

I am seeing some failed jobs with a blank Error section, as in the following screenshot:
with_out_fix

This is due to the job stalling and not having a `stacktrace` to display, so in cases like that, I'd like to see the `failedReason` instead (screenshot with fix):
with_fix


src/ui/hooks/useDetailsTabs.tsx
```diff
@@ -47,6 +47,9 @@ export function useDetailsTabs(currentStatus: Status) {
           return (
             <>
               {!failedReason && <div className="error">{'NA'}</div>}
+              {stacktrace && stacktrace.length === 0 && (
```
felixmosh, line 50: If the `stacktrace` is empty there is no good reason to render `<Highlight>`...
It should look like
```js
     {stacktrace?.length === 0 ? (
        <div className="error">{!!failedReason ? failedReason : "NA"}</div>
      ) : (
        <Highlight language="stacktrace" key="stacktrace">
          {stacktrace}
        </Highlight>
      )}
```

felixmosh (changes requested):

felixmosh (approved):

felixmosh:
Thank you for the PR 🙏🏼

josmas:
@felixmosh thanks to you for the quick feedback and merge!

Issue #178, opened 2020-12-01
Mark the failed jobs


gohiei:
If there are some failed jobs, it maybe a serious problem.
I suggest some warnings or marks, such as: red background color, or `!` icon...etc.

felixmosh:
Where?

gohiei:
Left menu item (NavLink).
Modify the item color, or append icon.

felixmosh:
Just to clarify, you want to mark a queue tab with at least one failing job?

gohiei:
Yes, it would be helpful.

FPNL:
Yes, maybe something like Kue js:
https://github.com/stonecircle/kue-ui

Though, it's deprecated, but it summary all the info at a glance.

felixmosh:
It looks like that the summary there was under accordion, which means that you need to click the queue name in order to see the data. The same thing is at bull-board, you can click on the queue name and see the stats.

Lets research ux/ui wise option on how to show these stats...

Ayora29:
Hello there, I will second this request. May be color the name of the queue in red if there is at least one error in the queue.
Like this it does not occupy any more space in the queue menu.

felixmosh:
Does the new Dashboard page solved this?
![dashboard](https://raw.githubusercontent.com/felixmosh/bull-board/master/screenshots/overview.png)

Ayora29:
Yes it did. thank you.

PR #252, opened 2021-04-15
Queue job count badges in menu

webdevian:
Under each queue name show colour coded badges for the job count of each status (if > 0)


felixmosh:
Thank for this PR 🙏

Can u attach the look of it?

webdevian:
@felixmosh Sure - 
image

The status text shows on hovering on the badge (with title attribute)

felixmosh:
It is kinda related to the task #178
Lets make a "dashboard" page with all the stats per page... WDYT?

webdevian:
I think it already does what that task is asking for. My inspiration for Kue-UI, we used that before switching to Bull

felixmosh:
@webdevian as I've told in #163, I don't like this cluttered UI solution, which will fill the screen with distracting info.

Let's work on the #258 (Dashboard page) task.

Issue #251, opened 2021-04-14
cluster support list shuffle


Heartz66:
I am using bull-board inside a cluster. All requests to bull-board router are round-robin. Every request will shuffle the queue list on the UI. This can be solved by sorting the queue list on alphabetical order.

felixmosh:
PR is very welcome

felixmosh:
@Heartz66 Thank you for the PR 🙏🏼

Can you elaborate on the reason u need to run `bull-board` in cluster mode?

felixmosh:
Closed by 6f2c487

Heartz66:
> @Heartz66 Thank you for the PR 🙏🏼
> 
> Can you elaborate on the reason u need to run `bull-board` in cluster mode?

I'm using bull-board on a node application running within a kubernetes cluster. Your most recent commit unfortunately does not solve this issue.

This issue occurs because queues are dynamically registered whey they become neccesary in the application.

![demo](https://user-images.githubusercontent.com/8833887/116795119-e498b800-aad2-11eb-8469-2c950ed2b399.gif)

felixmosh:
@Heartz66

Heartz66:
> @Heartz66, `bull-board` now keeps the order of inserted queues.
> The bug that you've mentioned is related to the fact that you are inserting them on each instance with a different order, can you sort the list before inserting?

I dynamically create Bull queues and insert them into bull-board, the timings are totally different on each application within the cluster.

If I can retrieve all the queues currently registered on the bull-board instance, I could also simply add the queue to the list and sort it everytime I add a queue to the bull-board instance.

felixmosh:
You can query `redis` directly with `KEYS bull:*:id` (based on https://github.com/OptimalBits/bull/issues/334) and then insert it to `bull-board`

Heartz66:
> You can query `redis` directly with `KEYS bull:*:id` (based on [OptimalBits/bull#334](https://github.com/OptimalBits/bull/issues/334)) and then insert it to `bull-board`

This does the job as well, thank you very much.

hoangtv49:
> You can query `redis` directly with `KEYS bull:*:id` (based on [OptimalBits/bull#334](https://github.com/OptimalBits/bull/issues/334)) and then insert it to `bull-board`

Can you give me code to insert queue to bull-board by query redis please ?

PR #259, opened 2021-04-26
Sort queue list for UI #251

Heartz66:
solves https://github.com/felixmosh/bull-board/issues/251


felixmosh:
Pls fix the lint errors

felixmosh:
Closed by 6f2c487

felixmosh:
@Heartz66, thank you for the PR
I've refactored the code to use `Map`s which preserve the insert order, which in my opinion this is better approach to keep the original order.

PR #267, opened 2021-05-07
Fix typo (bullMQAdapter to BullMQAdapter)

patdx:
According to this file:

https://github.com/felixmosh/bull-board/blob/master/bullMQAdapter.d.ts

The correct name should be `BullMQAdapter`  (with uppercase B).


felixmosh (approved):

felixmosh:
Thank you 🙏

Issue #265, opened 2021-05-04
Feature: Allow using different server frameworks


felixmosh:
There are several issues (#225 #212 #29 #103 #248 ) related to the fact that `bull-board` is coupled to `Express.js` as server framework.

The idea is to refactor the code to be framework agnostic & expose framework specific adapters

PR #280, opened 2021-05-27
Allow set queue limit using query parameters

ephillipe:
While we don't have full-featured pagination this code will allow asking the API for more items to show.


erickalmeidaptc:
When we navigate between the tabs, the UI uses query parameters to send the current tab to the LIB. I don't see this as half feature as it is only to remove the hardcoded limit that exists on the code. Instead of having an immutable hard code limiting the number of jobs returned, we can see it as a way to configure the behaviour.

image

felixmosh:
@erickalmeidaptc I agree, but nothing in the UI will propagate this query-string to the api calls.

I think that it is a pretty easy task to add a full blown pagination feature, and it is better to implement it, instead of workarounds that we will need to support.

Currently, I'm working on a multi-server feature which will solve multiple open requests (#265).
I will be more than happy to review a pagination review, or to implement it by my self after I finish my current task.

felixmosh:
Closed by #287

PR #295, opened 2021-06-12
Display stacktrace on delayed status page

cbovis:
This PR updates the job card to display the jobs stacktrace if the job is in the delayed status. This situation can occur if the job fails but is set to retry one or more times with a backoff strategy. Displaying the stacktrace here helps identify jobs which have been delayed due to error vs being delayed manually and ensures that we don't need to wait until all attempts have been exhausted to identify an error.

> **Note:** I've deliberately designed this to only show stacktraces for failed and delayed statuses because I feel it would be confusing/misleading for a stacktrace to be shown in other states. For example, displaying it in the active state may lead the user to believe that the current execution has failed and the job has not yet been moved to another status.

Screenshot 2021-06-12 at 11 33 15

### DIFF
--- a/packages/ui/src/components/JobCard/Details/Details.tsx
+++ b/packages/ui/src/components/JobCard/Details/Details.tsx
@@ -13,7 +13,10 @@ interface DetailsProps {
 }
 
 export const Details = ({ status, job, actions }: DetailsProps) => {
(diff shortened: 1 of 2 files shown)


felixmosh:
@cbovis In the latest version, failed job always will have the error tab (with the stack trace) no matter on queue status you are in.

For example a failed job that was retried and in "waiting" queue.
image

PR #306, opened 2021-07-26
add support for IE browsers

rfnunes01:


felixmosh:
Thank you for the PR.

Your approach takes us back, this PR will make bundle size of the front end much larger for *all* users even if your browser supports all features.

Check how next.js does it (module/nomodule pattern) which will create 2 bundles and load them to the right browser.

felixmosh (changes requested):
Please changes specified in the comment

PR #329, opened 2021-09-22
Remove unused return value from `createBullBoard()`

cincodenada:
The express just uses serverAdapter.getRouter(), so I think this is uneeded


felixmosh:
Thank you 🙏🏼

PR #334, opened 2021-09-30
Added capability to empty waiting list

leofmds:
Added a button "Empty" to clean the waiting list


felixmosh:
Thank you for this PR 🙏🏼
 The drain method is queue related method, it removes all waiting & delayed jobs, therefore I don't think that it should be under the waiting tab.

> Drains the queue, i.e., removes all jobs that are waiting or delayed, but not active, completed or failed.

If your requirement it to remove waiting only jobs, you can change your PR to use the `clean` end point which gets the status to remove.

Similar to cleanAllDelayed here   const cleanAllDelayed = (queueName: string) =>

felixmosh:
Ping?

leofmds:
Hi,
I've tried to use the clean method but it wasn't working, so I've implemented the empty method. Later I realised that it probably didn't work because the waiting list should be passed as 'wait' and not 'waiting', which is the default value for the waiting queue. I couldn't work on it the last few days, but will try to fix it on the next week.

PR #336, opened 2021-10-01
Simplify docs for BullMQ based on Fastify

mbaev:
I'll be glad to see your comments.

Thanks


felixmosh:
I don't think that the example is complicated.

There are some points that you've removed, and they are mandatory (like converting delay to number)

mbaev:
No problem, have a good day.

Issue #277, opened 2021-05-19
Feature: Add support for the new Flow method of bullmq


felixmosh:
BullMQ added a Flow method with `waiting-children` status.

It can be cool to add support for this, and maybe add links to the parent job

loicnestler:
Any updates on this? 👀

felixmosh:
I will be more than happy to review a PR for this feature 😀

PR #339, opened 2021-10-21
Add support for waiting-children

lucasavila00:
BullMQ flows have a waiting-children type of job.

This is a first step into having better support for flows, as discussed in: https://github.com/felixmosh/bull-board/issues/277

This PR adds a new tab for the waiting-children jobs, as well as the button to delete them, individually:
image

I have tested it manually and everything works. But I'm lost as to where/how I should add tests for the changes in this PR. Please let me know how that should be done...


jefftudorbenbria:
I am currently looking for a way to visualize the parent/child relationship so its pretty cool to see this.

A view of the aggregated child jobs of a flow would be ideal for my use case.

felixmosh:
Feel free to reopen if still relevant

all43:
Better to show waiting-children jobs without dependencies, than not showing job with children until it finishes.
If one of children job stuck or fail, parent job won't be visible at all.
As partial fix we could show them in waiting tab, maybe with asterisk or some symbol.
According to the docs, deleting parent job will automatically delete its children.

PR #358, opened 2021-12-08
Added getTitle support

### LINKED ISSUE #349: Feature Request: Extract data from job to use as job Title
A few versions back, I was "abusing" the named job parameter to have better titles on the dashboard. After upgrading to the latest; when named jobs are used we get the lovely "Missing process handler" error. I'll try working on a PR to add support to make this a little more generic which can be useful for other cases.

Original named jobs method that I was abusing to accomplish this goal:
```js
await bullQueue.add(jobItem.title, jobItem)
```

---

### Proposal 1:

Append `title` parameter to `QueueOptions`, which is just a generic JSON extractor.

Example:
```js
const bullQueue = new Bull('job-queue', { title: "jobTitle" })

...

const jobItem = {
  jobTitle: item.page.title,
  url: item.page.href,
}

await bullQueue.add(jobItem)
```

---

### Proposal 2:
Append `jobTitle?` parameter to `JobOptions`, which is just a generic JSON extractor.

Example:
```js
const bullQueue = new Bull('job-queue')

...

const jobItem = {
  jobTitle: item.page.title,
  url: item.page.href,
}

await bullQueue.add(jobItem, { jobTitle: "jobTitle" })
```

---

### Expected Output:

Expected title output would be something along the lines of:
```js
`${jobName} - ${jobTitle}`
```

---

### Bonus options:

- `replace?: boolean | undefined;` - Replace entire job title for the expected output section, instead of appending `jobTitle` to `jobName`.

bugs181:
Closes #349

Would something like this work?

### DIFF
--- a/packages/api/src/handlers/queues.ts
+++ b/packages/api/src/handlers/queues.ts
@@ -58,7 +58,7 @@ const formatJob = (job: QueueJob, queue: BaseAdapter): AppJob => {
     stacktrace,
     opts: jobProps.opts,
     data: queue.format('data', jobProps.data),
-    name: jobProps.name,
+    name: queue.getTitle(jobProps),
     returnValue: queue.format('returnValue', jobProps.returnvalue),
     isFailed: !!jobProps.failedReason || (Array.isArray(stacktrace) && stacktrace.length > 0),
   };
--- a/packages/api/src/queueAdapters/base.ts
+++ b/packages/api/src/queueAdapters/base.ts
@@ -4,16 +4,19 @@ import {
   JobStatus,
   QueueAdapterOptions,
   QueueJob,
+  QueueJobJson,
 } from '../../typings/app';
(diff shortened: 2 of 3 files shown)


felixmosh:
Thank you for the PR, we do have such functionality, review my [last comment ](https://github.com/felixmosh/bull-board/issues/349#issuecomment-988592807)

bugs181:
Cross posting from the issue; just in case the reasoning behind this PR solution versus the `formatters` pattern.

> Your proposed solution could work. Not sure what other implications there are of formatting the data, though. It might be tricky to evaluate where the data came from if it modifies the data object in the dashboard, however. With my proposed solution, and PR #358 - we know exactly how the formatting happens and that it's only purpose is for display.

bugs181:
Closing in favor of #359

PR #359, opened 2021-12-08
feat: add job name formatter

### LINKED ISSUE #349: Feature Request: Extract data from job to use as job Title

felixmosh:
closes #349

### DIFF
--- a/packages/api/src/handlers/queues.ts
+++ b/packages/api/src/handlers/queues.ts
@@ -58,7 +58,7 @@ const formatJob = (job: QueueJob, queue: BaseAdapter): AppJob => {
     stacktrace,
     opts: jobProps.opts,
     data: queue.format('data', jobProps.data),
-    name: jobProps.name,
+    name: queue.format('name', jobProps),
     returnValue: queue.format('returnValue', jobProps.returnvalue),
     isFailed: !!jobProps.failedReason || (Array.isArray(stacktrace) && stacktrace.length > 0),
   };
--- a/packages/api/src/queueAdapters/base.ts
+++ b/packages/api/src/queueAdapters/base.ts
@@ -1,4 +1,5 @@
 import {
+  FormatterField,
   JobCleanStatus,
   JobCounts,
   JobStatus,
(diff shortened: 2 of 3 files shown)


bugs181:
Definitely a lot cleaner than the solution I was working on!

bugs181:
Whoopsie. It appears that when no formatter is available, the whole `jobProps` object is passed into the `name` property. Unless I'm missing something.

bugs181:
Fix for this:
```js
const nameProp = queue.format('name', jobProps)

return {
    name: (typeof nameProp === 'string') ? nameProp : jobProps.name,
}
```

Not the cleanest solution, but keeps the UI from crashing on my end.

felixmosh:
Ha, you are correct :|

bugs181:
The above fix would also take into account of the user accidentally passing the wrong info back. The `name` prop will always be guaranteed to be a string.

felixmosh:
If a user passes a function, he "takes" responsibility to it's hand, no? ;]

bugs181:
That is true; but sanitization and type safety is never a bad thing. Why else do we use tools like TypeScript? ;)

felixmosh:
Yeah, I've thought about the same thing 😅
Actually, it is an easy thing to make TS check this
image

bugs181:
Fair points. I still recommend doing something other than passing the whole `jobProps` into the `name` prop. Just for future proofing code.

felixmosh:
I've used your suggestion, https://github.com/felixmosh/bull-board/commit/b30647767de50953b6acc9f8acfd65c67db5b1bb

bugs181:
Very nice. Thanks for all your hard work. I hope to help contribute more in the future. Hopefully without being a thorn in your side haha.

PR #384, opened 2022-02-18
Add canRetry options to queue

jgillick:
For some queues, I'd like to be able to clear jobs but never retry them. 

This PR adds a `allowRetries` option that defaults to `true` and is superseded by `readOnlyMode`.

**Example**
```ts
const boardQueues = [new BullMQAdapter(commsQueue, { allowRetries: false })];
createBullBoard({
    serverAdapter: new ExpressAdapter(),
    queues: boardQueues,
});
```

**Use Case**
I have a queue responsible for notifications that are only relevant for a short period of time and if they fail, should not be retried. The bull dashboard is available by many people inside my company and I don't want others to accidentally hit the retry button and cause confusion for our users. However, clearing jobs is an important feature to keep.

**readOnlyMode**
Setting the `readOnlyMode` option will supersede this option and set `allowRetries` to `false`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ const { setQueues, replaceQueues } = createBullBoard({
 ```
 
 2. `allowRetries` (default: `true`)
-Removes the retry job(s) buttons in the UI for a queue. This option will be ignored if `readOnlyMode` is `true`.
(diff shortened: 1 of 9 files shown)


jgillick:
@felixmosh That sounds good to me. I've updated the PR and added a couple unit tests.

felixmosh:
Released in `v3.10.0`
Thank you for this PR 🙏🏼

Issue #258, opened 2021-04-22
Feature: Dashboard page


felixmosh:
There are several requests for a single place that will summarize queues status, number of fails, etc...
Lets make a list of requirements from this page.

Design inspirations
![image](https://user-images.githubusercontent.com/9304194/115781404-fedfe100-a3c2-11eb-9b8f-5efa5627181a.png)

Deadly0:
@felixmosh exprected to see dashboard looks like it was in version 0.x where were list of all queues and task statuses in one page.

PengWeibin-Mars:
@felixmosh  Is it possible to use Chart.js (https://www.chartjs.org/docs/latest/samples/bar/vertical.html) as the UI of the dashboard

felixmosh:
> @felixmosh Is it possible to use Chart.js (https://www.chartjs.org/docs/latest/samples/bar/vertical.html) as the UI of the dashboard

Thank you for the suggestion, looks like a cool chart lib :]

PengWeibin-Mars:
Hey @felixmosh ,When is this feature expected to go live

felixmosh:
I don't have an estimation time, this is an open source project, You can contribute to it :]

Any PR regarding this feature is welcomed

Jivings:
I would prefer proper bar charts rather then wheels to show things like response/execution times and throughput during a day. 

I might have some time to work on this as I will definitely make use of it!

felixmosh:
Dashboard page should be split into 2 areas, a global stats & per queue stats.

The main purpose of this page should be a quick view of the entire queues. (maybe suitable to show on monitors).

wootwoot1234:
Are there any updates on this?  I would love to see some graphs of my jobs :)

felixmosh:
@wootwoot1234 actually I didn't started it yet... there are some other features that are in line.

Can you make a simple PR, and we will evolve it from there?

wootwoot1234:
@felixmosh sure, I would be happy to play with it a little. Can you help me get started by creating a branch with a new dashboard/metrics tab for each queue? Then I'll add the graphs.

You want to use chart.js? I don't personally have a favorite library for charts.

mrfrase3:
Just thought I might throw in my two cents, I like the way drone.io presents build stats:

![image](https://user-images.githubusercontent.com/3320430/200717888-c3ea22a0-e75f-4ef8-b765-9ba07323cba7.png)

[...]

PR #394, opened 2022-04-01
Added Queue Status to Nav

Slind14:
image

### DIFF
--- a/packages/ui/src/components/Menu/Menu.module.css
+++ b/packages/ui/src/components/Menu/Menu.module.css
@@ -108,3 +108,10 @@
   border: 2px solid #4abec7;
   outline: 0;
 }
+
+.navLi {
+  display: flex;
+}
+.navLi a {
+  width: 196px;
+}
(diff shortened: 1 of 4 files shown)


felixmosh:
Thank you for the PR, we had many ideas on how to make these stats available, the final decision come to a Dashboard page with all of these stats #258.

I will really appreciate any help with it.

Slind14:
We really would like this as part of the nav for UX.

felixmosh:
The UX is not good.
I get that you need this info, but it makes the UI very clattered (a lot of noise), 
image
It is not clear what each number represents.

Slind14:
I see. I suppose we need to work with the fork then.

wyantb:
@felixmosh perhaps this could be defaulted to false, but available behind a configuration option passed to `createBullBoard` or similar? Call it `showsStatisticsInNavigation` or some such. I can empathize with your concern that this may not be ideal for most users, but speaking to my own operational needs this would be perfect.

PR #395, opened 2022-04-01
Queue Backlog, Waiting and Processing Percentile Stats

Slind14:
image


felixmosh:
Thank u for this awesome PR 🙏🏼

I think that this make the UI bulky and steals focus from the main purpose of the page - to manage the a Queue.
In my opinions such stats should be on a dashboard page, #258, a one place to show stats & info regarding all queus.

Let's work on it together.

Slind14:
Thank you. I'm afraid I don't want to invest any additional time on this since the issues we were facing with missing information for fine-tuning queues are resolved.

Slind14:
I moved it to the completed tab, since it is only relevant with that data.

PR #396, opened 2022-04-02
Made Card Scrollbar useable, changed style of others to match

Slind14:


felixmosh:
Can you share a screen with the problem & the solution pls?

Slind14:
Before there was the big nav bar with a fading effect overlaying. This effect caused it to be unclickable. 
This is replaced with a scrollbar that only shows when hovering the scrollable element now.

chrome_MscjAjBnGD

The bulky device one in the sidebar nav was adjusted as well to take up less space and match the style:

image
(mostly for windows)

felixmosh:
Thank you for the PR, but custom scroll bar is a big no-no for me.
This lib size is ~70k just for this.

Due to #399, I'll just remove the fade-out effect since there is no way in CSS to detect that scroll bar appeared.

Issue #413, opened 2022-05-05
Truncate large size job's payload


maxime-lenne:
Hi,

I have an issue with bullboard + fastify on queues that have a  large size job's payload. The queue page failed with a timeout.
To prevent this issue it's possible to truncate large job's payload object (like sidekiq in ruby).

I propose a PR maybe it's not the good way fore the project but tell me or create a new one to resolve this issue.

Thanks.

PR #414, opened 2022-05-05
Add truncate for large job's payload

maxime-lenne:
for this issue #413

### DIFF
--- a/packages/api/src/handlers/queues.ts
+++ b/packages/api/src/handlers/queues.ts
@@ -41,6 +41,38 @@ const getStats = async (queue: BaseAdapter): Promise<ValidMetrics> => {
   return validMetrics;
 };
 
+const truncate = function (dataToTruncate: any, curDepth = 0): any {
+  if (dataToTruncate === undefined) {
+    return undefined;
+  }
+  if (dataToTruncate === null) {
+    return null;
+  }
+  if (curDepth < 4) {
+    const newDepth = curDepth + 1
+    const json: any = Object.assign({}, dataToTruncate);
+    Object.entries(dataToTruncate).forEach(([key, value]) => {
+      switch (typeof value) {
+        case "string":
+          json[key] = value.length > 10000 ? "[String-Truncated]" : value;
+          break;
+        case "object":
+          if (Array.isArray(value)) {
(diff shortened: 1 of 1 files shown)


felixmosh:
Thank you for this PR 🙏🏼
Can you add some images / payload example of the problem & the solution?

maxime-lenne:
Hi,

[large_payload.txt](https://github.com/felixmosh/bull-board/files/8638466/large_payload.txt)
Capture d’écran 2022-05-06 à 09 41 22
Capture d’écran 2022-05-06 à 09 39 23

I have proposed this solution :

- limit string and array value to 10K characters
- limit the objects depth to 4
- up to this two limit replace the value of the json key with string : "[String-Truncated]", "[Array-Truncated]" or "[Object-Truncated]"

Capture d’écran 2022-05-06 à 09 30 24

packages/api/src/handlers/queues.ts
```diff
@@ -41,6 +41,38 @@ const getStats = async (queue: BaseAdapter): Promise<ValidMetrics> => {
...
+  if (dataToTruncate === undefined) {
+    return undefined;
+  }
+  if (dataToTruncate === null) {
+    return null;
+  }
+  if (curDepth < 4) {
+    const newDepth = curDepth + 1
+    const json: any = Object.assign({}, dataToTruncate);
+    Object.entries(dataToTruncate).forEach(([key, value]) => {
```
felixmosh, line 54: There is type change when value is 
1. `string` => 
 ```js 
 truncate('some-string') /* returns {
    "0": "s",
    "1": "o",
    "2": "m",
    "3": "e",
    "4": "-",
    "5": "s",
    "6": "t",
    "7": "r",
    "8": "i",
    "9": "n",
    "10": "g" */
}
```
3. `array` =>
 ```js 
 truncate([1,2,3]) /* returns {
    "0": 1,
    "1": 2,
    "2": 3
} */
```

felixmosh (changes requested):

maxime-lenne:
@felixmosh Thanks for this better way! I understand why you don't will to "hard coded" this.
May be document this formater feature for people that need like me to manipulate this fields and/or create an example. I try to do that when I'm less busy ;)

PR #421, opened 2022-06-10
chore: upgrade to fastify 4

### LINKED ISSUE #418: Fastify V4 Support
Fastify V4 is GA now, would be nice to have support 🥇

kevcodez:
Fastify 4 is generally available

Resolves #418 

FYI: Fastify 4 dropped support for Node 12 and added support for Node 18


felixmosh:
Hi, do you know if it backwards compatible?
Meaning if someone will install v3 with bull-board, will it work?

kevcodez:
AFAIK it's not backwards compatible. As far as I can tell, the fastify view plugin (previously point-of-view) had breaking changes, same goes for static (requires 4.x)

felixmosh:
Thank you for this PR.

PR #429, opened 2022-07-14
Optional chain queue find to fix blank dashboard

theoilie:
I sometimes encounter this error when going to the path for a valid queue. The screen is blank, and this is what the Console shows:
Screen Shot 2022-07-13 at 8 35 21 PM

Note that this is different from when I enter an invalid path -- in that scenario I see the expected "Queue Not found" message.

When looking into the stack trace, it was from the line I changed in this PR. Since the error was on `queues.find`, optional chaining fixes this until the state updates and the queue can be found.

### DIFF
--- a/packages/ui/src/components/App.tsx
+++ b/packages/ui/src/components/App.tsx
@@ -28,11 +28,11 @@ export const App = ({ api }: { api: Api }) => {
                   path="/queue/:name"
                   render={({ match: { params } }) => {
                     const currentQueueName = decodeURIComponent(params.name);
-                    const queue = state.data?.queues.find((q) => q.name === currentQueueName);
+                    const queue = state.data?.queues?.find((q) => q.name === currentQueueName);
 
                     return (
                       <QueuePage
-                        queue={queue}
+                        queue={queue || undefined}
                         actions={actions}
                         selectedStatus={selectedStatuses}
                       />


felixmosh:
Hi, thank you for this PR 🙏🏼

Can you elaborate on when this can happened? How can I reproduce the issue? 
I'm working with the board on a daily basis, and I didn't encounter such error.

theoilie:
> Hi, thank you for this PR 🙏🏼
> 
> Can you elaborate on when this can happened? How can I reproduce the issue? I'm working with the board on a daily basis, and I didn't encounter such error.

Hi! It seems like some kind of race condition so I can't provide any steps to reliably reproduce it. I do notice it [here](https://creatornode2.audius.co/health/bull/queue/state-monitoring-queue) fairly often, though. A couple other notes about the bug:
- I've mostly (maybe only) ever noticed it happen to the first queue in the list
- I'm able to still view Bull Board if I go to another queue.
- Sometimes I'm able to switch back to the first queue, but usually it runs into the same error again. See example video below (first/broken queue is `state-monitoring-queue` in the video, but notice how it's able to display just fine after I switch to other queues and then back to `state-monitoring-queue`)
- It could possibly have to do with truncating if maybe the queue isn't added to the list of queues until truncation logic is performed? Either way, this optional chaining should resolve the issue 😃 

https://user-images.githubusercontent.com/4657956/179028051-9d90914f-98da-4286-9f7d-fd6ca9a2def7.mov

felixmosh:
Thank you for the explanation, can you check what is the api response when you get the exception, specifically the queues array?
I'm not able to reproduce it on your site.

theoilie:
Ah I see that now -- didn't even notice the rest of that stack trace before. It looks like that's an issue in my truncation logic where I checked if something was `typeof` object and then did `Object.keys()` on it, but `null` technically has the type object as well so `Object.keys(null)` throws. Will fix on my end, thanks!

felixmosh:
Your change was released in `v4.0.1` (no breaking change if you are not using fastify)

PR #439, opened 2022-08-02
Add pin job functionality

### LINKED ISSUE #424: Feature: Pin job
It's currently difficult with fast-moving queues to view a job's data for more than 10 or 20 seconds because new jobs come in and push the job I'm looking at out of view. I saw another request about slowing down the refresh rate, but the ability to pin a job for viewing purposes would also be very helpful. Is that a planned feature or possibility? Thanks!

theoilie:
- Closes #424 
- Adds a new "Pin" button to the right of each job's ID
- Click the Pin button to toggle pinning the job
- Pinned jobs stay at the top of the page and have a border and highlighted Pin button to differentiate them
- Pinned jobs maintain their order (sorted by ID)

https://user-images.githubusercontent.com/4657956/182306289-11e8a847-f6a0-49b3-88d8-934153c76b44.mov


felixmosh:
Hi @theoilie thank you for the PR.

Few concerns, 
1. Does it apply the pinning for all job statuses or only for the `latest` one?
2. What happens if the job finished & it pinned?
3. The place that you've put the button won't work with long ids (this is the default), maybe it is better to put to button with the job action button? WDYT?

theoilie:
Hey @felixmosh thanks for your feedback --

1. The typical flow for me is to look at completed jobs so I'd only pin in that status, but currently this pins across all statuses within the same queue.
2. It will stay pinned at the top no matter what the status is or what status tab you're viewing. So if it finishes it doesn't move away from the top.
3. I can move the button elsewhere, but making it as an action prevents using it in readOnly mode. In my use case we use readOnly but we still want to be able to pin a job because it's just visual and doesn't actually change the job. What if I put it as a new row/div below the job ID instead of to the side of it? (see video below)

Updated demo (I also fixed a sorting bug after taking this demo, so please ignore that part -- it's correctly sorting by `processedOn` now):

https://user-images.githubusercontent.com/4657956/182493195-ea96d112-3b35-4a23-9ca0-2a6c19693697.mov

felixmosh:
Thank you for the update.

Do you know what will happen if the job has `removeOnComplete` flag?

theoilie:
> Thank you for the update.
> 
> Do you know what will happen if the job has `removeOnComplete` flag?

Good question -- supporting this directly would've been messy with supporting the `KeepOptions` flag and others, so I pushed a change to go the more generic route and only show jobs in the Status that is currently in view. This means that now `removeOnComplete` and `removeOnFail` will be respected, and changing Status tabs will also unpin jobs. This should be fine because the typical use case for pinning a job is only to view it in that same Status tab.

PR #459, opened 2022-10-05
Add option to overwrite the ui base path

jRexhmati:
…dapter app

### DIFF
--- a/packages/koa/src/KoaAdapter.ts
+++ b/packages/koa/src/KoaAdapter.ts
@@ -15,15 +15,22 @@ import views from 'koa-views';
 
 export class KoaAdapter implements IServerAdapter {
   private basePath = '';
+  private uiBasePath = '';
   private bullBoardQueues: BullBoardQueues | undefined;
   private errorHandler: ((error: Error) => ControllerHandlerReturnType) | undefined;
   private statics: { path: string; route: string } | undefined;
   private viewPath: string | undefined;
   private entryRoute: AppViewRoute | undefined;
   private apiRoutes: AppControllerRoute[] | undefined;
+  private app: Koa | undefined;
 
   public setBasePath(path: string): KoaAdapter {
     this.basePath = path;
+    return this;
+  }
+
+  public setUiBasePath(path: string): KoaAdapter {
+    this.uiBasePath = path;
     return this;
   }
 
@@ -60,7 +67,7 @@ export class KoaAdapter implements IServerAdapter {
     return this;
   }
 
-  public registerPlugin() {
+  public setupApp() {
     if (!this.statics) {
       throw new Error(`Please call 'setStaticPath' before using 'registerPlugin'`);
     } else if (!this.entryRoute) {
@@ -75,12 +82,12 @@ export class KoaAdapter implements IServerAdapter {
       throw new Error(`Please call 'setErrorHandler' before using 'registerPlugin'`);
     }
 
-    const app = new Koa();
+    this.app = new Koa();
     const router = new Router({
       strict: true,
     });
 
-    app.use(async (ctx, next) => {
+    this.app.use(async (ctx, next) => {
       try {
         await next();
       } catch (err) {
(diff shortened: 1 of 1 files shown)


felixmosh:
What is ui-base-path? The path to the view template?

jRexhmati:
Yes, the path of the public side


Currently fails to load the assets if the basePath is /queues
But you want to mount it to /admin/queues

jRexhmati:
The plain problem is that I want to add a middleware for the entire dashboard only (Not application-wide)

To do that I have this setup:

```const serverAdapter = new KoaAdapter();

  createBullBoard({
    queues,
    serverAdapter,
  });

  // Provisional app to add auth middleware
  const _app = new Koa();
  _app.use(Middlewares.centralizedAuthorizedUser);

  // Attach bullboard to provisional app with the auth middleware
  serverAdapter.setBasePath('/');
  _app.use(serverAdapter.registerPlugin());

  // Attach provisional app to the main app with bullboard inside
  app.use(mount('/queue-dashboard', _app));
```


You can see the problem here.

The APIs will be available at /queue-dashboard
But the assets will be accessed at `localhost/static/...`

felixmosh:
Can you try to set serverAdapter.setBasePath('/queue-dashboard');

?

jRexhmati:
In that case, to still retain the same full URL ('/queue-dashboard')
I have to update the script this way 

```

  createBullBoard({
    queues,
    serverAdapter,
  });

  // Provisional app to add auth middleware
  const _app = new Koa();
  _app.use(Middlewares.centralizedAuthorizedUser);

  // Attach bullboard to provisional app with the auth middleware
  serverAdapter.setBasePath('/queue-dashboard');
  _app.use(serverAdapter.registerPlugin());

  // Attach provisional app to the main app with bullboard inside
  app.use(mount('/', _app));
```

But now the auth middleware is global for all the routes at the root level

felixmosh:
setBasePath does what you need, to allow mointing on sub path it should be aligned with the mount path.

keep your original code (which mounts the app under the subpath) & set the basepath...
It should work.

If no, this is a bug. Check the express / fastify examples, the mount point is always alined yhe the base path.

jRexhmati:
I think this is just a case which has not been thought of

Let me turn it around.

How would you mount the dashboard on a sub-route?


Because you are using the basePath as a mounting point 
https://github.com/felixmosh/bull-board/blob/master/packages/koa/src/KoaAdapter.ts#L133

And also as the base path of the UI
https://github.com/felixmosh/bull-board/blob/master/packages/koa/src/KoaAdapter.ts#L109

While that's not always the case, the mounting point can have prefixes (Like my example)

felixmosh:
OK, I think that I get the problem, let me check what can be done.
Can you open an issue with explanation & example of how to reproduce (you can use the koa example)

jRexhmati:
Sure, thanks

jRexhmati:
Besides the uiBasePath might be a good idea to apply the open closed principle in the Adapter class for the app instance

felixmosh:
I'm not using KOA, so not familiar with it much, maybe the issue is using publicPath in the mount? if we remove it from there, basePath will work as it works for rest of the frameworks.

WDYT?

jRexhmati:
I meant to use the [Open-Closed](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle) principle from SOLID
Changing the app to a property so other users can extend the class if they want to expand the ServerAdapter logic

Another option is to separate registering the app and mounting the map, but this change will not be backward compatible

jRexhmati:
I think you can do both, pushed some new changes to demonstrate my point, this does not have to be the end solution but you get the idea.

Make few extendable decisions when setting up the class so other users can extend the class and override what they need 

Also access to the app instance is important I think

felixmosh:
This PR contains 2 separate improvements, I prefer that it will be split into 2 separate PRs

jRexhmati:
@felixmosh Updated this PR to only allow uiBasePath modification and a fallback to original logic if uiBasePath is missing

jRexhmati:
This does work for me. However in regards to the way how you are giving the option to control the UI and the mounting point, I'd say it is not 100% explicit

felixmosh:
I prefer this kind of solution.
It is available on `v4.4.0`

PR #507, opened 2022-12-27
feat: Allow retrying completed jobs when using BullMQ

### LINKED ISSUE #506: Feature: Retry completed jobs
The bullmq retry method accepts either `failed` or `completed` as the state but bull-board only allows retrying jobs in the `failed` state.

I feel like this should be a fairly minimal change that would help a tone in certain cases. Like for example debugging jobs that you expect to fail but they don't.

dumbasPL:
This PR adds support for retrying completed jobs when using the BullMQ adapter. Retrying completed jobs is not possible in bull, only in BullMq.

closes #506

### DIFF
--- a/packages/api/src/handlers/queues.ts
+++ b/packages/api/src/handlers/queues.ts
@@ -11,7 +11,6 @@ import {
 } from '../../typings/app';
 import { STATUSES } from '../constants/statuses';
 import { BaseAdapter } from '../queueAdapters/base';
-import { BullMQAdapter } from '../queueAdapters/bullMQ';
 
 const formatJob = (job: QueueJob, queue: BaseAdapter): AppJob => {
   const jobProps = job.toJSON();
@@ -96,7 +95,7 @@ async function getAppQueues(
         pagination,
         readOnlyMode: queue.readOnlyMode,
         allowRetries: queue.allowRetries,
-        allowCompletedRetries: queue instanceof BullMQAdapter,
+        allowCompletedRetries: queue.allowCompletedRetries,
         isPaused,
       };
     })
--- a/packages/api/src/handlers/queues.ts
+++ b/packages/api/src/handlers/queues.ts
@@ -11,6 +11,7 @@ import {
 } from '../../typings/app';
 import { STATUSES } from '../constants/statuses';
 import { BaseAdapter } from '../queueAdapters/base';
+import { BullMQAdapter } from '../queueAdapters/bullMQ';
 
 const formatJob = (job: QueueJob, queue: BaseAdapter): AppJob => {
   const jobProps = job.toJSON();
@@ -95,6 +96,7 @@ async function getAppQueues(
         pagination,
         readOnlyMode: queue.readOnlyMode,
         allowRetries: queue.allowRetries,
+        allowCompletedRetries: queue instanceof BullMQAdapter,
         isPaused,
       };
     })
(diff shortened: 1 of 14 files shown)


felixmosh:
Hi @dumbasPL 
Can we make it possible to fake retries of completed jobs for bull as well?

packages/api/src/handlers/queues.ts
```diff
@@ -95,6 +96,7 @@ async function getAppQueues(
         pagination,
         readOnlyMode: queue.readOnlyMode,
         allowRetries: queue.allowRetries,
+        allowCompletedRetries: queue instanceof BullMQAdapter,
```
felixmosh, line 99: You don't need to check the type of the adapter, this field can be part of a specific adapter.
maybe put a default value on the Base class & override it when needed.

packages/ui/src/components/JobCard/JobActions/JobActions.tsx
```diff
@@ -43,7 +45,7 @@ export const JobActions = ({ actions, status, allowRetries }: JobActionsProps) =
     return null;
   }
   if (!allowRetries) {
-    buttons = buttons.filter((btn) => btn.actionKey !== 'retryJob');
+    buttons = buttons.filter((btn) => btn.actionKey !== 'retryFailedJob' && btn.actionKey !== 'retryCompletedJob');
```
felixmosh, line 48: Let's extract it to a function.
```js
function isRetry(actionKey) {
  return ['retryFailedJob', 'retryCompletedJob'].includes(actionKey);
}

...

if (!allowRetries) {
  buttons = buttons.filter((btn) => !isRetry(btn.actionKey));
}
```

packages/ui/src/hooks/useStore.ts
```diff
@@ -158,8 +172,10 @@ export const useStore = (): Store => {
     state,
     actions: {
       promoteJob,
-      retryJob,
-      retryAll,
+      retryFailedJob,
```
felixmosh, line 175: Instead of forking by type of the status, can we pass the status as a parameter?
This means that there will be 2 retry functions which will get `queueName` & `status` as parameters, and pass it to the proper api, it can be query-string.
It can eliminate the need for a separate api end point.

dumbasPL: done

dumbasPL:
Changes implemented, please review.

> Hi @dumbasPL Can we make it possible to fake retries of completed jobs for bull as well?

Do we want to do that? So far this project wasn't trying to do things that aren't already natively supported. Doing this without making it obvious to the user what's actually going to happen might introduce unnecessary confusion and a lot of potential bugs.

How would we even go about doing this? When retrying jobs stuff like the job id remains constant. How about user-defined job IDs? Should we remove the job before adding a new one? How about repeated jobs? How about the attempts counter?

Personally, I think this is a terrible idea. A lot of ambiguity and unexpected behavior.

packages/ui/src/hooks/useStore.ts
```diff
@@ -102,10 +102,10 @@ export const useStore = (): Store => {
       confirmJobActions
     );
 
-  const retryAll = (queueName: string) =>
+  const retryAll = (queueName: string, status: JobFinishedStatus) =>
     withConfirmAndUpdate(
-      () => api.retryAll(queueName),
-      'Are you sure that you want to retry all jobs?',
+      () => api.retryAll(queueName, status),
+      'Are you sure that you want to retry all failed jobs?',
```
felixmosh, line 108: The message is not modified by the status, it can be `failed` or `completed`.

dumbasPL: That's why I initially implemented it using two separate functions. `cleanAll` already has 3 variants for `failed`, `completed`, and `delayed` so I assumed this approach would be fine for retries as well

felixmosh:
Thank you for this PR

felixmosh:
Released in `v4.10.0`

PR #521, opened 2023-01-14
feat(): dark mode for bull board experimental 

luluhoc:
Some changes are yet to be made and test if all components are in dark mode.

Result of some fast late night coding ;)

image

image


luluhoc:
@felixmosh

felixmosh:
Hi, thank you for the initiative...
Why did you changed cx lib with classnames?

I think that a better approach is using css custom props, which means you set one class which changes all colors...

PR #532, opened 2023-02-07
fix: incorrect value of job name when job name is undefined

CodyTseng:
## Description

https://github.com/felixmosh/bull-board/blob/9c17cec31e296787fe0d8b53f0ae27806d6b2c0d/packages/api/src/handlers/queues.ts#L32

When the `jobProps.name` is undefined, the default value of `BaseAdapter.format(field, data, defaultValue = data)` will be `jobProps`. And then React will throw `Error: Minified React error #31;`.

### DIFF
--- a/packages/api/src/handlers/queues.ts
+++ b/packages/api/src/handlers/queues.ts
@@ -29,7 +29,7 @@ const formatJob = (job: QueueJob, queue: BaseAdapter): AppJob => {
     stacktrace,
     opts: jobProps.opts,
     data: queue.format('data', jobProps.data),
-    name: queue.format('name', jobProps, jobProps.name),
+    name: queue.format('name', jobProps, jobProps.name || ''),
     returnValue: queue.format('returnValue', jobProps.returnvalue),
     isFailed: !!jobProps.failedReason || (Array.isArray(stacktrace) && stacktrace.length > 0),
   };


packages/api/src/handlers/queues.ts
```diff
@@ -29,7 +29,7 @@ const formatJob = (job: QueueJob, queue: BaseAdapter): AppJob => {
     stacktrace,
     opts: jobProps.opts,
     data: queue.format('data', jobProps.data),
-    name: queue.format('name', jobProps, jobProps.name),
+    name: queue.format('name', jobProps, jobProps.name || ''),
```
felixmosh, line 32: It is better to make a default value outside of the format it self

```
queue.format('name', jobProps, jobProps.name) || '',
```

CodyTseng: When the `jobProps.name` is undefined, the `format()` will return `jobProps`. So, we can't make a default value outside of the `format`.

CodyTseng: I think this might be good too:

```typescript
queue.format('name', jobProps.name, '')
```

felixmosh:
Thank you @CodyTseng for the PR.

I will release it in few mins

CodyTseng:
I took a closer look and I think `queue.format('name', jobProps.name)` is better.

Because the second parameter will be passed to the `setFormatter`. From the documentation, this value should be `jobProps.name`. 

image

felixmosh:
The docs are wrong :|
There is a use-case that some users want job name to be consist from other job props, therefore, We pass the entire jobProps to the formatter.
Your solution is enough :].

PR #559, opened 2023-04-21
Menu : hilight queue name in red if failed count > 0

Ayora29:
Currently if we want to see if there is any failed job, we need to click each queue.
Here is a quick modification to write queue name in red if there is any failed job.
Tested and working.

### DIFF
--- a/packages/ui/src/components/Menu/Menu.tsx
+++ b/packages/ui/src/components/Menu/Menu.tsx
@@ -47,7 +47,7 @@ export const Menu = ({
                     }`}
                     activeClassName={s.active}
                     title={queueName}
-                    style={{ color: counts['counts'] ? 'red' : 'inherit' }}
+                    style={{ color: counts['failed'] ? 'red' : 'inherit' }}
                   >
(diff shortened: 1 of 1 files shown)


felixmosh:
Hi @Ayora29 thank you for this PR.
I think that instead of trying to bypass the requirement, we should focus on the #258 task...

What if i want to see it there are waiting jibs in the queue, your solution (coloring the queue name) won't help us.

What do you think, can you start this task?

Ayora29:
Hello @felixmosh, 
Yes I saw this thread concerning a dashboard.
I honestly do not have time to start working on another task sorry.
While dashboard feature is not a reality yet, this is just a quick modification in the meantime. You could remove it once the dashboard is ready.
I saw other people asking for this highlight as well, this is why I'm proposing a PR.

felixmosh:
I don't want the add a feature that is a temp solution....

felixmosh:
I will review a solution that will show colors on the nav element, and it will be enabled with a settings...
Maybe this will help as well

felixmosh:
@Ayora29 check https://github.com/felixmosh/bull-board/issues/258#issuecomment-1519128654

Issue #564, opened 2023-04-30
Undeclared peer dependency between @bull-board/api and @bull-board/ui breaks with yarn's PnP system


JasonMan34:
Yarn's [PnP system](https://yarnpkg.com/features/pnp) considers accessing the `@bull-board/ui` package without listing it as a dependency dangerous, so it fails to run [(repro here)](https://github.com/JasonMan34/bull-board-yarn-dep-bug/tree/main)

This causes the `@bull-board/express` package to break when using yarn with PnP:
```
Error: @bull-board/api tried to access @bull-board/ui, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

Required package: @bull-board/ui (via "@bull-board/ui/package.json")
Required by: @bull-board/api@npm:5.1.1 (via /Users/user/repos/bull-board-yarn-dep-bug/.yarn/cache/@bull-board-api-npm-5.1.1-533fa06b5d-b4570ef330.zip/node_modules/@bull-board/api/dist/src/)
```

Listing `@bull-board/ui` is a peer dependency or `@bull-board/api` should resolve this issue

PR #565, opened 2023-04-30
Add @bull-board/ui as peerDep of @bull-board/api

JasonMan34:
This closes https://github.com/felixmosh/bull-board/issues/564

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -36,6 +36,9 @@
     "bullmq": "^1.80.6",
     "ioredis-mock": "^7.2.0"
   },
+  "peerDependencies": {
+    "@bull-board/ui": "5.1.1"
+  },
   "publishConfig": {
     "access": "public"
   }
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -37,7 +37,7 @@
     "ioredis-mock": "^7.2.0"
   },
   "peerDependencies": {
-    "@bull-board/ui": "5.1.1"
+    "@bull-board/ui": "*"
   },
   "publishConfig": {
     "access": "public"


packages/api/package.json
```diff
@@ -36,6 +36,9 @@
     "bullmq": "^1.80.6",
     "ioredis-mock": "^7.2.0"
   },
+  "peerDependencies": {
+    "@bull-board/ui": "5.1.1"
```
felixmosh, line 40: lets not restrict the version here, make it `*`

JasonMan34: Wasn't sure about compatibility issues so I copied the version from the inverse dependency (ui's dependany on api@5.1.1)

Changed to *

felixmosh:
Released in `v5.1.2`, can you try this version?

JasonMan34:
Works now :)

Although the solution feels off to me. `ui`'s dependency of `api` is obvious, but the reverse dependency only exists because of this piece of code:
```ts
export function createBullBoard({
  queues,
  serverAdapter,
  options = { uiConfig: {} },
}: {
  queues: ReadonlyArray<BaseAdapter>;
  serverAdapter: IServerAdapter;
  options?: BoardOptions;
}) {
  const { bullBoardQueues, setQueues, replaceQueues, addQueue, removeQueue } = getQueuesApi(queues);
  const uiBasePath = path.dirname(eval(`require.resolve('@bull-board/ui/package.json')`));

  serverAdapter
    .setQueues(bullBoardQueues)
    .setViewsPath(path.join(uiBasePath, 'dist'))
    .setStaticPath('/static', path.join(uiBasePath, 'dist/static'))
    .setUIConfig(options.uiConfig)
    .setEntryRoute(appRoutes.entryPoint)
    .setErrorHandler(errorHandler)
    .setApiRoutes(appRoutes.api);

  return { setQueues, replaceQueues, addQueue, removeQueue };
}
```

Perhaps if `createBullBoard` is inherently coupled with `@bull-board/ui`, it should be exported from there? That would remove the cyclic dependency and resolve the issue (without declaring a peerDep)  
Alternatively, perhaps `createBullBoard` should be decoupled from `@bull-board/ui` entirely, and accept another required parameter `uiBasePath`. Then `@bull-board/ui` could export a basePath to be used when calling `createBullBoard`

felixmosh:
Thanx for the feedback.

1. `createBullBoard` is a builder of api, it should be on the server side part.
2. I've thought about it, but then each user of the lib will need to pass the same piece of code
3. Good possible solution, I'll consider this refactor, thanx

PR #577, opened 2023-06-06
feat: make whole QueueCard clickable

nicokunz:
Thanks for a great library. I have extended the new Queue Cards by making the whole card clickable rather than just the queue name.


felixmosh:
Released in `5.2.1`

PR #584, opened 2023-06-16
feat: Allow changing the page title

goferito:
Allows changing the page title.

### DIFF
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -21,6 +21,7 @@ export function createBullBoard({
     .setQueues(bullBoardQueues)
     .setViewsPath(path.join(uiBasePath, 'dist'))
     .setStaticPath('/static', path.join(uiBasePath, 'dist/static'))
+    .setPageTitle(options.pageTitle)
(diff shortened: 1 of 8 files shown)


goferito:
@felixmosh The distribution method is a bit new to me. Could you point me in the right direction on how I could use my bull-board fork in my production project? Is it possible to add the dependency of each package from github? Or do I need to publish to my npm account each package (express/ui)? 

I tested it using the example.ts provided and seems to pass your validations. Let me know if I can improve something. If we can merge this, I can do the same with the favicon, so the bull board can integrate better into existing projects.

Cheers!

goferito:
If you like the solution I will also update the README with the new feature.

goferito:
@felixmosh  Thanks for the feature. What about the favicon? Could you add support to change that as well? I can give it a try if you want. Maybe also allow the boardTitle to different than the pageTitle? (that was my attempt with my PR.

goferito:
I'd love to make a useful contribution. I don't think I can allocate time for this today, but I will take care of it as soon as I have some free time. Thank you!

PR #591, opened 2023-06-21
Add globalPrefix option to BullBoardModuleOptions

SimonRosenau:
This PR fixes an issue when using ```app.setGlobalPrefix('/api/v1')``` in NestJS. When only using the `route` option, specifying  `/queues` as the route, the index page would properly load at `/api/v1/queues`, but all assets would be loading from `/queues/...` instead of `/api/v1/queues/...` which would then fail.

### DIFF
--- a/packages/nestjs/src/bull-board.root-module.ts
+++ b/packages/nestjs/src/bull-board.root-module.ts
@@ -16,7 +16,7 @@ export class BullBoardRootModule implements NestModule {
   }
 
   configure(consumer: MiddlewareConsumer): any {
-    this.adapter.setBasePath(this.options.route);
+    this.adapter.setBasePath(this.options.globalPrefix ? `${this.options.globalPrefix}${this.options.route}` : this.options.route);
 
     if (isExpressAdapter(this.adapter)) {
       return consumer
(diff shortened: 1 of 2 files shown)


felixmosh:
Hi, thank you for your contribution.
I'm not sure why do you need this additional parameter, if you concat it? Why not change the route, to contain the entire prefix? (Noob question)

felixmosh:
Hi, thank you for your explanation, I'm not sure, can we derive (get) this global value somehow? It doesn't make sense that you as a consumer need to know about the usage of this globalPrefix.

DennisSnijder:
@felixmosh Hmm, putting it as an option feels like a double declaration (since you already declared it in your NestJS app), It would be nicer to have the module to automatically recognise the global prefix.

Let me check if that's actually possible...

DennisSnijder:
@felixmosh @SimonRosenau seems like I can grab the `globalPrefix` variable in NestJS directly from `ApplicationConfig`.
I can't edit @SimonRosenau his PR, let me open a new one.

Thank you for addressing this issue! @SimonRosenau !

SimonRosenau:
@DennisSnijder Thanks for taking this over! 🙌

PR #592, opened 2023-06-23
[NestJS] Global prefix support

DennisSnijder:
## Context
See #591

## What has been done
- Checked for the existence of a global prefix in the ApplicationConfig from NestJS


DennisSnijder:
@felixmosh ready for review 😋

felixmosh:
closes #591

packages/nestjs/src/bull-board.root-module.ts
```diff
@@ -2,21 +2,27 @@ import { DynamicModule, Inject, MiddlewareConsumer, Module, NestModule, Provider
...
     private readonly adapterHost: HttpAdapterHost,
+    private readonly applicationConfig: ApplicationConfig,
     @Inject(BULL_BOARD_ADAPTER) private readonly adapter: BullBoardServerAdapter,
     @Inject(BULL_BOARD_OPTIONS) private readonly options: BullBoardModuleOptions
   ) {
   }
 
   configure(consumer: MiddlewareConsumer): any {
-    this.adapter.setBasePath(this.options.route);
+
```
felixmosh, line 20: I prefer it to be something like this:
```js
const globalPrefix = this.applicationConfig.getGlobalPrefix() || '';
this.adapter.setBasePath(`${ globalPrefix }${ this.options.route }`);
```

DennisSnijder: Good one!

felixmosh:
closes #591

packages/nestjs/src/bull-board.root-module.ts
DennisSnijder: Done!

felixmosh:
Thank you @DennisSnijder & @SimonRosenau 🙏🏼
Released in `v5.5.2`

piotr-skowronski-prpl:
@DennisSnijder It looks like there's still a bug here.
When setting the global prefix without the `/` at the beginning, like so `app.setGlobalPrefix('api');`, assets are taken from `/api/api/queues`.
And the official NestJS documentation does not use the leading `/` in the global prefix section so people will be tripping over this.

DennisSnijder:
> @DennisSnijder It looks like there's still a bug here. When setting the global prefix without the `/` at the beginning, like so `app.setGlobalPrefix('api');`, assets are taken from `/api/api/queues` which doesn't exist, so it fails. And the official NestJS documentation does not use the leading `/` in the global prefix section so people will be tripping over this.

Good catch! i'll take a look into it later today!

PR #598, opened 2023-07-05
single job page view

jamesgweber:
Greetings - my team is using bull-board for a project and the need arose for having the ability to deep link into a failed job.
This is meant to be a first pass to see what you think & get some feedback. It lacks completeness as I still don't totally understand all the interaction patterns for paused / delayed jobs but I feel its headed in the right direction.

Kapture 2023-07-05 at 08 16 57


packages/ui/src/pages/JobPage/JobPage.tsx
```diff
@@ -0,0 +1,123 @@
...
+    await actions.cleanJob(queue?.name)(job)();
+    history.push(`/queue/${queue.name}`);
+  };
+
+  useEffect(() => {
+    if (!queue || !job) return;
+
+    if (jobState === 'active') {
+      // transition from active to completed
+      if (job?.progress === 100 && !job.failedReason) {
```
jamesgweber, line 60: I haven't been able to figure out how to interrogate a job to see which state queue it's in at any given time. If we had that info much of this complexity could go away 🔥

packages/ui/src/pages/JobPage/JobPage.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const fetchJob = async () => {
+    const jobResponse = await actions.getJob(name)(jobId)();
+    setJob(jobResponse[jobId]);
+  };
+
+  const pollForJob = () => {
+    if (!!intervalRef.current) return;
+    fetchJob();
+
+    intervalRef.current = setInterval(() => {
```
felixmosh, line 35: check out the useInterval hook

felixmosh (changes requested):

felixmosh:
Thank you for this PR.... 
I'm not sure why would you implement an entire page for this scenario?

It sounds to me as a filter on the requested queue, you can add a querystring which will make queue list return the specific job in the list... this way, you don't need to re-implement 80% of the code.

WDYT?

jamesgweber:
Interesting - to me a router pattern for `/queues/:queue/:jobId` was the natural choice. Otherwise we're kind of emulating a restful access pattern with query strings. I can see other people have shown interest in filters on a queue, I thought about doing that but it has the potential for way more complexity depending on your overall vision for the product. I thought a "view one job" page was a safe bet for getting a single job, along with the action buttons desired for administration / operations tasks to do with that job only. Ultimately it's your project (which we appreciate getting to use) so let me know if you feel strongly about this.

felixmosh:
Yeah, so I guess we can proceed with this... I'll need to review this PR more throughly ...

packages/api/src/handlers/job.ts
```diff
@@ -0,0 +1,24 @@
...
+import { queueProvider } from '../providers/queue';
+import { jobProvider } from '../providers/job';
+
+async function getJobState(
+  _req: BullBoardRequest,
+  job: QueueJob
+): Promise<ControllerHandlerReturnType> {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  const state = await job?.getState();
```
jamesgweber, line 11: I'm not 100% sure what to do here. Exposing `QueueJob#getState(): Promise<Status>;` in `typings/app.ts` causes a big conflict between `BullAdapter` and `BaseAdapter`.

packages/api/typings/app.ts
```diff
@@ -33,6 +33,8 @@ export interface QueueJob {
   retry(state?: JobRetryStatus): Promise<void>;
 
   toJSON(): QueueJobJson;
+
+  getState(): Promise<Status | 'stuck' | 'waiting-children' | 'unknown'>;
```
jamesgweber, line 37: Apparently this is the full intersection of states 🤷‍♂️  - idk if this is going to cause problems, any advice here would be greatly appreciated.

felixmosh: Why do you need to specify all of the statuses (which the board not supports yet, as far as i know, bull doesn't support waiting-children as well)

This interface must be a unify version between bullmq & bull

felixmosh (changes requested):

packages/api/typings/app.ts
jamesgweber: I'm getting a build error unless I include them all.
https://gist.github.com/jamesgweber/cbf1e8bbf29f169964d37b6adb8c405a

felixmosh:
Hi @jamesgweber I didn't forget about this PR...
I have a new born which overloads the entire system :)

Will review it asap

felixmosh (changes requested):

felixmosh:
Hi @jamesgweber I've reviewed the code, it looks OK.
There are 1 issue that arises since Job page doesn't need the entire polling of queues.

After requested changes, I will approve this PR, and will make several refactors to solve this issue (basically break down the useStore hook into small parts)

packages/ui/src/pages/JobPage/JobPage.tsx
```diff
@@ -0,0 +1,74 @@
...
+}) => {
+  const { search } = useLocation();
+  const history = useHistory();
+  const { name, jobId } = useParams<any>();
+  const [job, setJob] = useState<AppJob>();
+  const [status, setStatus] = useState<Status>(selectedStatus[queue?.name || '']);
+
+  useInterval(() => {
+    fetchJob();
+  }, 5000);
```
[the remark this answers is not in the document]
jamesgweber: > should use users config of interval time
idk what this means - if you can point me to an example I'm happy to make the change

>Do we pause the interval of the QueuePage?
no - I just let the polling continue because it keeps everything in sync and it's simple. traffic is not a real concern for an app like this imo. You probably wouldn't have 1000s of concurrent users.

packages/ui/src/pages/QueuePage/QueuePage.tsx
```diff
@@ -47,6 +50,7 @@ export const QueuePage = ({
         <JobCard
           key={job.id}
           job={job}
+          jobUrlPath={`/queue/${encodeURIComponent(queue.name)}/${job.id}${search}`}
```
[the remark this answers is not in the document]
jamesgweber: can do - this again conflicts with some old types deep in the app
`QueueJobJson["id"]` can be `null` for some reason. I can't think of how or why this would happen. -- I'll make the change but it's ugly -> 
```
encodeURIComponent(job.id ?? '')
```

felixmosh:
Thank you for the PR
I need to refactor the code to meet with some issues... then I'll publish a new version

PR #613, opened 2023-07-26
improvement: Ensure jobs per page has a max of 50

tayclark:
The `InputField` `max` doesn't prevent someone from inputting a higher value, it only limits how high the component will let you _click_ the up arrow. Since a large page size w/ continuous polling on an open window could cause issues the jobs per page configuration will now be capped at 50 if an input > 50 is provided


felixmosh:
Hi, thank you for this PR.

I think that instead of adding an additional config, it is better to solve the max issue, it can be done by limiting num of chars or even with a verification of the max number, WDYT?

tayclark:
Hi @felixmosh. Yea, that works. I've updated the PR

felixmosh:
Thanx for the change ,🙏🏼

felixmosh:
Released in `v5.7.1`

PR #614, opened 2023-08-02
Added search functionality

CJSarao:


felixmosh:
Hi @CJSarao thank you for this PR.

The filtering that search that you've implemented is only filters the current page list, which means that it partial.
Feature such search must filter on the server to show correct results.

WDYT, can you try to implement it?

CJSarao:
> Hi @CJSarao thank you for this PR.
> 
> The filtering that search that you've implemented is only filters the current page list, which means that it partial. Feature such search must filter on the server to show correct results.
> 
> WDYT, can you try to implement it?

Hi Felix, thanks for your time in looking at this. I actually didn't mean to open this PR for the original repo, only for my own fork for now (first time doing this, my bad). I meant to play around and test it more as well. I'll definitely give a shot at filtering for all jobs, that's what I'm going for. 

When that happens, should I reopen a PR for felixmosh:master? I've noticed some issues with people asking for this, just decided to give it a shot myself.

PR #616, opened 2023-08-08
Search functionality proposal

CJSarao:
Hi @felixmosh, this solution overloads the getAppQueues method a bit, let me know what you think and if you had something else in mind.. 

- The search functionality keeps the number of jobs per page and scales the number of pages accordingly. 
- Search is done with onBlur on the textfield, so clicking away or pressing enter triggers the search. 
- Search is looking at job ids and data properties, could add more values if needed
- Retry all and clean all actions are still shown even if the jobs are filtered, but still affect all of the jobs (not only the filtered ones)
->Is this ideal behavior? 
->Should the bulk actions be hidden if we're filtering or should they affect only the filtered jobs? 

Also, since the newest release, I have typing errors when I try to test locally, I see that they were in the github action test as well on the last PR...


CJSarao:
Hey, waiting for feedback on the proposed solution before I pull the most recent changes and address conflicts

packages/api/src/handlers/queues.ts
```diff
@@ -71,6 +71,7 @@ async function getAppQueues(
   return Promise.all(
     pairs.map(async ([queueName, queue]) => {
       const isActiveQueue = decodeURIComponent(query.activeQueue) === queueName;
+      const isFiltered:Boolean = query.searchPrompt;
```
felixmosh, line 74: if it should be boolean, it should be something like:
```js
const isFiltered= Boolean(query.searchPrompt);
// or
const isFiltered= !!query.searchPrompt;
```

packages/api/src/handlers/queues.ts
```diff
@@ -84,15 +85,31 @@ async function getAppQueues(
...
+      const formattedJobs = jobs.filter(Boolean).map((job) => formatJob(job, queue));
 
       const description = queue.getDescription() || undefined;
 
+      let filteredFormattedJobs:AppJob[] = [];
+      let filteredPagination:Pagination = pagination;
+      if (isFiltered) {
+        const filter = query.searchPrompt.toLocaleLowerCase();
+        const allJobs = isActiveQueue
+          ? await queue.getJobs(status)
```
felixmosh, line 97: There is any other way to search the data in Redis or something like this?
I think that fetching all jobs in order to filter them out is not efficient.

felixmosh (changes requested):

felixmosh:
Hi @CJSarao, thanx for this PR, as search feature is really required by others, let's see the best options to implement it properly.

CJSarao:
Hey @felixmosh, thanks for your review! I agree with most of your suggestions. 

As to why I needed to persist the value, it felt the best UX wise for my needs in my use case of the library. I see no problem with not storing it. 

> There is any other way to search the data in Redis or something like this?
I think that fetching all jobs in order to filter them out is not efficient.

Regarding that, I'm not aware of bull/bullmq exposing this capability, did you mean for this library to interact with redis directly? My understanding is that we're already polling the full list pretty regularly, which is why I wasn't too concerned with efficiency.. I'd like to understand what you have in mind, not always easy with async messages haha

felixmosh:
We are mot pulling all the list... It gets pagination on the redis sode.

I've opened a feature request in bullmq, hope that they will help.

CJSarao:
> We are mot pulling all the list... It gets pagination on the redis sode.
> 
> I've opened a feature request in bullmq, hope that they will help.

That's right, good point. Keep me posted I'd be glad to work on this some more and make it better!

felixmosh:
https://github.com/taskforcesh/bullmq/issues/2135

Issue #618, opened 2023-08-14
Feature: Metrics support


alko89:
Bull and BullMQ support metrics, which can be enabled with metrics option:

```typescript
// Bull
new Queue3(
  queueName, 
  {
    metrics: {
      maxDataPoints: MetricsTime.ONE_WEEK 
    }
  }
);

// BullMQ (in worker)
new Worker(
  queueName,
  async (job) => { ... },
  {
    metrics: {
      maxDataPoints: MetricsTime.ONE_WEEK,
    },
  }
);
```

This enables additional data in Redis, which can be fetched by using `getMetrics` method and provides information about job completion/failure over time with some additional metadata:

```typescript
export interface Metrics {
    meta: {
        count: number;
        prevTS: number;
        prevCount: number;
    };
    data: number[];
    count: number;
}
```

This would allow displaying a graph of completed and failed jobs over time, it should also provide information about job response time, process time and throughput.

![image](https://github.com/felixmosh/bull-board/assets/2822030/ae74ca1c-005d-41ad-a23a-405e9cfc55a6)

Relates to #258

PR #619, opened 2023-08-14
feat: add metrics api endpoint

alko89:
Added api endpoint for metrics. Started working on adding UI support for completed/failed graph, but I suck at React so it might take a while :sweat_smile: 

Updates #618

### DIFF
--- a/example.ts
+++ b/example.ts
@@ -1,6 +1,6 @@
 import * as Bull from 'bull';
 import Queue3 from 'bull';
-import { Queue as QueueMQ, Worker } from 'bullmq';
+import { MetricsTime, Queue as QueueMQ, Worker } from 'bullmq';
 import express from 'express';
 import { BullMQAdapter } from '@bull-board/api/src/queueAdapters/bullMQ';
 import { BullAdapter } from '@bull-board/api/src/queueAdapters/bull';
@@ -15,7 +15,7 @@ const redisOptions = {
 
 const sleep = (t: number) => new Promise((resolve) => setTimeout(resolve, t * 1000));
 
-const createQueue3 = (name: string) => new Queue3(name, { redis: redisOptions });
+const createQueue3 = (name: string) => new Queue3(name, { redis: redisOptions, metrics: { maxDataPoints: MetricsTime.ONE_WEEK } });
 const createQueueMQ = (name: string) => new QueueMQ(name, { connection: redisOptions });
 
 function setupBullProcessor(bullQueue: Bull.Queue) {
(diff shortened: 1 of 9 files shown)


packages/api/src/queueAdapters/base.ts
```diff
@@ -56,6 +58,8 @@ export abstract class BaseAdapter {
 
   public abstract getJobLogs(id: string): Promise<string[]>;
 
+  public abstract getMetrics(type: JobRetryStatus, start?: number, end?: number): Promise<QueueMetrics | undefined | null>;
```
felixmosh, line 61: Why did you used the `JobRetryStatus` instead of `JobStatus`

example.ts
```diff
@@ -15,7 +15,7 @@ const redisOptions = {
 
 const sleep = (t: number) => new Promise((resolve) => setTimeout(resolve, t * 1000));
 
-const createQueue3 = (name: string) => new Queue3(name, { redis: redisOptions });
+const createQueue3 = (name: string) => new Queue3(name, { redis: redisOptions, metrics: { maxDataPoints: MetricsTime.ONE_WEEK } });
```
felixmosh, line 18: You are using here bullmq variable with bull lib, is that OK?

felixmosh (changes requested):

felixmosh:
Hi @alko89 thank you for this PR, it definitely something that I would like to add to `Bull-Board`. 
Since this data is "on demand" it should not be presented all the time in front of the user, therefore, I think that it can be inside the queue actions drop down, which will open a modal with the graphs, WDYT?

packages/api/src/queueAdapters/base.ts
alko89: From what I gathered, only valid status for metrics is `completed` and `failed`. I can also create a new type for this

alko89:
@felixmosh personally I would love to have metrics visible as soon as I click on a desired queue. Maybe add it as the first tab in queue page if they are available, else hide the tab?

image

I'm also fine with other options that you fine more appropriate :)

packages/api/src/queueAdapters/base.ts
felixmosh: So it is better to create a new type for it

alko89:
Think the API part is done. But I haven't got the time to work on the FE part yet unfortunately.

nocodehummel:
2Cs, when adding Metrics as additional Tab it would be good to reduce the spaces for Tabs. On my screen the tabs-menu requires often horizontal scrolling while other content is visible.

kyleolsondesign:
Any chance we could ship the API endpoint and work on the FE as a separate branch?

acgfbr:
lets merge this lmfao, 2 years wtf

felixmosh:
This API alone won't help u since you will need to implement the Front-end as well...

burnmaniac:
I would like to implement front-end part of this feature, but I would love if someone could give me some guidance on the UI, preferably as wireframe(s). I also think I'd do this insanely quickly once I know what is the requirement for the UI.

felixmosh:
This PR should me rebased.

felixmosh:
Closing this in favor of existing materics

PR #624, opened 2023-09-13
Enhanced Time Formatting for Short-duration Jobs

fer112233:
**Current Behavior:**
For jobs that complete in under 5 seconds, the system currently displays the duration as "less than 5 seconds". This approach obscures the details for tasks that finish in mere milliseconds.

**Proposed Change:**
To provide more granular information about job execution times, this pull request introduces an enhanced time formatting mechanism:

Jobs that take less than 1 second will display their duration in milliseconds (e.g., "320 miliseconds").
Jobs that take between 1 to 5 seconds will display their duration with a precision of two decimal places (e.g., "3.25 seconds").

**Rationale:**
This change ensures that users have a clear understanding of the job durations, especially for those that complete rapidly. It will be particularly valuable for performance monitoring and optimization efforts, where millisecond-level precision can be crucial.


felixmosh:
Thanx for this PR 🙏🏼
Released in `v5.8.2`

PR #625, opened 2023-09-13
[express] Make ExpressAdapter fields protected instead of private

### LINKED ISSUE #626: [express] Make ExpressAdapter fields protected instead of private
Hi, it make it difficult to extends the `ExpressAdapter` provided because of private fields. Why not use protected fields instead?

RedbeanGit:
Private fields make it difficult to extends ExpressAdapter. Using protected instead make ExpressAdapter open to extension.

Close #626


felixmosh:
thank you for this PR

released in v5.8.3

PR #630, opened 2023-09-28
Add headers UIConfig option

djmadeira:
Adds the option to set request headers through UI config. This is useful if you have CSRF protection enabled on your app, and need to tell bull to pass a certain header with every request.


felixmosh:
This allows you to set a static header, which means it is not a good solution as a csrf protection... Am I correct?

djmadeira:
@felixmosh oh, I see the problem now. UIConfig is set on initialization of the express router, not per request. I will open a new PR with a different approach

felixmosh:
You can add a middlewere, that adds random cookie, which the client should attach as the header (this part should be added)

PR #639, opened 2023-10-18
fix(nestjs): edit bullboard path have global prefix when using fastify

lytaitruong:
Hi @felixmosh, first of all thank you and everyone build this repo. It nice

- I using NestJS + Fastify, and now try implement the BullMQ, so I add your library, it nice!
- But when I test it not recognize if I have a global prefix in Fastify
Issue

- I've read all the issues of this repo but not find the best way to fix this
- So I try on your code and figure out the issue in here
Bug

The result after change my code is working correctly my expectation
Fixway

I create this PR maybe it make it right. If you think not more problem you can merge it

Thank and have a good day

### DIFF
--- a/packages/nestjs/src/bull-board.root-module.ts
+++ b/packages/nestjs/src/bull-board.root-module.ts
@@ -29,7 +29,7 @@ export class BullBoardRootModule implements NestModule {
     if (isFastifyAdapter(this.adapter)) {
       this.adapterHost.httpAdapter
         .getInstance()
-        .register(this.adapter.registerPlugin(), {prefix: this.options.route});
+        .register(this.adapter.registerPlugin(), {prefix: `${ globalPrefix }${ this.options.route }`});
 
       return consumer
         .apply(this.options.middleware)


felixmosh:
Hi @lytaitruong thank you for this PR.

Unfortunately I'm not familiar with Nest.js, so I'm not sure if this solves the problem, maybe @DennisSnijder can review this PR?

DennisSnijder:
Hi @lytaitruong thank you for this!

I've been debugging / trying to get the prefix working with Fastify for quite some time now, previously it was due to a prefixing bug in NesJS with the Fastify adapter I wasn't able to get it to work.

I'll try out the PR in the next few days!

Thank you!

h4sohail:
Can we merge this in? I'm having a similar issue with using global prefixes.

felixmosh:
> Can we merge this in? I'm having a similar issue with using global prefixes.

Looks like it is an issue with Nest.js FastifyAdapter, threfore, I don't think it should be solved in this lib.

anteqkois:
@felixmosh  Any progress? It also don't work for me when I use global prefix

anteqkois:
@felixmosh I checed this PR and it fix issue for NestJS with Fastify, but we can not use it in production or we must fork it and change one line only for this.

felixmosh:
closed by #704

PR #648, opened 2023-11-20
upgrade-express-ex: upgrade versions + remove QueueScheduler

ivnnv:
I just discovered this project (thanks for sharing it with the community btw!) and wanted to test it before migrating my current bull-arena, this example needed upgrade as BullMQ's QueueScheduler is not needed nowadays plus the very own express adapter and express.


felixmosh:

PR #652, opened 2023-11-23
Update QueueAdapterOptions typing

AlexisAnzieu:
Thanks for your awesome work!

Here is a typing PR to add the ability to modify one attribute only and not the whole option object.

image

### DIFF
--- a/packages/api/typings/app.ts
+++ b/packages/api/typings/app.ts
@@ -27,10 +27,10 @@ export type JobStatus<Lib extends Library = 'bullmq'> = Lib extends 'bullmq'
 export type JobCounts = Record<Status, number>;
 
 export interface QueueAdapterOptions {
-  readOnlyMode?: boolean;
-  allowRetries?: boolean;
-  prefix?: string;
-  description?: string;
+  readOnlyMode: boolean;
+  allowRetries: boolean;
+  prefix: string;
+  description: string;
 }
 
 export type BullBoardQueues = Map<string, BaseAdapter>;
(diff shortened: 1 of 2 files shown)


felixmosh:
Hi, @AlexisAnzieu thank you for wanting to contribute, but all QueueAdapters are defined with Partial<> as needed.
https://github.com/felixmosh/bull-board/blob/37a618f8cc9c58b0591f288a298293650ec2b133/packages/api/src/queueAdapters/bullMQ.ts#L13

What is the TS error that you get?

gboudreau:
Hi @felixmosh. (I work with Alexis.)

The error is:
```
error TS2739: Type '{ readOnlyMode: true; }' is missing the following properties from type 'QueueAdapterOptions': allowRetries, prefix, description

55       options: {
         ~~~~~~~

  node_modules/@bull-board/nestjs/dist/bull-board.types.d.ts:18:5
    18     options?: QueueAdapterOptions;
           ~~~~~~~
    The expected type comes from property 'options' which is declared here on type 'BullBoardQueueOptions'
```

Our code that triggers this error:
```ts
@Module({
  // ...
  imports: [
    // ...
    BullBoardModule.forFeature({
      name: QueueName.OPENAI,
      adapter: BullAdapter,
      options: {
        readOnlyMode: true,
      },
    }),
  ],
})

felixmosh:
Hi, thanx for the clarification, can you change the PR, so that the usage of this type will use Partial<> (inside nest module)

felixmosh:
Released in `v5.9.2`
Thank you for your PR

AlexisAnzieu:
Thank you for your reactivity 🚀

PR #669, opened 2024-01-05
Add H3 adapter

### LINKED ISSUE #623: Usage with Nuxt 3
Has there been a method found to get this working with Nuxt 3? Most of what I've found to get an express instance working with Nuxt, was using Nuxt 2. 

I tried to setup an express app as a middleware in my nuxt server then passing the express app to the event context so I could try to render the UI ejs in my route file but that didn't work, and wasn't very pretty.

Nuxt is built off of [Nitro](https://nitro.unjs.io/), and [h3](https://github.com/unjs/h3). So I was thinking we may need an adapter for h3? 

Once the adapter is setup it would need to have a catch all handler for a queues route for the adapter to take over routing. Server routes don't support a [...] route yet, so it may have to exist under the server/api routes.

Maybe a Nuxt module is the right way to go? 

I'm happy to spend the time creating an adapter, or module, whatever is needed, but I thought I'd open an issue in-case anyone else has some input on how this should be tackled...

genu:
Resolves #623 

Usage:

```ts
import { createApp, createRouter } from 'h3';
import { createBullBoard } from '@bull-board/api';
import { H3Adapter } from '@bull-board/h3';

const serverAdapter = new H3Adapter();

createBullBoard({
  queues: [...],
  serverAdapter,
});

export const app = createApp();

// your app router
const router = createRouter();
router.use(...)

app.use(router);

// Add the bullboard handlers
app.use(serverAdapter.registerHandlers());



```


packages/h3/src/H3Adapter.ts
```diff
@@ -0,0 +1,130 @@
...
+  public setBasePath(path: string): H3Adapter {
+    this.basePath = path;
+    return this;
+  }
+
+  private getStaticFile(path: string, filename: string) {
+    return fs.readFileSync(`${this.viewPath}${path}/${filename}`, "utf-8");
+  }
+
+  private getContentType(filename: string) {
```
genu, line 29: We could move this out into a `/helpers/getContentType` as other adapters do something similar.

packages/h3/src/H3Adapter.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  public setStaticPath(staticsRoute: string, _staticsPath: string): H3Adapter {
+    this.uiHandler
+      .get(
+        this.basePath,
+        eventHandler(async () => {
+          return ejs.renderFile(this.viewPath + "/index.ejs", {
+            basePath: `${this.basePath}/`,
+            title: this.uiConfig.boardTitle || "BullMQ",
```
AidanHibbard, line 63: I think `??` is preferred over `||`.

packages/h3/src/H3Adapter.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  public setApiRoutes(routes: AppControllerRoute[]): H3Adapter {
+    routes.forEach(({ route, handler, method }) => {
+      this.uiHandler.use(
+        `${this.basePath}${route}`,
+        eventHandler(async (event) => {
+          return handler({
+            queues: this.bullBoardQueues!,
+            params: getRouterParams(event) as any,
+            query: getQuery(event) as any,
```
AidanHibbard, line 102: Do you get an error unless this is `as any`?

```ts
getQuery(): Record<string, any>;

// should satisfy 

BullBoardRequest.query: Record<string, any>;
```

Same with params.

packages/h3/src/H3Adapter.ts
felixmosh: Why do you need this title field? 
I doesn't exist in other adapters (uiConfig is missing here)

genu:
@felixmosh I would consider this complete, let me know if there are any other changes needed.

packages/h3/src/H3Adapter.ts
```diff
@@ -0,0 +1,170 @@
...
+    const getStaticPath = (relativePath: string) => {
+      if (!this.statics) return '';
+
+      return `${this.statics.path}${relativePath.replace(
+        `${this.basePath}${this.statics.route}`,
+        ''
+      )}`;
+    };
+
+    const { method, route } = this.entryRoute as AppViewRoute;
```
AidanHibbard, line 102: Is this `as AppViewRoute` needed? `H3Adapter.entryRoute` is already a type of AppViewRoute.

AidanHibbard:
@genu Are you able to spin-up an example repository with this? I'm trying to test this by copying the libs over to my repo here -> https://github.com/AidanHibbard/bull-board-h3-test/

But I'm running into errors when I try to set up the route following your example...

There's a good chance I'm missing something.

genu:
> @genu Are you able to spin-up an example repository with this? I'm trying to test this by copying the libs over to my repo here -> https://github.com/AidanHibbard/bull-board-h3-test/
> 
> But I'm running into errors when I try to set up the route following your example...
> 
> There's a good chance I'm missing something.

The example that I provided in this PR is for `h3`. In your repository, you're trying to applying it in a Nuxt context.

Because Nuxt hijacks the routing, you can't really use the example I provided as is.

For nuxt, you have to

1. Define your route path as:
```
```

AidanHibbard:
@genu That worked swimmingly, thanks

genu:
Excellent! 😎

packages/h3/src/H3Adapter.ts
```diff
@@ -0,0 +1,169 @@
...
+      throw new Error(`Please call 'setApiRoutes' before using 'registerPlugin'`);
+    } else if (!this.bullBoardQueues) {
+      throw new Error(`Please call 'setQueues' before using 'registerPlugin'`);
+    } else if (!this.errorHandler) {
+      throw new Error(`Please call 'setErrorHandler' before using 'registerPlugin'`);
+    }
+
+    const getStaticPath = (relativePath: string) => {
+      if (!this.statics) return '';
+
```
felixmosh, line 95: this function currently allows to "get" any file in any directory
image

it should be much stricter, using `path.resolve` to remove all path traversal, then check if the result starts with static, and only then check if file exists.

examples/with-h3/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const serverAdapter = new H3Adapter();
+serverAdapter.setBasePath('/ui');
+
+export const app = createApp();
+
+const router = createRouter();
+
+app.use(router);
+app.use(serverAdapter.registerHandlers());
```
felixmosh, line 17: This throws an error, since you've called `registerHandlers` before you set any value.

I think that it is better to co-locate the code.

```ts
import { createApp, createRouter, eventHandler } from 'h3';
import { createBullBoard } from '@bull-board/api';
import { BullMQAdapter } from '@bull-board/api/bullMQAdapter';
import { H3Adapter } from '@bull-board/h3';
import { Queue as QueueMQ, RedisOptions, Worker } from 'bullmq';

const redisOptions: RedisOptions = {
  port: 6379,
  host: 'localhost',
  password: '',
};

const createQueueMQ = (name: string) => new QueueMQ(name, { connection: redisOptions });

function setupBullMQProcessor(queueName: string) {
  new Worker(
    queueName,
    async (job) => {
      for (let i = 0; i <= 100; i++) {
        await sleep(Math.random());
        await job.updateProgress(i);
        await job.log(`Processing job at interval ${i}`);

        if (Math.random() * 200 < 1) throw new Error(`Random error ${i}`);
      }

      return { jobId: `This is the return value of job (${job.id})` };
    },
    { connection: redisOptions }
  );
}

const exampleBullMq = createQueueMQ('BullMQ');
setupBullMQProcessor(exampleBullMq.name);

const serverAdapter = new H3Adapter();

createBullBoard({
  serverAdapter,
});

const router = createRouter();

router.use(
  '/add',
  eventHandler(async () => {
    await exampleBullMq.add('myJob', { foo: 'bar' });

    return true;
  })
);

export const app = createApp();
app.use(router);
app.use(serverAdapter.registerHandlers());
```

packages/h3/src/H3Adapter.ts
```diff
@@ -0,0 +1,169 @@
...
+              type: getContentType(id),
+            };
+          },
+        });
+      })
+    );
+
+    this.apiRoutes.forEach(({ route, handler, method }) => {
+      this.uiHandler.use(
+        `${this.basePath}${route}`,
```
felixmosh, line 142: `route` has an interface of `string | string[]`, you should make sure it is an array, and run for each route.

One way to do so is using `[].concat(route)` which will convert it to an array of strings, then you can run for each on of the values.

felixmosh (changes requested):

packages/h3/src/H3Adapter.ts
```diff
@@ -0,0 +1,204 @@
...
+
+  private registerRoute(
+    routeOrRoutes: string | string[],
+    method: HTTPMethod,
+    handler: AppControllerRoute['handler']
+  ) {
+    const { bullBoardQueues } = this;
+
+    if (!bullBoardQueues) {
+      throw new Error(`Please call 'setQueues' before using 'registerPlugin'`);
```
felixmosh, line 172: instead of `registerPlugin` you called it `registerHandlers`, change the text.

packages/h3/src/H3Adapter.ts
```diff
@@ -0,0 +1,204 @@
...
+    this.errorHandler = handler;
+
+    return this;
+  }
+
+  public setApiRoutes(routes: AppControllerRoute[]): H3Adapter {
+    routes.forEach(({ route, handler, method: methodOrMethods }) => {
+      const methods = Array.isArray(methodOrMethods) ? methodOrMethods : [methodOrMethods];
+
+      methods.forEach((method) => {
```
felixmosh, line 60: redundant loop, `router.use` of H3 supports an array of methods, so you can pass the `methods` to `this.registerRoute`.

felixmosh (changes requested):

packages/h3/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "build": "tsc",
+    "clean": "rm -rf dist"
+  },
+  "dependencies": {
+    "@bull-board/api": "5.11.0",
+    "@bull-board/ui": "5.11.0",
+    "ejs": "^3.1.9"
+  },
+  "peerDependencies": {
+    "h3": "^1.10.0"
```
felixmosh, line 37: why `peerDeps`? this adapter is using `h3`, so it should be in `deps`. (otherwise the build is failing)

felixmosh:
Thank you for you contribution 🙏🏽

felixmosh:
Released in `v5.13.0` 🎊

genu:
@AidanHibbard If you are using nuxt, I've created a module that abstracts a lot of what this adapter is doing.

Have a look at: https://github.com/genu/nuxt-concierge

AidanHibbard:
@genu This looks great! Looks pretty similar to the module @productdevbook [wrote for Pergel](https://oku-ui.com/pergel/nuxt/bullmq/installation).

Issue #264, opened 2021-05-01
Feature: set default tab in BullAdapter settings


zxpectre:
Greate project already! Simple suggestion for non technical users UX:

If only one could set the default open tab in BullAdapter settings, for example to "logs" instead of data, a non experienced user could find the UI usefull and understandable enough. We devs love json in data and error tab but humans don't haha. Just simple stupid suggestion I could use with my team. 

const lastOne="also maybe set the array of tabs to show, to hide others.";

felixmosh:
Thank you for the suggestion, I'm not sure if this feature has a lot of demand, let's see if so, I will implement it.

If it is crucial for you, you can make a PR :]

sroettering:
I would also appreciate this feature.
We have jobs that can have a huge payload of around 100k uuids. The pages sometimes load terribly slow because every job starts on the data tab.

felixmosh:
@sroettering you can filter out the jobs payload by providing a formatter...

```js
const queueAdapter = new BullMQAdapter(queueMQ);

queueAdapter.setFormatter('data', (data) => 'FILTER OUT LARGE OBJECTS');
queueAdapter.setFormatter('returnValue', (returnValue) => 'FILTER OUT LARGE OBJECTS');
```

sroettering:
Oh wow didn't know that. I will give it a try. Is it also possible to maybe collapse the json by default? Sorry for being a little bit off topic here

felixmosh:
It is possible, can you open a new feature request please?

Crhis35:
Is there a way to achieve this in NestJS?
My current approach:

`class CustomBullMQAdapter extends BullMQAdapter {
  constructor(queue: Queue, options?: Partial<QueueAdapterOptions>) {
    super(queue, options);
    this.setFormatter('data', (data) => 'masked');
  }
}

@Module({
  imports: [
    BullModule.registerQueue({ name: 'test' }),
    BullBoardModule.forFeature({
      name: 'checks',
      adapter: CustomBullMQAdapter,
    }),`

amoniaka-knabino:
Hi! I noticed that the default job tab part of this issue was partially addressed by #676, which added a user-level “Default job tab” setting stored in the browser settings.

Would you still be interested in a server-side configuration for this, so app maintainers can set the initial default for all users, e.g. via `uiConfig.defaultJobTab` or a queue adapter option?

Also, is the second part of the original request still desired: allowing maintainers to configure which job detail tabs are shown/hidden?

[...]

PR #676, opened 2024-01-21
Add default job tab setting

yerlantemir:
Hi @felixmosh! First of all, thank you for this amazing tool! We use it everyday and really enjoy it. This is my first PR on this project, would love to get any feedback from you and hope it will be useful for other users as well 🙏

In our project, we were required to implement this feature https://github.com/felixmosh/bull-board/issues/264. So basically, we want to have "logs" tab to be opened by default. 

To make the solution more flexible I decided to create setting **default-job-tab** which is used for selecting initial job tab.

### Screens:
IMAGE 2024-01-21 10:42:21

IMAGE 2024-01-21 10:42:24

CleanShot 2024-01-21 at 10 43 38@2x

### Todos:
* Add pt-BR locales
### Description: 
This commit adds a new setting for the default job tab in the SettingsModal component. The selected default job tab is stored in the useSettingsStore and can be changed by the user.

### DIFF
--- a/packages/ui/src/hooks/useDetailsTabs.tsx
+++ b/packages/ui/src/hooks/useDetailsTabs.tsx
@@ -1,32 +1,48 @@
 import { useEffect, useState } from 'react';
 import { Status } from '@bull-board/api/typings/app';
 import { STATUSES } from '@bull-board/api/src/constants/statuses';
+import { useSettingsStore } from './useSettings';
 
-const regularItems = ['Data', 'Options', 'Logs'] as const;
+export const availableJobTabs = ['Data', 'Options', 'Logs', 'Error'] as const;
 
-export type TabsType = typeof regularItems[number] | 'Error';
+export type TabsType = (typeof availableJobTabs)[number];
+
+const FALLBACK_TAB: TabsType = 'Data';
 
 export function useDetailsTabs(currentStatus: Status, isJobFailed: boolean) {
   const [tabs, updateTabs] = useState<TabsType[]>([]);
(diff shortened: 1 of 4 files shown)


felixmosh:
Hi, thank you for this PR!
One thing, what happens if I select the `error` tab, but it doesn't exists?

packages/ui/src/hooks/useDetailsTabs.tsx
```diff
@@ -1,32 +1,44 @@
...
+
+const FALLBACK_TAB: TabsType = 'Data';
 
 export function useDetailsTabs(currentStatus: Status, isJobFailed: boolean) {
   const [tabs, updateTabs] = useState<TabsType[]>([]);
-  const [selectedTabIdx, setSelectedTabIdx] = useState(0);
-  const selectedTab = tabs[selectedTabIdx];
+  const { defaultJobTab } = useSettingsStore();
+
+  const [selectedTab, setSelectedTab] = useState<TabsType>(defaultJobTab);
```
felixmosh, line 16: defaultJobTab can be `Error` but the job is not failed, it should ignore the settings default if there is no available tab.

```js
const [selectedTab, setSelectedTab] = useState<TabsType>(tabs.find(tab => tab === defaultJobTab) || tabs[0]);
```

yerlantemir:
> Hi, thank you for this PR! One thing, what happens if I select the `error` tab, but it doesn't exists?

Hi, thanks for your quick reply!

I pushed changes suggested by you. Will be waiting your review.

felixmosh:
Thank you for you contribution 🙏🏽

PR #687, opened 2024-02-16
compactMode: add collapsible functionality + reog job card layout

ivnnv:
https://github.com/felixmosh/bull-board/assets/23552631/d208013c-5294-4063-a314-388304e9ec45


packages/ui/src/components/JobCard/Details/DetailsContent/DetailsContent.tsx
```diff
@@ -49,9 +50,7 @@ export const DetailsContent = ({ selectedTab, job, actions }: DetailsContentProp
           {t('JOB.SHOW_ERRORS_BTN')} <ArrowDownIcon />
         </Button>
       ) : (
-        <Highlight language="stacktrace" key="stacktrace">
-          {stacktrace.join('\n')}
-        </Highlight>
+        <Highlight language="stacktrace" key="stacktrace" code={stacktrace.join('\n')} />
```
felixmosh, line 53: Why did you changed the prop name from `children` to `code`?
`JSON` is not code :]

anyway, this is not related to the compact mode, so let's separate it to a diffrent PR.

packages/ui/package.json
```diff
@@ -42,8 +42,10 @@
     "@babel/runtime": "^7.17.9",
     "@pmmmwh/react-refresh-webpack-plugin": "^0.5.5",
     "@radix-ui/react-alert-dialog": "^1.0.4",
+    "@radix-ui/react-collapsible": "^1.0.3",
     "@radix-ui/react-dialog": "^1.0.4",
     "@radix-ui/react-dropdown-menu": "^2.0.5",
+    "@radix-ui/react-icons": "^1.3.0",
```
felixmosh, line 48: This lib is not used

packages/ui/src/components/JobCard/JobCard.tsx
```diff
@@ -35,49 +39,77 @@ export const JobCard = ({
...
         </div>
-      </div>
+
+        <Collapsible.Content>
+          <div className={`jobDetails ${s.details}`}>
+            <div className={`sideInfo ${s.sideInfo}`}>
+              <Timeline job={job} status={status} />
+            </div>
+
+            <div className={`jobContentWrapper ${s.contentWrapper}`}>
```
felixmosh, line 71: what is this static `jobContentWrapper` class?

packages/ui/src/components/JobCard/JobCard.tsx
```diff
@@ -35,49 +39,77 @@ export const JobCard = ({
   jobUrl,
 }: JobCardProps) => {
   const { t } = useTranslation();
+  const JobTitle = <h4 title={`#${job.id}`}>#{job.id}</h4>
+
+  // TODO: Get global config to set the initial state
+  // TODO2: Override to true when its a dedicated job page
+  const [isExpanded, setIsExpanded] = React.useState(true);
```
felixmosh, line 46: I think that this mode should be an "additional" display mode,
The current state or the collapsable one... it should not be merged to the current one.
Therefore, it is better to extract common components & use a switch for the display mode.

felixmosh (changes requested):

packages/ui/package.json
ivnnv: yeah, realized you have custom icons for arrows so I exchanged, its being deleted.

Btw, why did you chose not to use these icon pack? too slims?

packages/ui/src/components/JobCard/Details/DetailsContent/DetailsContent.tsx
ivnnv: children is very different from a pure string and this is not intented to have inside something like JSX.Element which is what childrens are.

Modifying this file actually has to do with this PR because the ref doesnt work when the card is collapsed and it doesnt show anything, also the current implementation has some sort of delay, now the tabs change the content straight away as seen in the video.

We could change `code` for `text` or something else, you name it! :)

And yeah Im happy creating an atomic PR for this only component before everything else.

packages/ui/src/components/JobCard/JobCard.tsx
ivnnv: I think this can be achieved just modifying this file (and you wont need to maintain twice) and it will only have little impact for users that dont activate the compact mode (basically just having a new arrow button top-right).

I undestand your position and the fear of having such a big diff showing right now in the PR but this really just wrapping everything on a outer container because the card needs to have a header with the id and the button, but the card body is basically the same than right now.

Now:

```
------------------------
| #ID         | Name   |
| TimeLine    | Data   |      
|             |        |
------------------------
```

Then:

```
------------------------
| #ID              Btn |
------------------------
| TimeLine    | Name   |
|             | Data   |      
------------------------
```

But if you are keen to have a different `JobCardCollapsible`, ill extract it.

packages/ui/package.json
ivnnv: removed 👍🏽

packages/ui/src/components/JobCard/JobCard.tsx
ivnnv: removed classes 👍🏽

ivnnv:
hey @felixmosh did you have the chance to look at this updated and comments? I really want to contribute to the main repo (have many other improvements in mind) but otherwise Id need to just fork.

felixmosh:
Hi, sorry I didn't managed to check it out yet, will do soon.
(I didn't forget about it)

packages/ui/src/components/Highlight/Highlight.tsx
```diff
@@ -1,58 +1,46 @@
...
-  }
-
-  private async highlightCode() {
-    const node = this.codeRef.current?.querySelector('code');
-    if (node) {
-      node.innerHTML = await asyncHighlight(this.props.children as string, this.props.language);
-    }
-  }
-}
+export const Highlight: React.FC<HighlightProps> = ({ language, text }) => {
```
felixmosh, line 13: Pls move this refactor to a separate PR 🙏🏽

felixmosh (changes requested):
Rest of the PR looks Ok to me.
Thank you.

(FYI, After this PR will be merged I will change a bit styles)

packages/ui/src/components/JobCard/JobCard.tsx
felixmosh: This exraction of the ID look ok 👍

packages/ui/src/components/Highlight/Highlight.tsx
ivnnv: sure! lets process this first 👍🏽

felixmosh (approved):

felixmosh:
Thank you for this awsome PR.
I will publish it after I refactor some styles.

felixmosh:
closes #687

felixmosh:
I probably needed to check the PR locally...
there is button inside a button...
image


And the collapse expand doesn't work for me...
I want to fix other bug, therefore, I will revert this PR... and will work on it later

PR #688, opened 2024-02-16
feat: show message under run time if delay has changed

emko:
looking a the issue i see you guys don't want to support getting the real run at time from the redis sorted sets score value instead  you want bullmq to add job.runAt field, can we at least have this ? 

added `*Delay changed; the new run time is currently unknown` under run at if the initial delay does not equal the current delay as Bullmq  does not return a timestamp for when the job will actually run, also i changed it to use the initial delay using the current delay will give you wrong run at time anyways.

image

### DIFF
--- a/packages/ui/src/components/JobCard/Timeline/Timeline.tsx
+++ b/packages/ui/src/components/JobCard/Timeline/Timeline.tsx
@@ -67,7 +67,8 @@ export const Timeline = function Timeline({ job, status }: { job: AppJob; status
         {!!job.delay && job.delay > 0 && status === STATUSES.delayed && (
           <li>
             <small>{t('JOB.WILL_RUN_AT')}</small>
-            <time>{formatDate((job.timestamp || 0) + job.delay, i18n.language)}</time>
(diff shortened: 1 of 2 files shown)


felixmosh:
Hi @emko thank you for the PR.
Can you elaborate on the issue this PR tries to solve? (example is prefered :] )

emko:
> Hi @emko thank you for the PR. Can you elaborate on the issue this PR tries to solve? (example is prefered :] )

https://github.com/taskforcesh/bullmq/issues/1728

Problem is if you ever change the delay on your job  with changeDelay BullMq does not store the new timestamp on the Job, so there is no way on the UI to know when this job will run. This PR just indicates that the jobs delay has been changed, we have many users who complain that jobs are not running when they said they will and this would tell them why it won't

we just have to wait for bullMQ to add something like this 
https://github.com/taskforcesh/bullmq/pull/1732

but for now i think there should be a warning if the delay has been changed

felixmosh:
Thank you for the explanation, I'll review that asap

felixmosh:
Thank you, released in `v5.15.0`

PR #694, opened 2024-02-26
serverInfo: show server info at MenuFooter

ivnnv:
This PR does 2 things (I can split this in two if you want) as I started thinking I could place the connection information next to the `Title`, so I refactored to avoid prop-drilling and each component to use exactly the info they need via hooks.

Then realized that would be better placed at the side just above the version so its not getting that much attention, this is mainly to be 100% sure of where your app is connected, as many times you are switching between local/qa/prod so need to be extra careful sometimes on what you are doing.

Screenshot 2024-02-26 at 00 43 41


packages/ui/src/App.tsx
```diff
@@ -7,7 +7,6 @@ import { HeaderActions } from './components/HeaderActions/HeaderActions';
 import { Loader } from './components/Loader/Loader';
 import { Menu } from './components/Menu/Menu';
 import { Title } from './components/Title/Title';
-import { useActiveQueue } from './hooks/useActiveQueue';
```
ivnnv, line 10: We dont really need this here, each component that needs can fetch with its own hook

packages/ui/src/components/Menu/Menu.tsx
```diff
@@ -1,15 +1,17 @@
...
+import { useQueues } from './../../hooks/useQueues';
 import { links } from '../../utils/links';
 import { SearchIcon } from '../Icons/Search';
+import { MenuFooter } from './MenuFooter';
 import s from './Menu.module.css';
 
-export const Menu = ({ queues }: { queues: AppQueue[] | null }) => {
+export const Menu = () => {
   const { t } = useTranslation();
+  const { queues } = useQueues();
```
ivnnv, line 13: also this way typescript infers the types and we dont need to import anything else

packages/ui/src/components/Title/Title.tsx
```diff
@@ -1,18 +1,21 @@
 import React from 'react';
 import s from './Title.module.css';
+import { useActiveQueue } from '../../hooks/useActiveQueue';
 
-interface TitleProps {
-  name?: string;
-  description?: string;
-}
+export const Title = () => {
+  const queue = useActiveQueue();
```
ivnnv, line 6: as hooks can be composed from other hooks, this hook doesnt need to have any arguments, it will resolve automatically which is the active queue

felixmosh:
I'm closing this since you can pass `uiConfig.boardTitle`, and even `uiConfig.boardLogo` to make this distiguish between boards.

```
createBullBoard({
    queues: [new BullMQAdapter(exampleBullMq), new BullAdapter(exampleBull)],
    serverAdapter,
    options: {
      uiConfig: { boardTitle: 'PROD server' },
    },
  });
```

image

felixmosh:
Regarding prop drilling, it looks OK, 
Can you prepare a clean, **focused** only on this task, PR?

ivnnv:
> focused
😆

sure mate :)

PR #697, opened 2024-02-27
compactMode: add collapsible functionality + reog job card layout

ivnnv:
Yeah this definitelly had a regression bug when I introduced getting the collapseJob settingsStore, this also corrects the double button, having the `Collapsible.Trigger` is not really needed if we control the state in our Button.


ivnnv:
hey @felixmosh, modified this and should be ready to merge now

felixmosh:
Cool, I'll review it asap

felixmosh:
I've reviewed the PR, it looks buggy...
Toggle on the settings should reflect on the screen, but it doesn't, refresh updates to the expected state.

"Collapse job" should define if the card can be collapseable or not (I prefere the full card view), this means that it should add / remove the arrow button.

packages/ui/src/components/JobCard/Timeline/Timeline.module.css
```diff
@@ -1,5 +1,5 @@
 .timeline {
-    padding: 1.5rem 1rem 1.5rem 0;
+    padding: 0.25rem 1rem 1rem 0;
```
felixmosh, line 2: why?

packages/ui/src/components/JobCard/JobCard.tsx
```diff
@@ -35,49 +40,82 @@ export const JobCard = ({
...
+            <div className={s.sideInfo}>
+              <Timeline job={job} status={status} />
+            </div>
+
+            <div className={s.contentWrapper}>
+              <div className={s.title}>
+                <h5>
+                  {t('JOB.NAME')}: {job.name}
+                  {job.attempts > 1 && (
+                    <span style={{ marginLeft: '0.5rem' }}>
```
felixmosh, line 82: move the inline style to css

packages/ui/src/components/JobCard/Timeline/Timeline.module.css
ivnnv: overall spacing is too big tbh

packages/ui/src/components/JobCard/JobCard.tsx
```diff
@@ -35,49 +40,82 @@ export const JobCard = ({
...
 }: JobCardProps) => {
   const { t } = useTranslation();
+  const { collapseJob } = useSettingsStore();
+
+  const [isOpen, setIsOpen] = React.useState<boolean>();
+
+  React.useEffect(() => {
+    if (isOpen === undefined)
+      setIsOpen(!jobUrl ? true : !collapseJob)
+  }), [];
```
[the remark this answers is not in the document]
ivnnv: it definitelly works, try it :)

this is "onComponentMount" so the initial state is undefined:

- If we are on a job page, the card is instantly expanded, and then theres no collapse button.
- If we are on the list page, the card takes the value defined in the app setting and from them controlled by only the state

ivnnv:
> I've reviewed the PR, it looks buggy... Toggle on the settings should reflect on the screen, but it doesn't, refresh updates to the expected state.
> 
This is not needed and it adds complexity to the state manager, a user will normally just define this for the whole session/forever, wont toggle this all the time.

> "Collapse job" should define if the card can be collapseable or not (I prefere the full card view), this means that it should add / remove the arrow button.
In this case a user would be in need of clicking expanding the card all the time to have a full list of jobs with only the ids, which is completely unusable

packages/ui/src/components/JobCard/JobCard.tsx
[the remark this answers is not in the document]
felixmosh: It doesn't, since the array of deps is outside of the useEffect :P

felixmosh:
@ivnnv think about it, I want the current view (without the clutter of collase expand at all... maybe only the change of the extraction of the id location to the head).

When I enable the `collapse job` I expect it to collapse all jobs (which is what happining if you refresh the page).

packages/ui/src/components/JobCard/Timeline/Timeline.module.css
felixmosh: Not relevant to this PR

ivnnv:
> @ivnnv think about it, I want the current view (without the clutter of collase expand at all... maybe only the change of the extraction of the id location to the head).
> 
> When I enable the `collapse job` I expect it to collapse all jobs (which is what happining if you refresh the page).

No yeah, I get this point, its def doable and expected if all of the other settings work real-time 👍🏽

packages/ui/src/components/JobCard/JobCard.tsx
ivnnv: done

packages/ui/src/components/JobCard/JobCard.tsx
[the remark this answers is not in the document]
ivnnv: I reworked the logic to make it useEffect-free and handle the global state "live":
it default to the state defined globally, if theres any state change at job level, this is preserved, even if the global state changed, because now the job uses the local as the preferred:


https://github.com/felixmosh/bull-board/assets/23552631/0daaa98a-d189-4518-8be4-e6351fd6d86e

packages/ui/src/components/JobCard/Timeline/Timeline.module.css
ivnnv: I think things look a bit "disconnected" from each other with such big gaps:

With current spacings:
curr


With proposed spacings:
proposed

But understand its technically out of the scope of this PR if its purelly on adding the collapsible functionality.

I have reverted it, hope you now find the same after this comparison and consider making items closer on a latter PR.

ivnnv:
@felixmosh this should be ready as per your requested changes now

PR #699, opened 2024-03-11
improved `useActiveQueue` to reduce prop depencency

ivnnv:
@felixmosh just the hook improvement and prop reducing in this PR as agreed 👍🏽


felixmosh:
Thank you for the PR, it looks awsome :]

Issue #607, opened 2023-07-14
global prefix doesn't work


codeweb05:
if I set global prefix on the api, the bull board ui doesn't work

app.setGlobalPrefix('api')

![Screenshot from 2023-07-14 16-39-52](https://github.com/felixmosh/bull-board/assets/38810092/a615b12c-f625-4cce-9f22-5538c3e783ee)

Kirkify:
Had a similar problem, but this was fixed in version 5.5.2 (See this [PR](https://github.com/felixmosh/bull-board/pull/592)) Follow this example [here](https://github.com/felixmosh/bull-board/tree/master/examples/with-nestjs).  When he sets the base path of the server adapter [here](https://github.com/felixmosh/bull-board/blob/master/examples/with-nestjs/src/queues/queues.module.ts#L58) Line 58.  Update it with your base path (so if it's 'api' like you wrote above then `/api/queues` and it should work.

felixmosh:
@codeweb05 does @Kirkify response solved your issue?

felixmosh:
@codeweb05, does #639 solves your issue?

henrysachs:
this issue is still present for me today. I just set the `/` before`api` and now it works :)

NatashkinAlex:
I have some problem(

vvtri:
> ~I'm using the [with-nestjs-module](https://github.com/felixmosh/bull-board/tree/master/examples/with-nestjs-module) example.~
> 
> ~Is it not possible to use the `routes` & `boardOptions` in the `BullBoardModule.forRoot` params along with the global prefix instead of creating a separate module just for config?~
> 
> ~going through the package, this [line](https://github.com/felixmosh/bull-board/blob/902bdb5709d487e55380b1fd963189387ab2d5ec/packages/nestjs/src/bull-board.root-module.ts#L21), implies that global prefix is respected but I get the same loading error too :(~
> 
> I just missed an "s"

I have the same problem when using nestjs with fastify adapter, global prefix and nestjs bull board module. Can you show me how you fix this ?

selabie68:
I have this same issue when using fastify.

Adding `/` before `api` does not fix the issue for fastify, it might fix it for express only.

Seems related to [this line](https://github.com/felixmosh/bull-board/blob/c3d4210ce37e6509aab57cee739756fe6a28d0d3/packages/nestjs/src/bull-board.root-module.ts#L32) where on the fastify adapter we are overriding the `prefix` option.

[...]

PR #703, opened 2024-03-21
Revert #592

dimbslmh:
Maybe something changed recently, but PR https://github.com/felixmosh/bull-board/pull/592 is actually causing issue https://github.com/felixmosh/bull-board/issues/607.

Reverting that change will make PR https://github.com/felixmosh/bull-board/pull/639 unnecessary.

The problem mentioned in PR https://github.com/felixmosh/bull-board/pull/591 doesn't happen anymore.

### DIFF
--- a/packages/nestjs/src/bull-board.root-module.ts
+++ b/packages/nestjs/src/bull-board.root-module.ts
@@ -17,8 +17,7 @@ export class BullBoardRootModule implements NestModule {
   }
 
   configure(consumer: MiddlewareConsumer): any {
-    const globalPrefix = this.applicationConfig.getGlobalPrefix() || '';
-    this.adapter.setBasePath(`${ globalPrefix }${ this.options.route }`);
+    this.adapter.setBasePath(this.options.route);
 
     if (isExpressAdapter(this.adapter)) {
       return consumer


felixmosh:
Hi, thank you for this PR.
Since I'm not familiar with Nest.js at all, how can I verify that this PR solves mentioned issues?

dimbslmh:
> Hi, thank you for this PR. Since I'm not familiar with Nest.js at all, how can I verify that this PR solves mentioned issues?

I'll try and setup a condesandbox for you later today.

dimbslmh:
@felixmosh actually.. this PR isn't the solution.

Apparently #592 only applies to Express and #639 adds Fastify support. But both lack a leading forward slash on the prefix to load the static folder correctly.
It was mentioned here https://github.com/felixmosh/bull-board/pull/592#issuecomment-1628526095 and again in #607.

NestJS + Express: https://codesandbox.io/p/devbox/vigilant-tdd-nhjg2r
Preview: https://nhjg2r-3000.csb.app/api/queues

NestJS + Fastify: https://codesandbox.io/p/devbox/naughty-kepler-w38fhq
Preview: https://w38fhq-3000.csb.app/api/queues

`main.ts`:  You can change the global prefix here with/without a leading slash. Or comment out the `setGlobalPrefix` method to see that it still works without it.

`app.module.ts`: Change the route to something else if you want.

Both use `"@bull-board/nestjs": "github:dimbslmh/bull-board-nestjs",` it makes sure there is a leading forward slash on the prefix.

Would you like that fix made in this PR or in a new one?

dimbslmh:
Closed in favor of #704

PR #739, opened 2024-05-10
Feature: Group Queues by Category

manfe:
This PR introduces:

- [x] Queue categories, but keeping backwards compatibility setting a default category.
- [x] Search Provider so we can reflect search results in all parts of the app.

How it works:

https://github.com/felixmosh/bull-board/assets/971608/f4c8f998-86bc-457e-8d30-d546009209ab


packages/api/src/queueAdapters/base.ts
```diff
@@ -24,12 +25,17 @@ export abstract class BaseAdapter {
...
     this.description = options.description || '';
+    this.category = options.category || 'default';
   }
 
   public getDescription(): string {
     return this.description;
   }
 
+  public getCategory(): string {
+    return (this.category || 'default').toUpperCase();
```
felixmosh, line 36: Why the uppercase?

felixmosh (changes requested):

felixmosh:
Hi @manfe thank you for the PR.
It is a great feature, but I would like you to make small changes.

1. Queue search should remain on the left menu (as it was) (your position dosn't make any sense)
2. Full backward compatability, to keep the current view (if there is only one category, show the corrent view, without collapse & group)
3. In order to make the code review easier for me, split this PR into several small PRs 🙏🏽,
4. Don't change `example.ts`, but add a new example in the `examples` folder.

packages/api/src/queueAdapters/base.ts
manfe: The ideia is to normalize the data, otherwise `Category 1` and `category 1` will be distinct.

manfe:
@felixmosh regarding your points.

1. With this PR the search turned global on app and I also want address some responsiveness. The search at top bar from UX works better, because the search is reflecting in the overview page. And for responsiveness the sidebar can be hidden without losing of search feature.

2. Originally I did that, but I thought that just a `default` word wouldn't be a problem, and also it was going to motive people using categories to organize their queues. PS: when there is only one category it always start as expanded), but if you really think it's a big deal, I can address it.

3. Ok, I am going to do these steps: 

- [ ] PR to add category property
- [ ] PR for global search
- [ ] PR for dropdown categories [this will be a little complex anyway]

4. I am going to create it, with Category Property PR

manfe:
How I am imagining the UX improvement:

image

packages/api/src/queueAdapters/base.ts
manfe: I am going to remove uppercase here and treat it in another leve, since the `category` property at base adapter is public, otherwise `getCategory` and `class.category` can return different data.

manfe:
After having https://github.com/felixmosh/bull-board/pull/740 and https://github.com/felixmosh/bull-board/pull/741

We can discuss the best implementation for UI/UX part.

That's work for you @felixmosh ?

felixmosh:
Hi thanks, I'm AFK till Wednesday.
Will review ASAP

PR #742, opened 2024-05-15
feat: toggle failed queues only

tsirilson:
Added a toggle to filter out non-failing queues.

Why?

When using Bull Board for production monitoring purposes, it's very useful when you want to know about failing jobs (not as a substitute for alerts).

In my case, I'm displaying it on a TV in my company's office and we have 20 different queues currently, which makes it hard to follow if any of them have failures, especially if one of the queues has a lot of successful jobs, so it's easy to miss that there's one failed job out of a thousand successful ones.

Filtering successful ones out should help us focus on the board only when we notice that there are any failing queues.

image

### DIFF
--- a/packages/ui/src/components/SettingsModal/SettingsModal.tsx
+++ b/packages/ui/src/components/SettingsModal/SettingsModal.tsx
@@ -26,6 +26,7 @@ export const SettingsModal = ({ open, onClose }: SettingsModalProps) => {
     collapseJobOptions,
     collapseJobError,
     defaultJobTab,
+    showOnlyFailedQueues,
     setSettings,
   } = useSettingsStore((state) => state);
   const { t } = useTranslation();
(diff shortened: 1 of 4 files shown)


felixmosh:
@tsirilson thank you for this PR, 
It looks like a overview page feature, so it probably should be implemented inside it.
In addition, this feature can't be tailor only for your needs, but should be more general, so maybe creating a modal that let's you to choose which statuses should be filtered by, and then the overview page will show queues with those statuses.

Will you be able to update your PR?

tsirilson:
@felixmosh Great point. What if instead of a modal, I convert the status names into clickable "buttons" that toggle the status on and off? It will also be less intrusive and won't require creating a modal for this feature only. There can also be a tooltip when hovering a status name that hints what will happen if you click it.

What do you think?

manfe:
Instead of creating a setting for that. I would create a route to receive a `status` as a parameter and then only show the queues with that status on that page.

on `App.tsx` would create:

```tsx
<Route path="/queue/by-status/:status" render={() => <QueueByStatusPageLazy />} />
```

It would be a copy of `OverviewPage` but filtering the queues by the corresponding status.

Also the statuses on Overview page could be turned into links, redirecting to this page.

felixmosh:
> @felixmosh Great point. What if instead of a modal, I convert the status names into clickable "buttons" that toggle the status on and off? It will also be less intrusive and won't require creating a modal for this feature only. There can also be a tooltip when hovering a status name that hints what will happen if you click it.
> 
> What do you think?

Sounds good to me 🙏🏼

tsirilson:
@manfe Interesting approach. Direct links is something I would also prefer, the option to link to a specific job is something I find very useful since it was added.

@felixmosh which do you think would benefit users the most? Same page filters or dedicated routes?

felixmosh:
It can be both, it is a matter of implementation.
Overview page can a query param with a list of statuses to show, by default shows all.

Issue #743, opened 2024-05-16
Datetime format


HitomaruKonpaku:
Hi there, is it possible to change datetime format to `YYYY-MM-DD hh:mm:ss` (24 hours) format?

Maybe in `boardOptions` 🤔 

<img width="172" alt="image" src="https://github.com/felixmosh/bull-board/assets/9608105/f289689a-cf0d-48f7-81c4-1a97bd3748bd">

felixmosh:
Hi, currently it formats for en, maybe you need to translate the app to your locale... :)

HitomaruKonpaku:
I'm using it for `en` too

It just that I prefer that format (like you can change in `dozzle`)

<img width="572" alt="image" src="https://github.com/felixmosh/bull-board/assets/9608105/8cb6d7df-27a7-49af-a716-4fe48db6a743">

Maybe I could check around and update to `boardOptions` if possible

andreasciamanna:
I know there is a link to the PR, but I systematically miss them in all the clutter.

So, for the future me stumbling again against this problem, add something like that to `boardOptions.uiConfig`:

```ts
dateFormats: {
  short: `hh:mm:ss`,
  common: `MMM-dd hh:mm:ss`,
  full: `yyyy-MM-dd hh:mm:ss`,
},
```

PR #745, opened 2024-05-17
feat: custom job timeline date format

HitomaruKonpaku:
#743

- New config `dateFormats` in `UIConfig`
  - `short` : When timestamp is in same day (today)
  - `common` : When timestamp is in same year
  - `full`
 
Seem like `Timeline` is the only place that have date/time format, I didn't check other page/component

### DIFF
--- a/packages/api/typings/app.ts
+++ b/packages/api/typings/app.ts
@@ -213,13 +213,23 @@ export type FavIcon = {
 export type DateFormats = {
   /**
    * When timestamp is in same day (today)
+   *
+   * @example `hh:mm:ss`
+   * @see https://date-fns.org/v3.6.0/docs/format
    */
   short?: string;
 
   /**
    * When timestamp is in same year
+   *
+   * @example `MM-dd hh:mm:ss`
+   * @see https://date-fns.org/v3.6.0/docs/format
    */
   common?: string;
 
+  /**
+   * @example `yyyy-MM-dd hh:mm:ss`
+   * @see https://date-fns.org/v3.6.0/docs/format
+   */
   full?: string;
 }
--- a/packages/api/typings/app.ts
+++ b/packages/api/typings/app.ts
@@ -202,9 +202,24 @@ export type UIConfig = Partial<{
   miscLinks: Array<IMiscLink>;
   favIcon: FavIcon;
   locale: { lng?: string };
+  dateFormats?: DateFormats;
 }>;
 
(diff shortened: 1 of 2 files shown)


felixmosh:
In my opinion this field should be inside locale field, WDYT?

HitomaruKonpaku:
> In my opinion this field should be inside locale field, WDYT?

Hmm I'm not really sure about that.

Will follow your call, but imo it isn't really related to locale since it will now follow the provided format without locale info

felixmosh:
It is a locale matter, like language (the lng property), no?

HitomaruKonpaku:
I think it is more of a standard (like ISO 8601) 🤔 

Anyway, do you wish to move it into `locale` ? 😄

felixmosh:
I'm consulting with you 😄 

You convinced me, I will review it soon.
Thank you for the PR

felixmosh:
I've reviewed the PR. I don't like the fact that you are using `dateFns.format`. can we make some map from strings to `Intl` config?

HitomaruKonpaku:
The thing with `Intl.DateTimeFormat` is that you can not specify exact format (or maybe I don't know how, please help me on this)

Like with this options (`Intl.DateTimeFormat(locale, options)`), based on `locale` config of the app (`en-US`)

- The result will be `05/18/2024, 04:47:24`
- If I try to have `locale` = `ja`, it will be `2024/05/18 04:47:24`
- But the format I expected is `2024-05-18 04:47:24`

```
    {
      year: 'numeric',
      month: '2-digit',
      day: '2-digit',
      hour: '2-digit',
      minute: '2-digit',
      second: '2-digit',
      hour12: false,
      formatMatcher: 'basic',
    }
```

felixmosh:
Hi, maybe the `undefined` option here will help you?
```js
console.log(event.toLocaleDateString(undefined, options));
// expected output (varies according to local timezone and default locale): Thursday, December 20, 2012
```
taken from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString

HitomaruKonpaku:
Hi there @felixmosh 

This is the result I got

image

felixmosh:
Ok, I see that it doesn't solves your issue.
I will approve it, and will refactor the it a bit, thank you 🙏

felixmosh:
Released in `v5.18.2`

PR #748, opened 2024-05-21
bump ejs dependency in package.json

### LINKED ISSUE #747: Bump `ejs` in package.json
In the latest release `ejs` package was updated due to a vulnerability, but not in the package.json, so installing the bull-board might still be with the vulnerable version of `ejs`.
I would like to bump the version in the package.json files to mandate it for installs

bsShoham:
closes #747


felixmosh:
Thanx

PR #749, opened 2024-05-22
Round Progress percent (JobCard)

antoinerousseau:
Here is an example showing 18.18181818%:

Screenshot 2024-05-22 à 15 51 26

We don't need the decimals. Or we could keep one or two if you prefer.


felixmosh:
Thank you :]

felixmosh:
Released in `v5.18.1`

PR #751, opened 2024-05-22
Close express security vulnerability (4.19.2) 

AdlerJS:
Per SYNK we should upgrade express to 4.19.2

image


felixmosh:
Thanx, pay attension that this is probably an issue of you lock file, since the usage of `^` should get latest v4

PR #756, opened 2024-05-28
update i18n

Diluka:
1. complete pt
2. add zh
3. use browser language as default
4. add language switch in settings

### DIFF
--- a/packages/ui/src/index.tsx
+++ b/packages/ui/src/index.tsx
@@ -18,7 +18,7 @@ const uiConfig = JSON.parse(
   document.getElementById('__UI_CONFIG__')?.textContent || '{}'
 ) as UIConfig;
 
-initI18n({ lng: uiConfig.locale?.lng || 'en-US', basePath }).then(() => {
+initI18n({ lng: uiConfig.locale?.lng || navigator.language || 'en-US', basePath }).then(() => {
   render(
     <BrowserRouter basename={basePath}>
       <UIConfigContext.Provider value={uiConfig}>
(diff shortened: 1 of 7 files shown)


packages/ui/src/index.tsx
```diff
@@ -18,7 +18,7 @@ const uiConfig = JSON.parse(
   document.getElementById('__UI_CONFIG__')?.textContent || '{}'
 ) as UIConfig;
 
-initI18n({ lng: uiConfig.locale?.lng || 'en-US', basePath }).then(() => {
+initI18n({ lng: uiConfig.locale?.lng || navigator.language || 'en-US', basePath }).then(() => {
```
felixmosh, line 21: This may break translations since we support only 3 languages

Diluka: https://github.com/felixmosh/bull-board/blob/8856fb5277f43b8bbe3be10289c54a3e9a99d180/packages/ui/src/services/i18n.ts#L22

What I understand about this configuration is that if the incoming language does not exist, en-US will be used instead, right?

felixmosh: Yeap, correct...

felixmosh, line 21: What will happen if you change the language from settings & refresh the page?
It looks like we are not using the settings value again

Diluka: add a commit to S/L lang settings

felixmosh, line 21: Is it possible to use the settings value first, then the global ui-config?

Diluka: tried, but can't use hooks here

felixmosh: How about using the store it self (I'm not sure if zustand supports that)

felixmosh (approved):

felixmosh:
Thank you for this PR 🙏🏽

PR #758, opened 2024-05-29
fix: days wrong if over 1 month

lzj960515:


felixmosh:
Hi @lzj960515 can you elaborate?

mariuszste:
I think there is a bit more work to do here. Passing a duration to js `Date` (it takes a unix timestamp) is just a bad idea in general because a duration is fundamentally not a date. The current solution just happens to work for a while since `0` is january first 1970 at midnight, but as soon as the uptime is longer than a month it switches to february and breaks.

Personally not a fan of reinventing the wheel when dealing with time. `luxon` is already a dependency of `cron-parser` and both `bull` and `bullmq` depend on it. So might as well use a battle tested solution that handles all the edge cases and even can do formatting using browser's native intl, so we get multi language support for free

felixmosh:
@mariuszste thank you for the explanation, this piece of code is old...
I've moved formatting to the client, so it will be localized.
image
Thanks for pointing this out

PR #762, opened 2024-06-05
Hide bull routes from schema/openAPI generation

bcomnes:
This hides the bullboard routes from openAPI schema generation. These routes don't have a schema defined, so they aren't useful in that context.

### DIFF
--- a/packages/fastify/src/FastifyAdapter.ts
+++ b/packages/fastify/src/FastifyAdapter.ts
@@ -122,6 +122,9 @@ export class FastifyAdapter implements IServerAdapter {
         fastify.route({
           method,
           url,
+          schema: {
+            hide: true
+          },
           handler: (_req, reply) => {
             const { name, params } = handler({ basePath: this.basePath, uiConfig: this.uiConfig });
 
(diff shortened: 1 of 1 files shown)


bcomnes:
I think it's better to just hide them. It seems unlikely any fastify server exposing a published openAPI spec based on its schema would also want to include unreleated bull-board routes in that spec. It seems much more likely of a scenario that someone has created an API with fastify, and wants to add bullboard as a backend management tool and node expose that detail to their public API docs via openAPI. 

If there does happen to be *someone* out there who does want it exposed, perhaps we could leave that option/schema as an open contribution task for them.

bcomnes:
Thanks for the quick turnaround! Enjoying bullboard so far.

felixmosh:
Hi, this PR breaks the build, looks like fastify schema doesn't have this `hide: boolean` prop.
image

I will revert this PR, in order to allow publishing other PR's.
Can you check this?

bcomnes:
Its supplemented in when you add the first party openapi plugin. Otherwise its ignored. 

https://github.com/fastify/fastify-swagger/blob/e01cd7ba3f0972cf3ba75f87bd3f35b2d5c1683a/index.d.ts#L34-L48

felixmosh:
Just ignore it, if someone will need the abillity to pass a schema, he/she will make a PR for it :]

Thanx

bcomnes:
Retrying here: https://github.com/felixmosh/bull-board/pull/764/files



Sorry about that.

PR #765, opened 2024-06-06
fix(README): use proper path to bullMQAdapter file

jtomaszewski:
I've just installed bull board by following the README example, and somehow building the project was failing to me with following message:

```

src/main/worker.module.ts:19:31 - error TS2307: Cannot find module '@bull-board/api/bullMQadapter' or its corresponding type declarations.

19 import { BullMQAdapter } from '@bull-board/api/bullMQadapter';
                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

After a bit of inspecting I've found out that import path (that I've took from the README example) is incorrect.


felixmosh:
Thank you 🙏🏽

PR #768, opened 2024-06-07
fix(hono): safely parse request body

marbemac:
The web UI does not include a request body in most of its requests to the API. Hono's `c.req.json()` method throws if it cannot parse the body (at least on v4, I noticed bull-board is still on v3). This causes all of the mutations in the UI to fail, except for those that include a body such as adding a job.

This PR is a small change to safely attempt to parse the request body.


felixmosh:

PR #772, opened 2024-06-15
fix(ui): correctly calculate the circumference and offset of the progress citcle

### LINKED ISSUE #771: Progress circle filled when progress < 100%
When the progress is > 90%, the circle is complete. This should only happen at 100%.

image

ChatGPT discusses why this is happening here: https://chatgpt.com/share/13b753fa-edc7-43dc-823c-b2cca392475d

henryStelle:
Closes #771 

The `stroke-dasharray` and `stroke-dashoffset` were improperly calculated, resulting in the progress circle "completing" when the percentage was only 90%.

Current version:

image

I refactored the styling to make it more declarative. I also removed the rounded edge on the border so there is a gap with the percentage > 98%.

Fixed version:

image

### DIFF
--- a/packages/ui/src/components/JobCard/Progress/Progress.module.css
+++ b/packages/ui/src/components/JobCard/Progress/Progress.module.css
@@ -6,6 +6,12 @@
 .progress circle {
   transform-origin: center;
   transition: stroke-dashoffset 500ms ease-in-out;
+  cx: 70;
+  cy: 70;
+  r: 65;
+  fill: none;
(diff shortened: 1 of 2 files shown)


felixmosh:
Hi, thank you for this PR 🙏🏼
Can you make sure it works properly in Safari & Firefox?

henryStelle:
Good call, firefox didn't work, but now it does!

felixmosh:
Released in `v5.20.2`
Thank you

PR #778, opened 2024-06-24
feat(adapters): support hono v4

nihalgonsalves:
Support Hono V4. Since the adapters are not versioned separately, I tried to make a minimally-breaking change - it requires no changes unless you use Cloudflare Workers, which has a breaking change in Hono v4[^1]. 

This makes the types in the `HonoAdapter` slightly unsound. A proper long-term fix would be to change the API to accept `handler: ({ root, rewriteRequestPath }) => MiddlewareHandler` (where `MiddlewareHandler` is the return type of `serveStatic`).

Consumers could then construct `serveStatic` themselves, including requirements such as the manifest. That would, of course, require _every_ consumer to update their `HonoAdapter` constructor, so maybe something to reserve for a major release.

Closes #721 (supersedes)

[^1]: https://github.com/honojs/hono/blob/main/docs/MIGRATION.md#servestatic-in-cloudflare-workers-adapter-requires-manifest

### DIFF
--- a/packages/hono/package.json
+++ b/packages/hono/package.json
@@ -31,7 +31,7 @@
     "@bull-board/api": "5.20.2",
     "@bull-board/ui": "5.20.2",
     "ejs": "^3.1.10",
-    "hono": "^3.12.0"
+    "hono": "^4.4.7"
   },
   "devDependencies": {
     "@cloudflare/workers-types": "^4.20231218.0",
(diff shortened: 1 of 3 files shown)


felixmosh:
Thank you for this PR 🙏🏽

dalechyn:
thanks king, right about time i started looking into it

Issue #789, opened 2024-08-03
Feature Request:  Ability to disable/enable polling interval option from backend


jb-abhi:
Problem
There is a settings option provided in UI through which user can select the polling interval. But the issue is that if the user selects a shorter polling interval, say 3-5 sec, then there will be an increased load on the backend.

Suggestion
Can add another optional option in UI Config Type with the name polling which can be set to off, which can then be used in UI to show/hide the polling interval option

Reasoning
This UI can be used by QA, Devs, clients etc. to monitor the jobs. If client or QA selects a shorter polling option, then that may incur additional requests and charges for different hosting services like azure, aws etc.

felixmosh:
Do you think that this should be under "read only" mode?

jb-abhi:
yes, it would be great that way

felixmosh:
I've added a new `uiConfig` for this.

```js
  createBullBoard({
    queues: [new BullMQAdapter(exampleBullMq), new BullAdapter(exampleBull)],
    serverAdapter,
    options: {
      uiConfig: {
        pollingInterval: {
          forceInterval: 5, // any valid value, -1 for off.
          showSetting: false, // hides the setting option from Settings modal
        },
      },
    },
  });
```

jb-abhi:
Oh, nice. Thanks a ton. 😊

PR #790, opened 2024-08-04
marked polling-interval option to off and disabled other polling options if readonly mode is enabled

jb-abhi:
Based on feature request 789, disabled polling options other than "Off" if in "read only" mode in the Settings Modal.


felixmosh:
Thank you for this PR 🙏🏽

felixmosh:
closes #789

PR #792, opened 2024-08-04
Fix: Polling Interval not being updated to "off" when selecting a queue.

jb-abhi:
Scenario: From the dashboard, select the polling interval as 3 sec, then select a read-only queue, then you can see that the polling will continue in 3 sec interval, although in the settings modal the polling intervals are hidden and set to off. 

Fix: Added a use effect, which will trigger when queue is set to read-only and update the polling interval to off, instead of only hiding the options


felixmosh:
Ham... this is weird, since "off" in polling interval means that the board won't fetch updates at all...
This is not the intented behaviour... readOnly mode is ment for cases that the board is given to none thechnical users, which the board should prevent them from making actions, but, the board should update it's data.

felixmosh:
I'm closing this PR since setting the poll interval to -1 will disable the update.
More over, I'll revert the previous change, since it will break the default behaviour.

PR #799, opened 2024-08-22
Ease of use helpers for config led dynamic queue defs with compose wrapper.

colinjonesx:
I needed these things to make bull-board more usable in my use case. 

Hope it'll help adoption.

Not sure if it can be templatised, to apply to the other server patterns?


examples/with-fastify/index.js
```diff
@@ -64,15 +81,14 @@ const run = async () => {
...
+  await app.listen({ host: serverOptions.host, port: serverOptions.port });
   // eslint-disable-next-line no-console
-  console.log('Running on 3000...');
-  console.log('For the UI, open http://localhost:3000/ui');
-  console.log('Make sure Redis is running on port 6379 by default');
+  console.log(`For the UI, open http://localhost:3333/ui`);
+  console.log('Make sure Redis is configured in env variables. See .env.example');
   console.log('To populate the queue, run:');
-  console.log('  curl http://localhost:3000/add?title=Example');
+  console.log('  curl http://localhost:3333/add?title=Example');
```
felixmosh, line 89: it should use the serverOptiosn.host & port, no?

examples/with-fastify/docker-compose.yml
```diff
@@ -0,0 +1,8 @@
+services:
+  bullboard:
+    build: .
+    env_file:
+      - .env
+
+    ports:
+      - 3333:3000
```
felixmosh, line 8: map `3000:3000`

felixmosh (changes requested):

examples/with-fastify/index.js
colinjonesx: Good challenge!

Sorry - an oversight, the exposed port is configured in docker-compose.yml, which is pinned to 3333 in config.

I should remove serverOptions and hardcode to 0.0.0.0 and 3000 as this is implicit in compose config

examples/with-fastify/docker-compose.yml
felixmosh: What about this?
Can you change the 3333 to 3000 in all the places?

examples/with-fastify/index.js
colinjonesx: Latest commit addresses the issue

felixmosh: I still see usage of `3333` port, why?

colinjonesx: Port `3000` is an arbitrary albeit conventional port for nodejs applications. From my perspective moving the exposed port to anything is a reasonable thing to do, especially where the primary application (the one posting to or subscribing to the MQ is likely going to reside on that port.  With bull-board, as pretty as it is, being an ancillary support app, it seems to be a logical move.

If you would like me to revert to `3000` I can, but obvs I'll be running on `3333` locally.

felixmosh:
Thank you 🙏🏼

PR #816, opened 2024-09-13
Center align logo image in header

bamadesigner:
Hi! 

This is a minor tweak to the CSS for the header logo and image. It allows the image to be center-aligned vertically, and the image max height is set to the height of the container. This change allows users to set whatever height and width they want for the image, and it will auto-scale in the container.

Thanks!

### DIFF
--- a/packages/ui/src/components/Header/Header.module.css
+++ b/packages/ui/src/components/Header/Header.module.css
@@ -27,11 +27,13 @@
   box-shadow: 0 2px 5px 0 rgba(0, 0, 0, 0.1);
   z-index: 2;
   display: flex;
+  align-items: center;
   padding: 0 1rem;
   text-decoration: none;
 }
 
 .header > .logo > .img {
+  max-height: 100%;
   margin-right: 0.4em;
 }
 
--- a/packages/ui/src/components/Header/Header.module.css
+++ b/packages/ui/src/components/Header/Header.module.css
@@ -34,6 +34,7 @@
 
 .header > .logo > .img {
   max-height: 100%;
+  width: auto;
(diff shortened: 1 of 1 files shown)


felixmosh:
Hi @bamadesigner thank you fo this PR.
Can you share some images with the previous state and the new one?

bamadesigner:
Of course!

Here is the previous state if I don't have a width or height set on the logo:
Screenshot 2024-09-14 at 11 21 20 AM

Here is the previous state if I set a width and height on the logo:
Screenshot 2024-09-14 at 11 23 56 AM

Here is the new state with the `align-items: center` set on the `<a>` element:
Screenshot 2024-09-14 at 11 24 39 AM

You could only add the `align-items: center` to the `<a>` element and that would fix the main vertical align issue.

But as a helpful bonus, you can also add `width: auto` and `max-height: 100%` to the `` element, and it makes a better layout for any images that are set to be bigger than the space.

When I first added my image, I set the width and height as 100px because I had not yet determined that the header container was hard-set to 80px tall:
Screenshot 2024-09-14 at 11 35 13 AM

So adding the `width: auto` and `max-height: 100%` to the `` element resolved that:
Screenshot 2024-09-14 at 11 41 06 AM

But, really, I can live without the `width: auto` and `max-height: 100%` and just set 50px for the width and height of my image, as long as I have the `align-items: center` on the `<a>` element:
Screenshot 2024-09-14 at 11 39 15 AM

felixmosh:
Thank you @bamadesigner for the contribution 🙏🏼

bamadesigner:
Thank you @felixmosh !

PR #828, opened 2024-10-09
Show percentage when progress is an object

difagume:
I am using bull board with bullMQ, in the job progress I am sending an object e.g.: 
`{ progress: 70, total: 100, completed: 50 }`

when I set the job progress as an object the progress percentage is no longer displayed in bullboard.

I updated the component to show the progress whether it is a number or an object.

I think this change will be beneficial to many people who pass an object in the job progress.

### DIFF
--- a/packages/ui/src/components/JobCard/Progress/Progress.tsx
+++ b/packages/ui/src/components/JobCard/Progress/Progress.tsx
@@ -5,7 +5,7 @@ import { Status } from '@bull-board/api/typings/app';
 import { STATUSES } from '@bull-board/api/src/constants/statuses';
 
 interface ProgressProps {
-  progress: number | object;
+  progress: number | { progress?: number };
   strokeWidth?: number;
   status: Status;
   className?: string;
@@ -19,11 +19,9 @@ export const Progress = ({ progress, status, className, strokeWidth = 6 }: Progr
     strokeWidth,
     ['transform-origin']: 'center',
   };
-  
-  const getPercentage = (): number => {
-    if (typeof progress === 'number') return progress;
-    return (progress as { progress?: number }).progress ?? 0;
-  };  
+
+  const percentage = typeof progress === 'number' ? progress : progress.progress ?? null;
+  if(!percentage) return null
 
   return (
     <svg className={cn(s.progress, className)} width="100%" height="100%">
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/JobCard/Progress/Progress.tsx
```diff
@@ -5,35 +5,41 @@ import { Status } from '@bull-board/api/typings/app';
 import { STATUSES } from '@bull-board/api/src/constants/statuses';
 
 interface ProgressProps {
-  percentage: number;
+  progress: number | object;
```
felixmosh, line 8: Please change the `object` to `{progress?: number}`

packages/ui/src/components/JobCard/Progress/Progress.tsx
```diff
@@ -5,35 +5,41 @@ import { Status } from '@bull-board/api/typings/app';
...
     cx: '50%',
     cy: '50%',
     r: `calc(50% - ${strokeWidth / 2}px)`,
     strokeWidth,
     ['transform-origin']: 'center',
   };
+  
+  const getPercentage = (): number => {
+    if (typeof progress === 'number') return progress;
+    return (progress as { progress?: number }).progress ?? 0;
```
felixmosh, line 25: Remove the type cast

packages/ui/src/components/JobCard/Progress/Progress.tsx
```diff
@@ -5,35 +5,41 @@ import { Status } from '@bull-board/api/typings/app';
...
+export const Progress = ({ progress, status, className, strokeWidth = 6 }: ProgressProps) => {
   const commonProps = {
     cx: '50%',
     cy: '50%',
     r: `calc(50% - ${strokeWidth / 2}px)`,
     strokeWidth,
     ['transform-origin']: 'center',
   };
+  
+  const getPercentage = (): number => {
```
felixmosh, line 23: why do you need it as a function?
Calculate the value, and use it on both places.

felixmosh (changes requested):

felixmosh:
Thank you for reporting this issue.

packages/ui/src/components/JobCard/Progress/Progress.tsx
difagume: I have made the requested changes

PR #840, opened 2024-10-21
feat: add status filters on overview page

arnaudvalle:
Following the discussion on https://github.com/felixmosh/bull-board/pull/742, I added some simple links to the `StatusLegend` so it's now possible to filter queues based on a given status. This can be useful to direct link to a list of failed queues from some alerting system for example:

image

I've tried to reuse as much as I could from existing components but I'm unsure whether you tend to duplicate similar things or not to future-proof them (for example I've reused css vars from `status-menu` in the StatusLegend module css but maybe that's an anti-pattern for you?).

Also, a current known limitation to this is that there is no `ALL` link (but we can click on the logo for example to go back to the initial state) as I wasn't sure on how best to handle this (I'm a first-time contributor to this project so I'm not comfortable with the architecture of it all just yet). Would adding a new status and associated translation for this make sense?

### DIFF
--- a/packages/ui/src/components/StatusLegend/StatusLegend.module.css
+++ b/packages/ui/src/components/StatusLegend/StatusLegend.module.css
@@ -6,18 +6,55 @@
   text-transform: uppercase;
 }
 
-.legend > li:before {
+.legend > li {
+  display: flex;
+  align-items: center;
+  border-bottom: 2px solid transparent;
+}
+
+.legend > li.isSelected {
+  border-color: var(--status-menu-active-border);
+  color: var(--status-menu-active-text);
+  font-weight: 500;
+}
+
+.legend > li > a {
+  text-decoration: none;
+  margin: 0 0 -2px;
+  padding: 0.5em 1em;
+  color: var(--status-menu-text);
+  transition: border-bottom-color 200ms ease-in-out, color 0.1s ease;
+  display: flex;
+  align-items: center;
+}
+
+.legend > li > a:before {
   content: '';
   background-color: var(--item-bg);
   border-radius: 50%;
   width: 0.5rem;
   height: 0.5rem;
-  display: inline-block;
+  display: flex;
   margin-right: 0.5rem;
 }
 
-.legend > li + li {
-  margin-left: 2rem;
+.legend > li > a span {
+  flex: 1;
+  white-space: nowrap;
+}
+
+.legend > li > a span:before {
+  display: block;
+  content: attr(title);
+  font-weight: 600;
+  height: 0;
+  overflow: hidden;
+  visibility: hidden;
+}
+
+.legend > li > a:hover,
+.legend > li > a:active {
+  border-color: #7a8085;
 }
 
 .waiting {
--- a/packages/ui/src/components/StatusLegend/StatusLegend.module.css
+++ b/packages/ui/src/components/StatusLegend/StatusLegend.module.css
@@ -34,7 +34,7 @@
   border-radius: 50%;
   width: 0.5rem;
   height: 0.5rem;
-  display: flex;
+  display: inline-block;
   margin-right: 0.5rem;
 }
 
--- a/packages/ui/src/components/StatusLegend/StatusLegend.tsx
+++ b/packages/ui/src/components/StatusLegend/StatusLegend.tsx
@@ -1,26 +1,27 @@
 import React from 'react';
 import { queueStatsStatusList } from '../../constants/queue-stats-status';
 import s from './StatusLegend.module.css';
-import { NavLink } from 'react-router-dom';
+import { NavLink, useLocation } from 'react-router-dom';
 import { useTranslation } from 'react-i18next';
 import { toCamelCase } from '../../utils/toCamelCase';
+import cn from 'clsx';
 
 export const StatusLegend = () => {
   const { t } = useTranslation();
+  const { search } = useLocation();
 
   return (
     <ul className={s.legend}>
       {queueStatsStatusList.map((status) => {
         const displayStatus = t(`QUEUE.STATUS.${status.toUpperCase()}`).toLocaleUpperCase();
 
-        return (<li key={status} className={s[toCamelCase(status)]}>
+        const query = new URLSearchParams(search);
+
+        return (<li key={status} className={cn(s[toCamelCase(status)], {
+          [s.isSelected]: query.get('status') === status,
+        })}>
           <NavLink
             to={`/?status=${status}`}
-            activeClassName={s.active}
-            isActive={(_path, { search }) => {
-              const query = new URLSearchParams(search);
-              return query.get('status') === status;
-            }}
             className={s[toCamelCase(status)]}
             key={`overview-${status}`}
           >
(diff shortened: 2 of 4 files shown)


packages/ui/src/components/StatusLegend/StatusLegend.module.css
```diff
@@ -6,18 +6,55 @@
...
+}
+
+.legend > li > a:before {
   content: '';
   background-color: var(--item-bg);
   border-radius: 50%;
   width: 0.5rem;
   height: 0.5rem;
-  display: inline-block;
+  display: flex;
```
felixmosh, line 37: why `flex` is here?

felixmosh:
Hi @arnaudvalle, thank u for this PR.
Is it better to just filter queues in the client side? (and show all statuses of the queue)?

packages/ui/src/pages/OverviewPage/OverviewPage.tsx
```diff
@@ -3,16 +3,21 @@ import { QueueCard } from '../../components/QueueCard/QueueCard';
...
 import { useQueues } from '../../hooks/useQueues';
 import s from './OverviewPage.module.css';
+import { useSelectedStatuses } from '../../hooks/useSelectedStatuses';
 
 export const OverviewPage = () => {
   const { actions, queues } = useQueues();
   actions.pollQueues();
 
+  const selectedStatus = useSelectedStatuses();
+  const overviewPageStatus = selectedStatus[''];
```
felixmosh, line 13: `selectedStatus['']`- can u explain this statment?

packages/ui/src/components/StatusLegend/StatusLegend.tsx
```diff
@@ -1,18 +1,34 @@
...
+        const displayStatus = t(`QUEUE.STATUS.${status.toUpperCase()}`).toLocaleUpperCase();
+
+        const query = new URLSearchParams(search);
+
+        return (<li key={status} className={cn(s[toCamelCase(status)], {
+          [s.isSelected]: query.get('status') === status,
+        })}>
+          <NavLink
+            to={`/?status=${status}`}
+            className={s[toCamelCase(status)]}
```
felixmosh, line 25: There is `status` on the parent `li`, why do we need here?

packages/ui/src/components/StatusLegend/StatusLegend.tsx
```diff
@@ -1,18 +1,34 @@
...
-import { toCamelCase } from '../../utils/toCamelCase';
 import s from './StatusLegend.module.css';
+import { NavLink, useLocation } from 'react-router-dom';
+import { useTranslation } from 'react-i18next';
+import { toCamelCase } from '../../utils/toCamelCase';
+import cn from 'clsx';
 
 export const StatusLegend = () => {
   const { t } = useTranslation();
+  const { search } = useLocation();
```
felixmosh, line 11: `useQuery` instead

felixmosh (changes requested):

packages/ui/src/components/StatusLegend/StatusLegend.module.css
arnaudvalle: You're right, this is not necessary anymore. I think I added it before changing the markup of the filters.

packages/ui/src/components/StatusLegend/StatusLegend.tsx
arnaudvalle: I believe that can go as well - that was before I moved the logic out to the `<li>` so it's redundant indeed 👍

arnaudvalle:
> Is it better to just filter queues in the client side? (and show all statuses of the queue)?

So you mean if I only ever have queues with FAILED jobs (like in my screenshot) then I would only ever show the FAILED tab filter in the list and not the other ones? Something like:

image

I'm not sure about how other people use this page tbh but if you say that's ok then it's fine by me 👍

felixmosh:
> > Is it better to just filter queues in the client side? (and show all statuses of the queue)?
> 
> So you mean if I only ever have queues with FAILED jobs (like in my screenshot) then I would only ever show the FAILED tab filter in the list and not the other ones? Something like:
> 
> image
> I'm not sure about how other people use this page tbh but if you say that's ok then it's fine by me 👍

No, i ment filter queues only in client side, "show me queues that have count on selected status that is greater than 0"

arnaudvalle:
> No, i ment filter queues only in client side, "show me queues that have count on selected status that is greater than 0"

I see - I believe that's how I got started on this but somehow got lost along the way... I'll fix [the condition on the `queues.filter` in `OverviewPage`](https://github.com/felixmosh/bull-board/pull/840/files#diff-748d22eb606eda8b77cb8159283c949f58c4f67aca8ba46f93ad2fec81dcee09R20) and that should be it. That's what you mean, right?

felixmosh:
Instead of making a request with the "selected status", we can filter out all queues which has count of "selected status" > 0.
This way, we can show all other statuses of the queue but focus only on queues that has at lease one "failed" job.
Current implementation, whenever you filter, shows that the queue has only `failed` jobs (but it is not true).

arnaudvalle:
Oh so you mean move the `queues.filters` I added to the `OverviewPage` into the `useQueues` hook, right before the [setState](https://github.com/felixmosh/bull-board/blob/master/packages/ui/src/hooks/useQueues.ts#L55) for instance?

felixmosh:
No, I mean that instead of filter queues with an API call, filter it only in client side.

arnaudvalle:
> No, I mean that instead of filter queues with an API call, filter it only in client side.

Ok sorry, I must be misunderstanding something here. I thought that's already what [my change was doing in the OverviewPage](https://github.com/felixmosh/bull-board/pull/840/files#diff-748d22eb606eda8b77cb8159283c949f58c4f67aca8ba46f93ad2fec81dcee09R20)? 

The `overviewPageStatus === 'latest'` bit in the condition is because the overview page (aka `''` in the queues coming from the `useQueues` hook) defaults to a value of `latest` when no status is selected (see in [useSelectedStatus hook](https://github.com/felixmosh/bull-board/blob/master/packages/ui/src/hooks/useSelectedStatuses.ts#L8)). 

Maybe it's my fault and my screenshot was misleading to star with (since I was only showing queues with failed jobs), but when I have a queue with both completed AND failed jobs for example, I can either see the queue when I select the `completed` status OR the `failed` status:

completed is selected:
image

failed is selected:
image

no status selected:
image

status with no job in any queues is selected:
image

Thanks for your patience and feedbacks btw!

arnaudvalle:
Oh, no rush, don't worry! I'll stop bothering you now, enjoy your time off 👍

felixmosh:
I've refactor a bit the code...
Added a message when ever the filter is 0...
It will be released in the next version
Thank you 🙏🏽

arnaudvalle:
Great stuff, thanks for the cleanup 👍

tsirilson:
@arnaudvalle Thanks for this! I dropped #742 because I didn't have time to get to @felixmosh's comments. I just upgraded bullboard and discovered this feature by accident.

arnaudvalle:
@tsirilson no worries, glad it's of use to you as well! I did come across your issue and it pointed me in the right direction, so thank you as well 👍

PR #852, opened 2024-11-15
Fix h3 adapter route param decode

skirsten:
If the queue or job was encoded (e.g. because it had a `/`) it could not be found because h3 does not decode the route param by default.


felixmosh:
Thank you 🙏🏼

PR #855, opened 2024-11-20
fix(JobActions): show actions for "waiting children" jobs just as for "waiting" ones

jtomaszewski:
Currently there's no possibility to delete a job which is in "waiting children" status. Yet, such a job can be awaiting another job which is simply in "waiting" status, which can be removed... and if you do that, you'll end up with a "zombie" job in "waiting children", that will never succeed, nor can be removed through the Bull Board.

Screenshot of the current behaviour:
image

This PR fixes it, making actions appear in a "waiting children" job just as in a "waiting" one.

Fixed behaviour:

image


felixmosh:
Hi, thank you for this PR.

PR #859, opened 2024-12-16
feat: add devcontainer

spotlesscoder:
this allows setup inside modern IDEs without having to install any nodejs / third party code to your system and throw away the entire development environment when not needed anymore without leaving leftover on the host system

this is entirely optional for anyone that clones the repo
also, the extensions from the devcontainer.json are only installed, when a user installs the devcontainer


felixmosh:
Thank you for this PR

PR #860, opened 2024-12-16
Fix/with-nestjs-example

spotlesscoder:
project was not building successfully
 - @nestjs/config missing
 - ExpressAdapter type was incompatible to the used bull board version

### DIFF
--- a/examples/with-nestjs/src/queues/basic-auth.middleware.ts
+++ b/examples/with-nestjs/src/queues/basic-auth.middleware.ts
@@ -1,7 +1,7 @@
 import { Injectable, NestMiddleware } from '@nestjs/common';
 import { ConfigService } from '@nestjs/config';
-import { NextFunction, Request, Response } from 'express';
 import * as bcrypt from 'bcrypt';
+import { NextFunction, Request, Response } from 'express';
 
 @Injectable()
 export class BasicAuthMiddleware implements NestMiddleware {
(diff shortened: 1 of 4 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,21 @@
+{
```
felixmosh, line 1: Please remove this file from this PR

examples/with-nestjs/src/queues/basic-auth.middleware.ts
```diff
@@ -30,7 +31,10 @@ export class BasicAuthMiddleware implements NestMiddleware {
       return;
     }
 
-    const isPasswordValid = await await bcrypt.compare(password, this.passwordHash);
+    const isPasswordValid = await await bcrypt.compare(
```
felixmosh, line 34: why there is 2 `await` here?

felixmosh (changes requested):

.devcontainer/devcontainer.json
spotlesscoder: done

examples/with-nestjs/src/queues/basic-auth.middleware.ts
spotlesscoder: maybe I forgot to save before commit. will push a fix

felixmosh:
Thank you for you contribution  🙏🏽

spotlesscoder:
you're welcome :) thank you too for making this tool - it helps me keeping track of things :)

PR #866, opened 2024-12-27
Update H3Adapter.ts use that for unjs/nitro

Maxttier:
basePath is applied to all routes, app.use() in nuxt/nitro I think not work well

### DIFF
--- a/packages/h3/src/H3Adapter.ts
+++ b/packages/h3/src/H3Adapter.ts
@@ -30,6 +30,7 @@ export class H3Adapter implements IServerAdapter {
   private bullBoardQueues: BullBoardQueues | undefined;
   private viewPath: string | undefined;
   private uiConfig: UIConfig = {};
+  private routes: Array<any> = [];
 
   public setBasePath(path: string): H3Adapter {
     this.basePath = path;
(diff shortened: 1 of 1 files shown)


felixmosh:
Hi thanks for the PR.
What issue do you try to solve?

Makkalay:
I try to use that with @bull-board/nestjs which seems to auto call some of that !this.entryRoute),  !this.viewPath), !this.errorHandler) and !this.uiConfig) before setBasePath
``

Makkalay:
If use in nitrojs plugin i get 404 error

```
import { H3Adapter } from '@bull-board/h3';
import { eventHandler } from 'h3';
import { createBullBoard } from '@bull-board/api';
import { BullMQAdapter } from '@bull-board/api/bullMQAdapter';
import { Queue as QueueMQ, RedisOptions, Worker } from 'bullmq';

export default defineNitroPlugin((nitroApp) => {
    let serverAdapter = new H3Adapter();
    serverAdapter.setBasePath('/b');

    const redisOptions: RedisOptions = {
        port: 6379,
        host: 'localhost',
        password: '',
    };
  
    const createQueueMQ = (name: string) => new QueueMQ(name, { connection: redisOptions });

    async function setupBullMQProcessor(queueName: string) {
        new Worker(
          queueName,
          async (job) => {
            for (let i = 0; i <= 100; i++) {
                await sleep(Math.random());
                await job.updateProgress(i);
                await job.log(`Processing job at interval ${i}`);
      
                if (Math.random() * 200 < 1) throw new Error(`Random error ${i}`);
              }
      
              return { jobId: `This is the return value of job (${job.id})` };
            },
            { connection: redisOptions }
        );
    }
    const exampleBullMq = createQueueMQ('BullMQ');
    setupBullMQProcessor(exampleBullMq.name);
    createBullBoard({
        serverAdapter,
    });

    nitroApp.h3App.use(serverAdapter.registerHandlers());

    nitroApp.router.use(
        '/add',
        eventHandler(async () => {
            await exampleBullMq.add('myJob', { foo: 'bar' });
            return true;
        })
    );
})
  
```

Makkalay:
I try `http://localhost:3000/api/api/b`, `http://localhost:3000/api/b` and `http://localhost:3000/b`

Makkalay:
In your example is `app.use()` but work with router.use like `nitroApp.router.use('/**', serverAdapter.registerHandlers().handler);`. But that path '/**' will cause problems it will match other routes, wil return 204 no content for some routes and other

Makkalay:
Plus you create a second router in adapter this.uiHandler = (0, h3_1.createRouter)() but not pass existing router (created in h3 example) or not get only routes and attach its to that exisiting router (form h3 expale index.ts)

Issue #633, opened 2023-10-12
Feature Request: Group queues by prefix


JensSpanier:
It would be great if I could set an delimiter so the queues could get sorted hierarchicaly. E.g. `service1:queue1:subqueue2` and `service2:queue1` become:
![grafik](https://github.com/felixmosh/bull-board/assets/42373861/f5a6e5b7-4cbf-4ae4-afcf-bac82faf89ee)

<details>
  <summary>service1</summary>
<details>
  <summary>queue1</summary>
subqueue2
</details>
</details>
<details>
  <summary>service2</summary>
queue1
</details>

Gunacan:
Are there any plans on implementing this feature or something similar? I think this would be great addition. I am working on an event system. I have events and their handlers. An event can have many handlers. I have a dedicated queue for each handler. It would be great if I could just pass the prefix (the event name) for each handler queue and the dashboard just groups them in a directory by event.

vitalyiegorov:
Yea! We would love this feature, or some way to define groups of queues

PR #867, opened 2025-01-02
Feature: Group queues in Menu

ahnwarez:


felixmosh:
Hi, thank you for this PR.
Can you share how the ui looks?
I'll review the code ASAP

ahnwarez:
Thanks for your quick reply 🙂 

This PR implements #633 
I've always wanted to have this feature in Bull-Board where I can see all queues like a tree.

Here are notable things in this PR:
- The user decides how they name their queues and how they are delimited when creating queues and when passing the queues to `createBullBoard`
```js
const newRegistration = createQueueMQ('Notifications.User.NewRegistration');
const resetPassword = createQueueMQ('Notifications.User.ResetPassword');

createBullBoard({
    queues: [
      new BullMQAdapter(newRegistration, { delimiter: '.' }),
      new BullMQAdapter(resetPassword, { delimiter: '.' }),
    ],
    serverAdapter,
  });
```
- Expanding and collapsing is done using `details` and `summary` tags. All `details` are expanded by default.
- Added a new React component `AppQueueTree` to recursively render nodes until a leaf is found.

I welcome any feedback. Thanks to all contributors. 

https://github.com/user-attachments/assets/e78b4b82-5b25-41d4-92af-7e47c2bd03d9

packages/ui/src/utils/toTree.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export interface AppQueueTree {
+  [key: string]: {
+    children: AppQueueTree;
+    queue?: AppQueue;
+  };
+}
+
+export function toTree(array: AppQueue[]): AppQueueTree {
+  const tree: AppQueueTree = {};
```
felixmosh, line 11: There is no any enefit to use here an Object (it even deoptimizes the JS engine).

I think that It should be an array of objects that has children

It will be much simpler to traverse it

felixmosh (changes requested):
Overall the code looks awsome 🙏🏽
Thank you for this PR, it will implement a much request feature :]

ahnwarez:
Thank you for your review @felixmosh, really appreciate your feedback.
I've pushed the new changes and addressed your feedback regarding using objects.

felixmosh:
Im AFK at the weekend, will publish it on Monday.
Thank you 🙏🏽

felixmosh (approved):

felixmosh:
I've noticed a bug, the nav link links to queues final part (which is not the full queue name), therefore, it returns, Queue not found.

I'll fix it.

felixmosh:
I've improved a bit the design...
image

felixmosh:
Thank you for this PR, it released in `v6.6.0`

Diluka:
@ahnwarez I don't use delimiter but some prefixes. How can I use this feature?

felixmosh:
You can try to set the prefix to be a; , b;
Then set the delimiter to be ;

ahnwarez:
There are optional parameters to set the delimiter and in the example below it's `.`, so in your case you need to change it to `{ delimiter: '-'}`

```js
  // before
  createBullBoard({
    queues: [
      new BullMQAdapter(exampleBullMq, { delimiter: '.' }),
      new BullAdapter(exampleBull, { delimiter: '.' }),
      new BullMQAdapter(newRegistration, { delimiter: '.' }),
      new BullMQAdapter(resetPassword, { delimiter: ':' }),
    ],
    serverAdapter,
  });
 ```
 
 ```js
// after 
 createBullBoard({
    queues: [
      new BullMQAdapter(exampleBullMq, { delimiter: '-' }),
    ],
    serverAdapter,
  });
  ```
  
  I hope this answers your question 🙂

Diluka:
My queues are named as `prefix-the-rest-of-name`. the `prefix` is prefix and `the-rest-of-name` is the queue name. And I have multiple prefixes.

andreasciamanna:
This works.

However, with my queues are all wrapped with curly braces (e.g. `{my-queue-something}`), the look of the grouping is a bit ugly:

image

As of why I name queues like that, see [Hash tags](https://github.com/OptimalBits/bull/blob/develop/PATTERNS.md#redis-cluster).

Is there an option I can use to handle this case?

felixmosh:
@andreasciamanna can you open a separate bug?

smagdziarz:
Hi @ahnwarez, thanks for this great functionality, will be very handy for us!

> ```js
> new BullMQAdapter(resetPassword, { delimiter: ':' }),
> ```

However, when I try to use the colon as a separator I get an error: `Error: Queue name cannot contain :`, any idea why?

Using a fullstop as a separator works fine, but not the colon.

Issue #692, opened 2024-02-23
Feature: Pause/Resume all queues at once


happyhunter7:
I think it would be great to have a button somewhere on the main page which will allow you to pause all Queues at once and not go through all manually and pause them

I can see this as a useful thing when someone wants to deploy something new but before needs to wait till all jobs are finished and the pause-all will stop future jobs from being processed till the deployment is done

PR #868, opened 2025-01-02
F/692 pause resume jobs at once

ahnwarez:
This PR addresses a requested feature in issue #692 so that users can pause or resume all queues at once with a single click. I'm one of the users who is keen to have this feature as it makes easier to control all queues.

A new component is added to the `OverviewPage.tsx` that displays a drop-down list.

Screenshot 2025-01-03 at 9 09 03 AM

The frontend component calls an api to the backend to loop over every queue to pause or resume.
- `/api/queues/pause`
- `/api/queue/resume`


felixmosh:
I didn't forgot about this PR, I'm a bit busy, I'll review it ASAP

felixmosh:
Thank you for this PR. 
I'll refactor this dropdown, to show only the relevant option.

fred26s:
Is this button not being controlled by readonly: true? Even with readonly enabled, this button is still visible.

felixmosh:
@fred26s yeap, this is a bug 😬

felixmosh:
@fred26s fixed in `v6.7.8`

fred26s:
> @felixmosh  fixed in `v6.7.8`

hi, I'm encountering a server error in v6.7.8 that seems to be related to packages/api/src/handlers/pauseAll.ts. 

The error is: TypeError: req.queues.values(...).filter is not a function. Could there be a potential issue with this file?

felixmosh:
Yeap, you are right :|

Fixed in v6.7.9

kfir-novella:
Was this feature removed? 
Oh, there is a bug?

As after "pause all", I don't see the option to resume all. It still shows me "pause all". 
image

fyi @felixmosh @fred26s

felixmosh:
@kfir-novella What version do you use?
Screenshot 2026-07-20 at 14-01-25 —

kfir-novella:
My bad, seems it exists..
The issue is probably it is very slow until all stopped

Or if all stopped but 1, i ccant so resume all.
I had to resume 1 by 1 (about 100 queues)

I did a query that does it in parallil for next time and test it, seem works very fast, wondering if doing some queries here in parallil will improve speed as well.


Thanks anyway

PR #877, opened 2025-01-18
fix: made 'Data' as default tab in error section

DarkNinja15:


DarkNinja15:
@felixmosh Please review it. This feature is needed by me as it is getting harder for me to look through data in the jobs in error section

felixmosh:
Hi, thank you for the PR... The error tab is the default tab when the job is fails.
This is by design :)

DarkNinja15:
@felixmosh Hey, but I really needed this. In the error section one always needs the job id or some metadata like for which user the job has failed then we switch to the error tab to see what was the error was the job that failed

felixmosh:
You can set the default tab from settings, no?

DarkNinja15:
It doesn't work. By default it is set to Data only but whenever I go to the failed section it switches to error tab

Mitrajit:
@felixmosh It seems the default tab setting isn't functioning as expected. Could you please take a look and address the issue? Your help would be greatly appreciated. Thank you!

felixmosh:
Ok, I'll check it, thank you 🙏🏼

PR #886, opened 2025-01-30
fix: reverse array to display recent errors

### LINKED ISSUE #862: BUG: failed jobs showing stale information in "Error" tab after manual retry
When I manually retry failed jobs in my and they fail again, the information shown in the "Error" is that of the first retry, not that of the latest.

This can be quite frustrating as not all jobs have proper logs, and it leaves me with no choice other than checking my app logs which is time consuming.

Error information should probably be that of the latest retry, or maybe an aggregation of all tries.

ahnwarez:
Closes #862 

A user reported that when they retry failed jobs they want to be able to see the latest error at the top and this PR fixes this problem so that users don't need to scroll down to find the last error for a failed job.
 
The only change made is in the `api` and that is reversing the order of the `stacktrace` array.


felixmosh:
Thank you 🙏🏽

PR #888, opened 2025-02-04
docs: QueueMQ object proper Instantiation

rrrokhtar:
Replaced new QueueMQ() with new QueueMQ.Queue() to be up to date with the latest bullmq


felixmosh:
Hi, thank you for the fix

Can you point out the change in the lib?

PR #901, opened 2025-02-23
Now able to interact with specific jobs with id's containing special chars

rodhoward:
Recurring jobs were causing a 404 error when individually promoted or cancelled or retried etc.


felixmosh:
Thank you for this PR 🙏🏼

PR #907, opened 2025-03-04
feat: add Redis client support and search functionality to queue adapters

Bmg09:


felixmosh:
Hi @Bmg09 thank you for this PR.
Unfortunately, I won't aprove it, since it exactly what I don't want to maintain.

I don't want Redis poking arround code... since it is using / exposing underline lib implementation.

Issue #913, opened 2025-03-21
ReferenceError: Can't find variable: require when building Bun project due to dynamic import with eval


olishevskii:
I'm encountering the following error when attempting to build a project using Bun:

<img width="1128" alt="Image" src="https://github.com/user-attachments/assets/e6817b7e-5a47-4f98-8321-8d2370429cd7" />

This issue seems to occur when I attempt to dynamically import modules via eval. It works fine during development, but breaks when attempting to build the project.

Bun issue: https://github.com/oven-sh/bun/issues/5809#issuecomment-2065310008

felixmosh:
Can you try to pass `options.uiBasePath`?

olishevskii:
@felixmosh 
I had to do this in production because the project wouldn’t build. However, to make this work, I had to read the project’s source code. And judging by the fact that others have faced the same [issue](https://github.com/oven-sh/bun/issues/5809#issuecomment-2065310008), it’s not obvious. I believe this problem should be solved in a more general way.

olishevskii:
@felixmosh 
If you still believe that this solution is more optimal, then let’s add a note somewhere that in order to build with Elysia (bun), this parameter needs to be added. The example also needs to be updated accordingly.

felixmosh:
Will you be able to make a PR that changes the Elysia example, to add this config?
Do you think it is a must, without it, it won't work?

olishevskii:
@felixmosh 
Yes, I can. The issue is that Bun cannot use require in eval, which is the reason for this.

olishevskii:
@felixmosh Can't we use asynchronous imports with ES modules to dynamically get the path without using eval?

felixmosh:
It won't work with bundles.
For bun, you can just pass the path yourself (via `options.uiBasePath`), and the eval thing won't be needed...

olishevskii:
@felixmosh 
Alright. Should I update the Elysia example for this case?

felixmosh:
Yes please 🙏🏽

olishevskii:
@felixmosh PR: https://github.com/felixmosh/bull-board/pull/915

Issue #310, opened 2021-08-05
Production Build Issue with Webpack


lukepolo:
I created a ticket in webpack and basically said to fix it id have to re-write part of that part that reads from the path

https://github.com/webpack/webpack/issues/13931

I can try to tackle this in a few days

felixmosh:
Thank you for reporting the issue, can you elaborate what is the issue?

I'm working with `bull-board v3.4.0` in production and it works perfectly.

lukepolo:
Yah when running in production mode for some reason with this webpack config i get that error

https://github.com/lukepolo/webpack-resolve-bug/tree/main/dist  is the reproduction

felixmosh:
Thanx, i will check it out

felixmosh:
Fix released in `v3.4.1`

lukepolo:
your amazing. thanks!

PR #914, opened 2025-03-21
Fix building in Elysia(bun)

olishevskii:
Issue: https://github.com/felixmosh/bull-board/issues/913

This pull request addresses an issue where the Bun build process fails with the following error:

Screenshot 2025-03-21 at 21 46 13

The error occurs when dynamic imports are used with eval. The issue was observed when attempting to build the project, but it works fine during development.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,5 @@
     "@types/react": "17.0.63",
     "@types/react-dom": "17.0.20"
   },
-  "dependencies": {},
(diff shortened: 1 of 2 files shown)


olishevskii:
Is it possible to use a static path, or is a dynamic path required here? I would also like to understand the reason behind it. If using a static path isn’t a good option, we can try doing it via import, but in that case, we'll need to modify the TypeScript configuration.

felixmosh:
Hi, this code uses eval in order to support bundles.... actually the origianl code was as you've suggests, 
https://github.com/felixmosh/bull-board/issues/310

I don't want to go back to it... this is the reason that I added the `options.uiBasePath`, so unique envs will be able to provide it.

PR #919, opened 2025-04-02
feat: Option to show queue stats in menu tree

mattpardee:
This implements an option to show queue stats in the sidebar so one can see at a glance the activity across all their queues.

Notes:

* I attempted to implement tooltips so when hovering over the number it would show but it seems the Tooltip implementation doesn't work with overflow containers
* Statuses with 0-counts are filtered out of rendered array

In the screenshots below, I used just a single example number to show a worst-ish case scenario, where the user has to scroll over if every queue has jobs in it. The "NewRegistration" queue shows what it looks like when scrolled all the way to the right:

Screenshot 2025-04-02 at 9 45 49 AM

Screenshot 2025-04-02 at 9 45 55 AM

Love bull-board, thank you for it!


mattpardee:
Hello! That's a fair point.

For us, we are typically in one queue and want to see what's happening in other queues without having to jump back to the dashboard.

However it's not a blocking issue, as there is a workaround by going to the dash

PR #931, opened 2025-04-07
Add job action buttons to "prioritized" status

### LINKED ISSUE #930: No action buttons on Prioritized jobs
When looking at a job under Waiting, Waiting Children, or similar, you have buttons to perform common actions on the job, like duplicating, editing, and deleting the job.
These buttons disappear when the job has a priority set and it appears under the Prioritized list.

**Waiting:**
Image

**Prioritized:**
Image

The jobs in this example are identical, except one of them has a `priority: 20` option while the other did not. No queue workers are running in this example, so this is not an effect of something being worked on.

Mange:
This status is similar to `waiting`, only that the job also has a priority set.

Fixes #930


felixmosh:
You are FAST, 
Thank you 🙏🏽

PR #933, opened 2025-04-08
Add autodiscover queues option for BullBoard

DamienBoue:
Adding an autodiscover option in bull board to let bull bord discover all the queues with workers that are defined by bull.

When creating the dashboard, fill the autodiscover option with the connection and the prefix to use to discover all the queues

The example.ts has been updated to help testing it

Make sure that a Bull Engine is running to have workers active on the queues. Otherwise it won't show up in the dashboard


felixmosh:
Hi @DamienBoue 
Thank u for this pr, but it violates one of my rules, I don't add any code that pokes around in the underline storage (redis).
This simple rule, help me maintain this package health.

From my point of view, pokeing in Redis is impelmentation details, that this lib should not aware of.
Thank u for your time.

PR #947, opened 2025-04-30
fix: use node:fs instead of `Bun.fs` in `ElysiaAdapter`

boredland:
this way we can use the same code for bun and node, as elysia is compatible to both.

(experimental) fs.glob/fs.promises.glob has been [added](https://nodejs.org/docs/latest-v22.x/api/fs.html#fspromisesglobpattern-options) in node >=22

Sorry for all the formatting. If it is too annoying, I can revert that.

Added a little fix for the mime-types for the static files being "" in my setup, which I think others could benefit from too.


felixmosh:
Hi,
Thank you for your contribution, can you remove the styling changes?

boredland:
> Hi, Thank you for your contribution, can you remove the styling changes?

Done!

felixmosh (approved):

felixmosh:
Thank you for your contribution.

Released in `v6.9.4`

Issue #831, opened 2024-10-10
Customize the "job id" click action (or add a separate button)


vidjuheffex:
Searched through the issues, both open and new, and didn't see this:

when I'm on a screen listing jobs and can see job `a7326b01-4a83-40e3-bb15-d1b8d5a8f12e` for example, when I click that, it takes me to a singular view of that job.

It would be cool if they had some thing like a 🌐 link next to them to take you to the "web view of that job" ie.. a place in your app where that job might live.

In our app that is `/job/{jobId}`

I'm sure many apps have a place they would like to send you from the diagnostic view so I figured I'd open this feature request.

felixmosh:
How would you pass this "base path" to the board?

felixmosh:
Will global (for all queus) url template will do the job?
`/my-app/{{job.id}}` it can be an absolute url (to other domain) as well.

jaroslaw-weber:
how about we could `url` to job options? this would be easy to implement and more customizable

felixmosh:
I'm not sure I get the resent question

jaroslaw-weber:
currently we have "description" on the queue that is showing on top of each queue.

what if we could also pass a callback in same way like this:
getUrl: jobId => url

felixmosh:
Function can't be serialized (config is done on serverside, the ui is client side...)

jaroslaw-weber:
how about we use templates?

https://yourserver.com/asset/preview/${jobId}

then it would be possible to serialize

vidjuheffex:
I'm so sorry I did not see this thread popping off.

I envisoned this like jaroslaw-weber is writing it.

felixmosh:
Yeah, a url template for each queue may work.

AidanHibbard:
@vidjuheffex @felixmosh 

I opened https://github.com/felixmosh/bull-board/pull/948 to implement this, very open to feedback, and have time to refactor if any changes are needed.

PR #948, opened 2025-05-03
feat: Add job URL template

AidanHibbard:
API - Adds an optional external job URL template for linking to job pages in BaseAdapter

```ts
type JobUrl = {
  displayText?: string
  href: ((job: QueueJobJson) => string)
}
```

UI - Adds an external link to the route provided by the template (see screenshot)

image


felixmosh:
Hi @AidanHibbard, thank you for your contribution.

I prefer that there will be an icon of external link near the job id.
This one https://fontawesome.com/icons/up-right-from-square?f=classic&s=solid WDYT?

AidanHibbard:
@felixmosh

That makes sense to me, I'll get that updated. Thanks for the feedback!

felixmosh:
BTW, this field can be a function, that gets the job, and returns a string.
This way, you are not restricting only to one job field

jaroslaw-weber:
also "text" for the url would be nice

AidanHibbard:
@jaroslaw-weber

Can you help me understand what you're after with these?
- can we get multiple urls?
- also "text" for the url would be nice

For your second question it sounds like @felixmosh is after a single icon?

But for multiple URLs would you need multiple links for a single job?

felixmosh:
@jaroslaw-weber what is the use-case for multiple urls for one job?

jaroslaw-weber:
you could have multiple links, for example:
- go to user dashboard
- see in internal tool
etc

it would be bit more flexible. sometimes depending on the issue i want to be able to have options to go to different places. im not at work right now but would be good to be able to customize, not just have one icon.

maybe interface like:

interface JobExternalUrl {
url: string
title?: string
icon?: string
}

then can pass with externalUrls: JobExternalUrl[]

but anyway its not super important, even one url will be helpful

jaroslaw-weber:
i think best to just start with single url, we can circle back to it after i find a good use case

AidanHibbard:
@felixmosh @jaroslaw-weber Changed up the DX a bit to support a FN for creating the URL, and using a displayText if passed.

Let me know what you think.

felixmosh:
Thank you, I want to change it a bit.

released in `v6.9.3`

felixmosh:
closes #948 https://github.com/felixmosh/bull-board/pull/948

felixmosh:
closes #831

Issue #949, opened 2025-05-03
bug:  [Elysia + Bun] The "options" argument must be of type object. Received undefined


thereis:
Currently, I'm facing this issue when running with Bun + Elysia:

```shell
server:dev: 23 | function validatePattern(pattern) {
server:dev: 24 |   if (@isArray(pattern))
server:dev: 25 |     @throwTypeError("fs.glob does not support arrays of patterns yet. Please open an issue on GitHub.");
server:dev: 26 |   return validateString(pattern, "pattern"), pattern.replaceAll("/", "\\");
server:dev: 27 | }
server:dev: 28 |   validateObject(options, "options");
server:dev:                      ^
server:dev: TypeError: The "options" argument must be of type object. Received undefined
server:dev:  code: "ERR_INVALID_ARG_TYPE"
server:dev: 
server:dev:       at mapOptions (internal:fs/glob:28:17)
server:dev:       at glob (internal:fs/glob:4:33)
server:dev:       at registerPlugin (C:\Projetos\sygnaldev\node_modules\.pnpm\@bull-board+elysia@6.9.5_el_31dfcf97028ed754dfb26847343f2e72\node_modules\@bull-board\elysia\dist\ElysiaAdapter.js:92:26)
server:dev:       at registerPlugin (C:\Projetos\sygnaldev\node_modules\.pnpm\@bull-board+elysia@6.9.5_el_31dfcf97028ed754dfb26847343f2e72\node_modules\@bull-board\elysia\dist\ElysiaAdapter.js:64:28)
server:dev:       at <anonymous> (C:\Projetos\sygnaldev\apps\server\src\http-instance.ts:98:10)
server:dev:       at createHttpInstance (C:\Projetos\sygnaldev\apps\server\src\http-instance.ts:30:35)
server:dev:       at <anonymous> (C:\Projetos\sygnaldev\apps\server\src\index.ts:11:11)
server:dev:       at main (C:\Projetos\sygnaldev\apps\server\src\index.ts:7:14)
server:dev:       at C:\Projetos\sygnaldev\apps\server\src\index.ts:29:1
```

Code:

```ts
import { BullMQAdapter } from '@bull-board/api/bullMQAdapter';
import { ElysiaAdapter } from '@bull-board/elysia';
import { createBullBoard } from '@bull-board/api';

const exampleBullMq = createBullQueue('device-setup');

const serverAdapter = new ElysiaAdapter('/ui');

createBullBoard({
[...]

PR #950, opened 2025-05-06
fix: use glob callback instead of promise

boredland:
Apparently there are issues with the node:fs compat layer in bun.

According to the bun docs, it _should_ work. Apparently it doesn't. Somehow the callback version works, tho, and I don't think there's much of a difference except the verbosity, so let's use it instead. Tested with the example in this repo.

Closes https://github.com/felixmosh/bull-board/issues/949


felixmosh:
Thank you 🙏🏼 
Will publish it asap

felixmosh:
Released in `v6.9.6`

PR #965, opened 2025-06-09
Respect the `basePath` argument

zackdotcomputer:
This makes four changes to the `@bullboard/fastify` package:
1. Run `yarn` to update the lockfile to the latest versions
2. Refactor the `as any` casts to be more precise casts (so that future fastify changes that might cause bugs will be detected as a mismatch by the compiler)
3. Default the `opts.basePath` value to the `this.basePath` value for the adapter
4. Most importantly: use the `basePath` as a prefix on the templates, static content, handlers, and apiRoutes

This means that now if you set `setBasePath` before calling `registerPlugin`, the bullboard UI and other routes will actually register beneath that base path.

### DIFF
--- a/packages/fastify/src/FastifyAdapter.ts
+++ b/packages/fastify/src/FastifyAdapter.ts
@@ -1,4 +1,4 @@
-import {
+import type {
   AppControllerRoute,
   AppViewRoute,
   BullBoardQueues,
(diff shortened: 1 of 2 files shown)


felixmosh:
Hi @zackdotcomputer thank you for this PR.

What is the problem that you are trying to solve?
The fastify example works properly, with prefix.

```
app.register(serverAdapter.registerPlugin(), { prefix: '/ui' }); 
// -----------------------------------------------^ this prefixes all routes of the plugin without the need to pass it down
```

basePath has diffrent usage, it defines the `<base />` tag.

It can be somtimes different.

PR #974, opened 2025-07-14
Support custom Prefix for bullMQ

sivakumar07:
Support custom Prefix for bullMQ


felixmosh:
This is not true, since bull-board does touches Redis nor the queue instance... It get configured queue instance, so you can define it as you wish.

PR #976, opened 2025-07-15
Fix job clean issue in Elysia

d4n5h:
There's an issue with cleaning/deleting a job via BullBoard+Elysia.
This PR patches it.

### DIFF
--- a/packages/elysia/src/ElysiaAdapter.ts
+++ b/packages/elysia/src/ElysiaAdapter.ts
@@ -32,6 +32,28 @@ export class ElysiaAdapter implements IServerAdapter {
       // @ts-ignore
       this.plugin.config.prefix = this.basePath;
     }
+
+    this.plugin.onParse(async ({ request }, contentType) => {
+      // Handles `Content-Type: application/json` with an empty body.
+      if (contentType === 'application/json') {
+        const text = await request.text();
+        if (text.length === 0) {
+          return {};
+        }
+        return JSON.parse(text);
+      }
+
+      // Handles PUT/POST/PATCH with `Content-Length: 0` and no `Content-Type`.
+      // Bull Board UI sends requests like this for some actions (e.g. clean job).
+      if (!contentType) {
+        if (request.method === 'PUT' || request.method === 'POST' || request.method === 'PATCH') {
+          const contentLength = request.headers.get('content-length');
+          if (contentLength === '0') {
+            return {};
+          }
+        }
+      }
+    });
   }
 
   public setStaticPath(staticsRoute: string, staticsPath: string): ElysiaAdapter {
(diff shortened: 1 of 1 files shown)


packages/elysia/src/ElysiaAdapter.ts
```diff
@@ -136,7 +158,7 @@ export class ElysiaAdapter implements IServerAdapter {
       const relativePath = path.substring(path.indexOf('dist') + 4).replaceAll('\\', '/');
       this.plugin.get(relativePath, async () => {
         const fileContent = await fsPromises.readFile(path);
-        return new Response(fileContent, {
+        return new Response(fileContent.toString(), {
```
felixmosh, line 161: How does this related to the "onParse" part?

felixmosh: Why your issues are not reproduces in the `with-elysia` example?

felixmosh: Yeah, but there was a PR that made useage of this Elysia adapter sutable in Node as well...

d4n5h:
> Yeah, you are right, I see that the request is with `content-length:0`, but still it not fails for me... so what is the differance? Can you clone the example and try it?
> 
> I think that it is better to send an empty object (it will be more future proof.) But I need to reproduce it.

I'm not sure why it doesn't fail for you, but with my patch it takes care of all the edge cases, and you can still send an empty body, and it streams the files instead of loading them into memory and then serving them. So there are a few improvements.

Anyway, I modified the PR to support both NodeJS and BunJS.

d4n5h:
> Yeah, I've misconfused it with the `toWeb()`
> 
> ```js
> import { Readable } from 'node:stream';
> 
> // 1. Create a Node.js fs.ReadStream
> const nodeReadStream = fs.createReadStream('./my-large-file.txt');
> 
> // 2. Convert it to a web ReadableStream
> const webReadableStream = Readable.toWeb(nodeReadStream);
> ```

Ideally, you're right.
But in this case, it also has a type error:
image

felixmosh:
Weird, anyway, make a separate PR for the stream thing...
Regarding `content-length:0` I must reproduce this issue first, in order to solve it.
I do understand that it works on your side, but maybe you have something else that changes the request (I've encountered with this already.)

Therefore, before accepting this issue. I would like to understand why is it works for me, but not for you. (official example)

d4n5h:
> can you pull the example repo, and check if it happens there?

The versions in the examples are also outdated, but after I updated the versions it still works.
So it's probably something that's related to my environment, I'll look into it later and figure out what's going on.

Anyway, I've removed the onParse code and only kept the file streaming patch.
I'll change the PR name and then you can pull it.

PR #982, opened 2025-07-21
feat: add danish translations

simplenotezy:
Adds support for danish.

* based on https://github.com/felixmosh/bull-board/pull/981 for simplicity


felixmosh:
Thank you for your contribution 🙏🏽

Issue #908, opened 2025-03-05
bug: Removing repeatable jobs added via job scheduler fails.


perezpaya:
Removing repeating jobs added via jobScheduler fails with a 500 error.

Also trying to do the same via clean all fails too.

`PUT /admin/queues/api/queues/default/repeat%3AUpdateTokensFromCoinGeckoJob%3A1741233600000/clean 500 40ms`

<img width="1167" alt="Image" src="https://github.com/user-attachments/assets/35407e57-c7fe-412d-9e1e-185701019938" />

We are running this versions:
```
"@bull-board/api": "^6.7.10"
"@bull-board/express": "^6.7.10"
"bullmq": "^5.41.7"
```

felixmosh:
Hi, thank you for reporting this issue.
Will you be able to create a reproduction repo?

willr-republic:
Got the same problem and did some investigation.
I'm not able to reproduct the issue when running bull-board by cloning this repo. Error occurs when I run the npm plublished package `6.7.10`.
The error returned by the server is:

```json
{
  "error":"Internal server error",
  "message":"Job repeat:repeat-TestJob:1741946400000 belongs to a job scheduler and cannot be removed directly. removeJob"
}
```

anudeep4n:
@willr-republic downgrading bullmq from  "^5.34.4" to "4.18.2" when i downgrade I am not able to do some operations on the queue like upsertJobScheduler


@felixmosh

PaddaSecurity:
Getting same error

PaddaSecurity:
> Got the same problem and did some investigation. I'm not able to reproduct the issue when running bull-board by cloning this repo. Error occurs when I run the npm plublished package `6.7.10`. The error returned by the server is:
> 
> {
>   "error":"Internal server error",
>   "message":"Job repeat:repeat-TestJob:1741946400000 belongs to a job scheduler and cannot be removed directly. removeJob"
> }
> Downgrading `bullmq` from `5.43.0` to `4.18.2` fixes the issue.

This fixed it for me as well

anudeep4n:
> [@anudeep4n](https://github.com/anudeep4n) currently there is no support for job scheduler in Bull-Board...

when can we expect this support

tarcisioandrade:
same error here

mbsanchez01:
I'm facing this issue too, using 
`"@bull-board/api": "^6.10.1",
  "@bull-board/express": "^6.10.1",
[...]

PR #856, opened 2024-11-27
Adding jobScheduler support for BullMQ queues

lpoulin:
Putting this work on hold for now as they continue to expand the support of jobSchedulers in BullMQ (check the last few weeks of commits). A couple of notes:

- I know this is very messy in spots (didn't add all translations, etc.), but I wanted to go ahead and open this up because I know others are waiting on it and I don't have the bandwidth at the moment
- We may want to reconsider where to navigate to the job schedulers page from, instead of from the job status bar like I currently have it. On smaller screens, the "Schedulers" status is already off of the page to the left. 
- I have (obviously lol) not completed the UI portion for the scheduler card. I put that on hold as I was waiting to see if they would expand support for the Job Scheduler API - it is pretty barebones at the moment.

PR #1003, opened 2025-09-03
support clean each jobScheduler job

for2gles:
It's related with #908.
It doesn't help to clean with `Clean all` Button.
But it will allow to delete specific jobScheduler job


felixmosh:
Thank you 🙏🏼

felixmosh:
released in `v6.12.8`

for2gles:
You right,
Maybe [#856](https://github.com/felixmosh/bull-board/pull/856) is already kinda is it?

PR #1008, opened 2025-09-19
feat(ui): add collapsible menu state with persistence feature

melalj:
- Introduced `useMenuState` hook to manage the open/closed state of menu items with localStorage persistence.
- Updated `MenuTree` component to utilize menu state, allowing users to toggle and persist their menu preferences across sessions.
- Each menu item's state is stored and retrieved using its path as a key, enhancing user experience by maintaining expanded menu sections between visits.


felixmosh:
Thank you for this PR!
It looks awesome, I'll modify it a bit since we are already using zustand with localStorage persistance, so it is better to use it

felixmosh:
I've tested your code, there was an issue with your implementation, onToggle is triggered whenever open field is changed.

Fix it by using `onClick` on summery & refactor the hook to use zustand.

Thank you!
Released in `v6.13.0`

PR #1020, opened 2025-10-23
feat(job-cleaning): add support for cleaning scheduled jobs

benatshippabo:
Allows users to clean up scheduled jobs. Taking over the effort from #1003 but properly obtains the `repeatJobKey` but this time properly uses the right queue. Added an integration test, to ensure it properly works. Also tested cleaning up some scheduled jobs through the bullboard ui and it works as well.


learningbizz:
Hey @benatshippabo, I'm in the same situation as meshuamam, you don't have any issue deleting any at the moment?

PR #1025, opened 2025-11-08
handle collapsible json for data and options

### LINKED ISSUE #909: Collapsible json for data
Allow the json (and its properties) data to be collapsed. Currently the scroll for the data section is nice, but being able to collapse one or two properties make them more navigable.

Being able to set a setting for defaulting object properties to be collapsed below the first set of children would be great also. My preference would be for `jobData` and its children properties could be visible, then any nested objects/arrays would be collapsed by default.

- Mentioned in #264

scottrepreneur:
- Closes #909


felixmosh (commented):
This looks good, but why did you decided to re-implement it, why not using some lib?

scottrepreneur:
You hadn't responded to the suggestion in the issue for library. Is [`react-json-view`](https://github.com/mac-s-g/react-json-view) preferred?

Issue #1031, opened 2025-11-20
When clicking 'Duplicate Job' the data should not be the output of `queueAdapter.setFormatter('data', ...)`


irian-codes:
Hi,

We're using `superjson` in BullMQ and since Bullboard is not compatible with it, we save the superjson stringified keys in string format. So when `JSON.parse` occurs inside Bullboard it displays an non human readable string like `"{\"json\":{\"date\":\"1970-01-01T00:00:00.000Z\"}, \"meta\": {\"values\": {\"date\": \"Date\"}}}"`.

So our solution si to create a data formatter with `queueAdapter.setFormatter('data', dataFormatter)` and now we can see only the `json` property in a nicely formatted way.

The problem is that when clicking on 'Duplicate job', the formatter runs too and generates a non `superjson` compatible JSON. So, the solution would be for Bullboard to display the raw data saved into Redis instead of the formatted data.

Job retries work with no issues, but not duplicating/adding a job manually.

irian-codes:
I'm finding difficult to prioritise this issue in the end. It's something we want to do but we're not using the feature as much as we thought so its still low priority for us. Feel free to mark this as stale as I cannot guarantee I'll finish the PR.

If its something you think adds value I don't have any issue in someone else finishing it, otherwise I don't know when I will be able to address this, sorry.

PR #1035, opened 2025-11-20
Return unformatted data on job

irian-codes:
Fixes https://github.com/felixmosh/bull-board/issues/1031

Hi! Thanks for making Bullboard, it's very useful!

Since we're using it at my company, we have the need to see the data without formatting. So I made this PR 😊

What it does it allows the user to get a copy of the original unformatted data when duplicating a job or adding a job. So it can be compatible with BullMQ original data without the formatting changes applied on top.

In the process, I also found what I believe was a bug in the format function (explained [here](https://github.com/felixmosh/bull-board/pull/1035/commits/17f001adc3be189c3bcfba0d137eea63a7290f77)), that I fixed.

I hope you can take a look at it, and I'm here if you need any changes. I tried to keep it as simple as possible and backwards compatible. That's why:
- I added a dashboard setting at `false` by default, to maintain backwards compatibility.
- I only touched the parts that affected manipulating data in modals, even though it could be extended when displaying data too in the Job Card.

I also translated the new i18n key but I'm not native in all these languages so sorry if there are mistakes. I am however proficient in English and Spanish so those too should be correct.

### DIFF
--- a/packages/api/src/queueAdapters/base.ts
+++ b/packages/api/src/queueAdapters/base.ts
@@ -56,7 +56,9 @@ export abstract class BaseAdapter {
 
   public format(field: FormatterField, data: any, defaultValue = data): any {
     const fieldFormatter = this.formatters.get(field);
-    return typeof fieldFormatter === 'function' ? fieldFormatter(data) : defaultValue;
+    return typeof fieldFormatter === 'function'
+      ? fieldFormatter(structuredClone(data))
+      : defaultValue;
   }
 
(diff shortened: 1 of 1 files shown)


felixmosh:
Hi, thank you for the PR... 
I've thought about this issue, and I think that it is better to "fetch" only on duplicate / edit clicked... so instead of returning the rawData for each job every time (formatter for data was a solution for performance, since there are jobs with alot of data), before  / edit clicks, fetch clean job, and just use it instead of what comes from the Queue.

In my opinion this will solve all these issues... (no need for settings for it)

PR #1050, opened 2025-12-25
refactor(bun): implement routes object paradigm for Bun adapter

hanzalagithub:
## Summary
This PR addresses the feedback from #[original PR number] and reimplements the Bun adapter to follow the proper Bun server paradigm.

## Changes Made
- ✅ Changed `BunAdapter` to return a routes object instead of a single monolithic fetch handler
- ✅ Renamed primary method to `getRoutes()` (as suggested by maintainer)
- ✅ Routes object structure: `{ [path]: { [METHOD]: handler } }`
- ✅ Updated example to use `Bun.serve({ routes: { ...serverAdapter.getRoutes() } })`
- ✅ Each route path is registered individually, letting Bun's router handle URL matching

## Implementation Details
Following the maintainer's feedback, the adapter now:
1. **Returns composable routes object** - Can be spread into `Bun.serve()`
2. **Builds routes once** - Inside `getRoutes()` method
3. **Follows FastifyAdapter pattern** - Similar structure for route registration
4. **Lets Bun handle routing** - No monolithic fetch handler

### Example Usage
```typescript
const serverAdapter = new BunAdapter();

createBullBoard({
  serverAdapter,
});

const bullBoardRoutes = serverAdapter.getRoutes();

Bun.serve({
  port: 3000,
  routes: {
    "/health": {
      GET: () => Response.json({ status: 'ok' })
    },
    ...bullBoardRoutes,  // ✅ Spread operator works!
  }
});

### DIFF
--- a/packages/bun/README.md
+++ b/packages/bun/README.md
@@ -1,0 +1,23 @@
+# <img alt="@bull-board" src="https://raw.githubusercontent.com/felixmosh/bull-board/master/packages/ui/src/static/images/logo.svg" width="35px" /> @bull-board/bun
+
+[Bun](https://bun.sh/) server adapter for `bull-board`.
+
+<p align="center">
+  <a href="https://www.npmjs.com/package/@bull-board/bun">
+    <img alt="npm version" src="https://img.shields.io/npm/v/@bull-board/bun">
+  </a>
+  <a href="https://www.npmjs.com/package/bull-board">
+    <img alt="npm downloads" src="https://img.shields.io/npm/dw/bull-board">
+  </a>
+  <a href="https://github.com/vcapretz/bull-board/blob/master/LICENSE">
+    <img alt="licence" src="https://img.shields.io/github/license/vcapretz/bull-board">
+  </a>
+<p>
+
+![Overview](https://raw.githubusercontent.com/felixmosh/bull-board/master/screenshots/overview.png)
+![UI](https://raw.githubusercontent.com/felixmosh/bull-board/master/screenshots/dashboard.png)
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/bun/src/BunAdapter.ts [hunk 1]
```diff
@@ -0,0 +1,283 @@
...
+    const routes: BunRoutes = {};
+
+    // Register static file routes with wildcard
+    const staticBasePath = this.joinPaths(this.basePath, this.staticRoute);
+    routes[`${staticBasePath}/*`] = {
+      GET: async (request: Request) => {
+        const url = new URL(request.url);
+        const pathname = url.pathname;
+        const relativePath = pathname.replace(staticBasePath, '');
+        const fullPath = `${this.staticPath}${relativePath}`;
```
felixmosh: This should be normilized to check if the path is still inside `this.staticPath`, so if an attacker crafts `/static/js/../../../.env`, it won't work

packages/bun/src/BunAdapter.ts
```diff
@@ -0,0 +1,283 @@
...
+              }
+
+              // Build query parameters
+              const query: Record<string, any> = {};
+              for (const [key, value] of url.searchParams.entries()) {
+                query[key] = value;
+              }
+
+              // Build headers
+              const headers: Record<string, string> = {};
```
felixmosh: Same as query

felixmosh (changes requested):

hanzalagithub:
Thank you for allowing me to contribute. I hope my input will be helpful to many people.

PR #1052, opened 2025-12-30
feat(api): add BullMQProAdapter with Groups support

### LINKED ISSUE #874: Support for BullPro Groups
I am encountering an unusual behavior when attempting to use BullMQPro with bull-board. When I have a paused queue and add a job with a group, the job does not appear in the Paused list (I believe this is due to groups having 'virtual queues'). However, if I unpause the queue, the job begins processing in the Active tab. Is there a way to resolve this issue with groups in bullMQPro? Specifically, is it possible to view group jobs that are in Waiting or Paused states?

perezpefaur:
This PR adds a new BullMQProAdapter that extends BullMQ functionality
to support BullMQ Pro's Groups feature.

Groups in BullMQ Pro create 'virtual queues' that allow rate limiting
and concurrency control on a per-group basis. Previously, jobs added
to groups were not visible in the Paused/Waiting/Delayed tabs because
groups store jobs separately from the main queue.

Key features:
- Displays jobs from groups in Waiting, Delayed, and Paused tabs
- Includes group job counts in the total queue counts
- Uses caching (10s TTL) to minimize Redis calls for group operations
- Efficient pagination through groups using getGroupsByStatus API
- Adds group identifier to job names for visibility in the UI

Usage:
  import { BullMQProAdapter } from '@bull-board/api/bullMQProAdapter';
  import { QueuePro } from '@taskforcesh/bullmq-pro';

  const myQueue = new QueuePro('my-queue', { connection });
  const adapter = new BullMQProAdapter(myQueue);

Closes #874

### DIFF
--- a/packages/api/bullMQProAdapter.d.ts
+++ b/packages/api/bullMQProAdapter.d.ts
@@ -1,0 +1,2 @@
+export { BullMQProAdapter } from './dist/queueAdapters/bullMQPro';
+
--- a/packages/api/bullMQProAdapter.js
+++ b/packages/api/bullMQProAdapter.js
@@ -1,0 +1,2 @@
+module.exports = require('./dist/queueAdapters/bullMQPro');
+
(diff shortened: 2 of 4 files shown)


paulvidal:
Thanks @perezpefaur for this contribution, we love it 🫶 and it works well in prod for us !

PR #872, opened 2025-01-06
F/277 display flow tree

ahnwarez:
Related to #277 and #681 

This PR displays a tree-like graph for any job that was generated by a `FlowProducer` and it allows users to navigate the tree, up or down the tree.

https://github.com/user-attachments/assets/bb5d765e-8de9-4878-b91a-287f31057a4f

The changes made cover `api` and `ui` packages.

The CSS and styling in `JobTree.module.css` are not that great. I admit that my skills aren't great. So, please make the necessary changes to remove any repetition or redundancies.

I'll close #870 because all changes are already included in this PR.

Thanks for the feedback on my previous PR.

PR #1065, opened 2026-01-24
feat: implement job flow tree visualization

IbrahimMohammed47:
https://www.loom.com/share/58b584c5fcdd4094bcfb3d19b54e43d4

Description
- Adds job flow tree visualization for BullMQ flows. Users can now see parent-child job relationships in a tree structure that show whole flow and allows easy navigation to all jobs within a flow.

Changes
- Backend: Added `/api/queues/:queueName/:jobId/flow` endpoint with job hierarchy traversal
- Frontend: New `JobFlow` component with tree visualization, improved `useJobFlow` hook, and instant job fetch on navigation

Features
- Interactive job flow tree with color-coded job states
- Instant data loading when clicking job names
- Graceful fallback for non-BullMQ queues
- Progress bars and job metadata display

Compatibility
- No breaking changes
- Backward compatible
- Non-BullMQ queues safely degrade

Note:
This PR addresses [issue #277](https://github.com/felixmosh/bull-board/issues/277). A previous attempt [PR #872](https://github.com/felixmosh/bull-board/pull/872) was closed due to inactivity. This implementation takes a different approach: it displays the complete job flow tree from any node that's part of a flow, providing full hierarchy visibility rather than just parent-child relationships.

### DIFF
--- a/packages/ui/src/pages/JobPage/JobPage.tsx
+++ b/packages/ui/src/pages/JobPage/JobPage.tsx
@@ -42,7 +42,6 @@ export const JobPage = () => {
     if (activeJobId) {
       actions.getJob();
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [activeJobId]);
 
   actions.pollJob();


packages/api/src/queueAdapters/bullMQ.ts
```diff
@@ -119,6 +119,18 @@ export class BullMQAdapter extends BaseAdapter {
     ];
   }
 
+  public async getFlowRoot(nodeId: string): Promise<JobNode | null> {
+    const client = await this.queue.client;
+    const flowClient = new FlowProducer({ connection: client });
+    const root = await flowClient
+      .getFlow({
```
felixmosh: `getFlow`, on the root job id, returns the entire flow tree, not only the root.

https://docs.bullmq.io/guide/flows/get-flow-tree

packages/api/src/handlers/jobFlow.ts
```diff
@@ -0,0 +1,84 @@
...
+        nodeId: jobId,
+        flowRoot: null,
+        isFlowNode: false,
+      },
+    };
+  }
+
+  let currJob = job as Job;
+  let flowRoot: JobNode | null = null;
+  while (currJob) {
```
felixmosh: Can you explain why there is a need for this `while` loop?

felixmosh (commented):
@IbrahimMohammed47 thank you for this PR, I would like to merge it, as it support usefull for many.

Do you think that flow job should be a "tab" inside a job details?

packages/api/src/handlers/jobFlow.ts
IbrahimMohammed47: It doesn't know it beforehand!
The frontend calls the `/flow` endpoint for the currently viewed job. The API tries to walk up to the root, fetches the full tree (if it exists), and returns an isFlowNode flag, which is true when the tree has children. The UI component only renders the flow visualization when isFlowNode is true; otherwise it renders nothing.

felixmosh (approved):
Thank u for this awsome feature, and fixes :]

PR #1124, opened 2026-03-30
Fixes Elysia example when it runs using bun

TheRaizer:
Solves: #1115 where using the with-elysia example in a bun only environment (i.e. bun install) would show errors on start up.


felixmosh:
Thank you, I've change it to be simpler example.

PR #1143, opened 2026-04-08
feat: add textColor and fontSize options to environment badge

Thoritie:
### Why this PR.
Add textColor and fontSize options to the environment UI config

Our team runs bull-board across multiple environments and wanted a more noticeable visual indicator with per-environment color customization. The existing environment config only supported a background color, so we extended it with textColor and fontSize to give teams full control over how each environment badge appears. 😁

The existing environment badge only supported a background color. This extends it with two optional

**fields:** 
  - textColor: controls the label text color
  - fontSize: controls the badge size (padding scales automatically via em)
                                                                                                       
No breaking changes both fields are optional. 
  

Example ui will look like this

Screenshot 2569-04-08 at 18 03 46

Thank you.

### DIFF
--- a/example.ts
+++ b/example.ts
@@ -169,6 +169,16 @@ const run = async () => {
       }),
     ],
     serverAdapter,
+    options: {
+      uiConfig: {
+        environment: {
+          label: 'Production',
+          color: '#e90e20',
+          textColor: '#fff',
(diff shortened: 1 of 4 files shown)


felixmosh (approved):
Thank you for this PR 🙏🏼

PR #1155, opened 2026-04-14
fix: allow jobFlow handler on read-only queues

### LINKED ISSUE #1154: fix: jobFlow handler returns 405 on read-only queues
### Bug

When navigating to a job detail page (e.g. `/admin/queues/queue/<queueName>/<jobId>`) on a queue registered with `readOnlyMode: true`, the UI continuously surfaces the toast error:

> Method not allowed on read only queue

### Reproduction

```ts
serverAdapter.setQueues([
  new BullMQAdapter(queue, { readOnlyMode: true }),
]);
```

Then open any job detail page for that queue. The "flow" polling request fails every interval and the error banner reappears.

### Root cause

[`packages/api/src/handlers/jobFlow.ts`](https://github.com/felixmosh/bull-board/blob/master/packages/api/src/handlers/jobFlow.ts) registers `jobFlowHandler` without `skipReadOnlyModeCheck`:

```ts
export const jobFlowHandler = queueProvider(jobProvider(getJobFlow));
```

`queueProvider` returns 405 when `queue.readOnlyMode` is true unless the handler opts out. Sibling read-only endpoints already opt out:

- `packages/api/src/handlers/job.ts` — `{ skipReadOnlyModeCheck: true }`
- `packages/api/src/handlers/jobLogs.ts` — `{ skipReadOnlyModeCheck: true }`

`getJobFlow` only calls `getState`, `findFlowRoot`, and `getFlowTree` — no mutating operations — so blocking it on read-only mode is unintentional. The handler was added in #1065 and the `skipReadOnlyModeCheck` opt-out simply wasn't included.

### Fix

```ts
export const jobFlowHandler = queueProvider(jobProvider(getJobFlow), {
  skipReadOnlyModeCheck: true,
});
```

Happy to open a PR.

childish-sambino:
## Summary

- `jobFlowHandler` was registered without `skipReadOnlyModeCheck`, so the polling `GET /queues/:name/:jobId/flow` request returned 405 on queues with `readOnlyMode: true`, surfacing "Method not allowed on read only queue" in the UI every poll interval on the job detail page.
- The flow endpoint only reads state (`getState`, `findFlowRoot`, `getFlowTree`), so it should behave like the other read-only endpoints (`job.ts`, `jobLogs.ts`) that already pass `{ skipReadOnlyModeCheck: true }`.

Closes #1154

## Test plan

- [x] Rebuilt `@bull-board/api` locally and verified the error no longer appears when loading a job detail page on a queue configured with `readOnlyMode: true`.

### DIFF
--- a/packages/api/src/handlers/jobFlow.ts
+++ b/packages/api/src/handlers/jobFlow.ts
@@ -69,4 +69,6 @@ async function getJobFlow(
   };
 }
 
-export const jobFlowHandler = queueProvider(jobProvider(getJobFlow));
+export const jobFlowHandler = queueProvider(jobProvider(getJobFlow), {
+  skipReadOnlyModeCheck: true,
+});


felixmosh:
Thank you for submitting this issue + PR 🙏🏼

Released in `v6.21.3`

PR #1195, opened 2026-05-13
Implement support for BullMQ Pro

kibertoad:
There is already https://github.com/felixmosh/bull-board/pull/1052 (which gets huge kudos for being a trailblazer and solving a very real problem!), but I believe that it has several improvements over the prior PR:

* Extends BullMQAdapter instead of cloning it (saves ~150 lines of drifting code, auto-inherits global concurrency methods);
* Zero compile-time dependency on @taskforcesh/bullmq-pro (so it works in CI without access to the paid library);
* Surface group identity via a dedicated groupId field and a default formatter instead of mutating Job.name;
* Cache is invalidated on writes; #1052 did not;
* Includes tests; #1052 had none.

I am happy to keep maintaining BullMQ Pro support and build more elaborate UX to support more of its features over the time, we use both BullMQ and bullboard very actively in Lokalise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,25 @@ app.listen(3000, () => {
 
 That's it! Now you can access the `/admin/queues` route, and you will be able to monitor everything that is happening in your queues 😁
 
+### BullMQ Pro
+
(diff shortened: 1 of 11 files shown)


felixmosh:
WOW, thank you @kibertoad 🙏🏼
I'll review it asap

felixmosh (approved):
Looks OK to me, thank you
Released in `v7.1.3`

PR #1197, opened 2026-05-13
Fix group nesting

### LINKED ISSUE #1196: Group Header Misaligned after Depth 3
When Create deeply nested group headers in the side bar. 

Service-Dispatch-Events-Complete

Where `-` is the delimiter. 

Everything displays correctly in tree format until depth 4

Service
     Dispatch
          Events
 Complete

-=-=-=-=-=-=-

The Actual Grouping logic works correctly, the drop down arrow hides and reveals the items. 
This looks to be a case where the depth is hard coded in the CSS and even if it wasn't it looks like the class name gets dropped after three, most likely since it doesn't exist in the css module after 3. 

```css
  &.level-1 {
    > li > a,
    > li > .groupHeader {
      padding-left: calc(0.75 * 2rem);
    }
  }

  &.level-2 {
    > li > a,
    > li > .groupHeader {
      padding-left: calc(0.75 * 3rem);
    }
  }

  &.level-3 {
    > li > a,
    > li > .groupHeader {
      padding-left: calc(0.75 * 4rem);
    }
  }
```

mwalkersigma:
- Inlined styles on dropdown to make them programatic 
- Removed hard coded styles 
fixes #1196

### DIFF
--- a/packages/ui/src/components/Menu/MenuTree/MenuTree.tsx
+++ b/packages/ui/src/components/Menu/MenuTree/MenuTree.tsx
@@ -36,7 +36,7 @@ export const MenuTree = ({
   }));
 
   return (
-    <ul className={cn(s.menu, level > 0)} style={{ paddingLeft: level = 0 ? undefined : `calc(0.75 * ${level}rem)`}}>
+    <ul className={cn(s.menu, level > 0 && s.level)} style={{ '--level': level } as React.CSSProperties}>
       {tree.children.map((node) => {
         const isLeafNode = !node.children.length;
         const displayName = node.name;
--- a/packages/ui/src/components/Menu/MenuTree/MenuTree.tsx
+++ b/packages/ui/src/components/Menu/MenuTree/MenuTree.tsx
@@ -36,7 +36,7 @@ export const MenuTree = ({
   }));
 
   return (
-    <ul className={cn(s.menu, level > 0 && s[`level-${level}`])}>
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/Menu/MenuTree/MenuTree.tsx
```diff
@@ -36,7 +36,7 @@ export const MenuTree = ({
   }));
 
   return (
-    <ul className={cn(s.menu, level > 0 && s[`level-${level}`])}>
+    <ul className={cn(s.menu, level > 0)} style={{ paddingLeft: level = 0 ? undefined : `calc(0.75 * ${level}rem)`}}>
```
felixmosh: I prefer to set `--level:1`  css var,  and styles will use the level...

```css
.level {
    > li > a,
    > li > .groupHeader {
      padding-left: calc(1.75rem * var(--level, 0))
     }
  }
```

felixmosh (changes requested):

packages/ui/src/components/Menu/MenuTree/MenuTree.tsx
felixmosh: I've updated my example

mwalkersigma: Would you be okay with something like this ?



`<ul className={cn(s.menu, level > 0 && s.level)} style={{ '--level': level } as React.CSSProperties}>`

felixmosh (approved):

felixmosh:
Thank you for the fix 🙏🏼

PR #1215, opened 2026-06-08
feat(ui): add German (de-DE) locale

MikeGrote:
## What does this PR do?

Adds a German (`de-DE`) locale to the dashboard UI.

## Changes

- Add `packages/ui/src/static/locales/de-DE/messages.json` (full translation, 151 keys — same key set as `en-US`; all `{{interpolation}}` placeholders, plural suffixes and `\@{tables=System.Object[]}(...)` references preserved).
- Register `de-DE` in `packages/ui/src/constants/languages.ts` so it appears in the Settings language selector.
- Map `de-DE` -> `de` for date-fns in `packages/ui/src/services/i18n.ts`.

## Notes

- Key parity with `en-US` verified programmatically (no missing/extra keys).
- Placeholder parity with `en-US` verified programmatically (identical interpolation tokens per string).

### DIFF
--- a/packages/ui/src/constants/languages.ts
+++ b/packages/ui/src/constants/languages.ts
@@ -1,6 +1,7 @@
 export const languages = [
   'en-US',
   'en-GB',
+  'de-DE',
   'es-ES',
   'fr-FR',
   'ja-JP',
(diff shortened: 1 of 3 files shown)


felixmosh (approved):
Thank you for the PR 🙏🏼

PR #1208, opened 2026-06-03
Add per-queue throughput metrics

TomKALBE:
Adds a throughput chart (jobs/min) to each queue page, backed by BullMQ/Bull getMetrics. Builds on the work started in #619 .

New endpoint GET /api/queues/:queueName/metrics → { completed, failed } per-minute series
New QueueMetrics card: recharts area chart (completed vs failed) + hover tooltip + summary (rate, peak, window total)

image

Notes

Series are wall-clock aligned over a 60-min window (meta.prevTS, idle minutes zero-filled).
Bull adapter coerces Redis string data points to numbers; BullMQProAdapter inherits the BullMQ impl.
recharts ^2.15.4 (React 17 compatible; v3 needs React 18).
i18n added for all locales; example.ts enables metrics for the demo.
Test: metrics.spec.ts (endpoint shape + 404).

PR #1216, opened 2026-06-09
docs: document uiConfig.showMetrics option

TomKALBE:
## What

Documents the `uiConfig.showMetrics` board option in the README, alongside the
other `uiConfig` options.

The per-queue throughput metrics feature itself was added in #1208, but the
`showMetrics` flag that toggles it was not documented.

## Details

The new entry covers:
- Default value (`false`) and what enabling it does (per-queue completed/failed
  throughput chart on each queue page).
- That the chart component and its `recharts` dependency are lazy-loaded, so
  they're only fetched when the option is on.
- That it requires worker-side BullMQ/Bull metrics collection to be enabled,
  otherwise the chart shows an empty state.

Also adds `showMetrics: true` to the existing `uiConfig` code example.

## Type of change

- [x] Documentation only — no code changes.


felixmosh:
There is no good reason to specify the the chart is lazy loaded in docs (nor the usage of rechart)

felixmosh (approved):
Thank you

Issue #1173, opened 2026-04-23
Open to adding Claude Code Review + a small CLAUDE.md?


Stosiu:
Hey @felixmosh, been thinking about CI for a bit, wanted to check before I send anything.

Three small things, ranked by what I think actually helps here:

1. Claude Code Review (`anthropics/claude-code-action`). Free on public repos. I'd wire it as mention-only, so it runs when someone comments `@claude review` on a PR and stays quiet otherwise. Why I think it earns a slot here specifically: 10+ framework adapters and one shared `packages/api`, so a change in the core can quietly break one adapter and not show up in review. An extra pass would catch that kind of thing.

2. A `CLAUDE.md` at the root. Short one, basically the stuff I keep having to figure out by greping: monorepo layout, that tests only live in `packages/api`, the Redis dev loop, oxlint vs Prettier. Helpful for anyone using Claude Code or Copilot, also just decent onboarding docs. I'd keep it tight so it doesn't overlap the README.

3. Codecov on `packages/api`. Tests are there, coverage isn't visible anywhere, would be nice to see it move when someone touches the adapters. OpenSSF Scorecard would give you a security badge too if you want it, runs weekly on its own, no babysitting.

Same deal as the docs one, happy to send each as a separate PR or skip any of them. If AI review isn't your thing that's fine, wanted to ask first rather than drop a workflow PR you'd have to close.

felixmosh:
I'll more than happy to add it... Not sure what is the benefit, but ok.

I prefer to increase test coverage of the API + each adapter, this will give us a better observability.

Now that Bull-board get some usage, we should make it more bulletproof

PR #1219, opened 2026-06-16
test: contract tests for every server adapter, plus CLAUDE.md (#1173)

Stosiu:
Follow-up to #1173. You said the thing you actually wanted was more test coverage across the API and each adapter, so I started there instead of with the CI bits.

This PR does two of the three things from the issue: the test coverage and the `CLAUDE.md`. The other two (Claude review, Codecov) I left out so this stays about tests. More on that at the bottom.

## Test coverage (the main thing)

Before this, only `packages/api` had tests and none of the nine adapters did. Hand-writing a separate suite per adapter tends to drift, with each one ending up checking slightly different things. So there's one contract instead (`@bull-board/test-utils`, private, not published) that every adapter runs against. Wiring up an adapter is about 25 lines: start the framework, hand back a `request(...)` function, done.

Every adapter runs the same 8 checks against real Redis and real BullMQ:

- serves the dashboard HTML with the injected basePath and UI config
- serves static assets
- `GET /api/queues` returns the queues as JSON
- `POST .../add` actually adds a job (verified by reading it back out of Redis, not just the status code)
- `PUT .../pause` returns success and the queue really is paused
- an unknown queue returns a 4xx, not a 200
- everything still resolves when the board is mounted under a `/ui` prefix
- that prefix shows up in the entry HTML

Covered: Express, Fastify, Hono, Koa, Hapi, H3, Elysia, NestJS, Bun. All of them.

## It found real bugs

This is the part that convinced me the suite was worth it. It caught three things that were already broken:

- **H3** threw away the status code the handlers return, so everything came back `200`, including errors, because it was returning h3 errors instead of throwing them. An unknown queue answered `200` with the error as the body.
- **Bun** only registered the entry route at `/ui`, never `/ui/`, so the trailing-slash URL 404'd.
- **Elysia** listed its static files with `fs.glob`, which only exists on Node 22+, so the adapter was broken on Node 20, the minimum this repo supports. The Node version matrix is what surfaced it. Swapped it for a recursive `readdir`.

All three are genuine adapter bugs, each fixed in its own `fix(...)` commit, not papered over in the tests.

## Per-adapter notes

- **Express** runs the whole suite twice, against express@4 and express@5, since v5 changed routing. Both pass.
- **Fastify** is v5 only. Its bundled `@fastify/static@9` and `@fastify/view@11` refuse to register on fastify@4, so the adapter doesn't actually support v4. The version matrix is what surfaced that.
- **Bun** runs under `bun test` (it needs the Bun runtime), so it has its own CI job and is excluded from the Node `yarn test`.
- **NestJS** boots a real Nest app and injects the seeded queue through the board instance. No production change needed there.
- While wiring this up I also deduped `bullmq` in the lockfile. `api` and `nestjs` were on 5.78.0 while everything else was on 5.78.1, which caused a type clash across the two copies.

## CLAUDE.md

A short root file: the monorepo layout, the Redis-backed test loop, oxlint vs prettier, and a section on how the adapter contract works and how to cover a new adapter. Kept tight so it doesn't just repeat the README.

## Not in this PR

Still happy to send these as separate PRs if you want them, same as the issue said:

- Claude Code Review, mention-only (`@claude review` on a PR, quiet otherwise)
- Codecov on `packages/api` / OpenSSF Scorecard. The coverage exists now, this would just make it visible.

## CI

Lint, build, and the full suite pass locally. CI runs everything on Node 20/22/24, plus a separate job for the Bun adapter.


.github/workflows/nodejs.yml
```diff
@@ -62,3 +62,48 @@ jobs:
...
+        run: yarn install --immutable
+        env:
+          CI: true
+
+      - name: Build
+        run: yarn build
+        env:
+          CI: true
+
+      - name: Test Bun adapter
```
felixmosh, line 104: Am I missing the test of rest of the packages?

felixmosh (changes requested):

.github/workflows/nodejs.yml
Stosiu: They're in the main `Build, lint, and test` job, not this one. That job runs `yarn test` on the Node 20/22/24 matrix, and `yarn test` is just `yarn workspaces foreach ... run test`, so it hits every workspace with a `test` script: `api` plus the Express, Fastify, Hono, Koa, Hapi, H3, Elysia, and NestJS adapters.

Bun is the odd one out. It needs the Bun runtime (`bun test`, not jest), so it can't go through that foreach and gets this dedicated job instead. That's the only reason this step just mentions `@bull-board/bun`.

So nothing's missing: `api` and the eight jest adapters run in the matrix job, Bun runs here. If that's confusing to read, I can drop a comment in the workflow, or split the matrix test step out per package so each one shows up as its own check.

ltm what do you think

felixmosh (approved):
Thank you for this PR, great as always 🙏🏼

PR #1221, opened 2026-06-17
test(api): measure coverage from source and raise it (62% to 90%)

Stosiu:
## What this does

`@bull-board/api` had no usable coverage number. The tests import `@bull-board/api` from `dist/`, but the coverage tool instruments `src/`, so `jest --coverage` reported 0% even with everything passing.

This adds a coverage-only jest config that maps `@bull-board/api*` to `src/`, then fills the biggest untested gaps with real integration tests (real Redis, BullMQ and Bull, no mocks).

| Metric | Before | After |
|---|---|---|
| Statements | 62.26% | 89.89% |
| Branches | 53.04% | 78.37% |
| Functions | 50.00% | 84.55% |
| Lines | 63.59% | 91.43% |
| Tests | 60 | 92 |

The new tests cover `providers/flow.ts` and the legacy `bull.ts` (both were 0%), the flow handler, the pause/resume-all handlers, and the entry-point and error handlers.

`yarn test` is untouched and keeps full type-checking. Only the coverage run relaxes it: the `src` vs `dist` `BaseAdapter` split trips type-checking, so the coverage config uses `isolatedModules`.

## Codecov

The wiring (`codecov.yml` plus the CI upload step) is already in this PR. Until a `CODECOV_TOKEN` secret exists the upload is non-blocking, so CI stays green and nothing changes. You can merge first and enable it later.

To turn it on:

1. Go to https://codecov.io and sign in with GitHub. It's free for open source. Authorize it for `felixmosh/bull-board`.
2. On the repo's Codecov settings page, copy the upload token.
3. In GitHub: **Settings → Secrets and variables → Actions → New repository secret**. Name it `CODECOV_TOKEN` and paste the value.

After that the next CI run uploads coverage, and the `api`-scoped gate starts posting a status check and a PR comment. It blocks if api coverage drops more than 1% from the base and asks for 75% on changed lines.

Two things worth knowing:

- Coverage uploads only from the Node 24.x job, so it runs once per CI rather than on every matrix leg.
- Codecov has mostly dropped tokenless uploads, so the secret is the reliable path rather than relying on the action picking it up automatically.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -63,6 +63,23 @@ jobs:
           # The default Redis port
           REDIS_PORT: 6379
 
+      - name: Coverage (@bull-board/api)
+        if: matrix.node == '24.x'
+        run: yarn workspace @bull-board/api test:coverage
+        env:
+          CI: true
+          REDIS_HOST: localhost
+          REDIS_PORT: 6379
+
+      - name: Upload coverage to Codecov
+        if: matrix.node == '24.x'
+        uses: codecov/codecov-action@v5
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: ./packages/api/coverage/lcov.info
+          flags: api
+          fail_ci_if_error: false # non-blocking until CODECOV_TOKEN is set
+
   test-bun:
     name: Test Bun adapter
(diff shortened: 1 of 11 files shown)


felixmosh (approved):
Thank you

Issue #485, opened 2022-11-17
bull-board not working with AWS serveless 


GBShadow:
I have a problem with the bull-board, I created my bull-board application with nest and using AWS Lambda, it works normally, but the bull-board does not, when I access the route only the text "Loading..." appears, but If I start without serverless, the bull-board works normally, does anyone have any idea what it could be?

felixmosh:
Hi @GBShadow I think that if you will ask in English, it will approach more ppl.

Anyway, I've translated your message, and looks like you have an issue with Nest & aws serverless, I'm not familiar with this setup, so If someone knows it, I will be more that happy for the help.

samuelcole:
This is a duplicate of #444, in my opinion.

bull-board does not work in serverless environments because it uses full paths to load some files.

felixmosh:
In serverless env the path to static assets of the ui is not calculated properly,
You can change the path to a the proper value by calling
```js
const uiBasePath = YOUR_CORRECT_VALUE
serverAdapter
  .setViewsPath(path.join(uiBasePath, 'dist'))
  .setStaticPath('/static', path.join(uiBasePath, 'dist/static'));
```
 **after** you pass it to `createBullBoard` function.

Can you try this and let me know if it helped you?

samuelcole:
same error, i believe it's throwing on this line:

https://github.com/felixmosh/bull-board/blob/06a1c9828c5c4873172981e60bf85d7d2fc96264/packages/api/src/index.ts#L18

which is executed when running `createBullBoard`

if we could pass a `uiBasePath` to `createBullBoard` instead of it being calculated, that would fix it?

felixmosh:
Probably... But i think that serverless is not a good fit anyway, since the board polling its api, you will connect to redis for each request (since serverless env is stateless)
What is the error that you get?

samuelcole:
same error as before: https://github.com/felixmosh/bull-board/issues/444#issuecomment-1235734277

felixmosh:
@samuelcole, Can you try `v5.2.0` which allows you to pass `uiBasePath` from the outside?

```js
createBullBoard({
    queues: [new BullMQAdapter(exampleBullMq)],
    serverAdapter,
    options: { uiBasePath: 'YOUR_PATH_TO_@bull-board/ui' } // <---
  });
```

fabiolnm:
@samuelcole this doesn't seem to work because node_modules isn't included into the resulting Vercel build source.

[...]

Issue #124, opened 2020-07-31
How to use with next.js (not custom express server) ?


trompx:
Hello,

All examples I've found use bull-board with a custom express server.
Is there a way to use it with next.js own server (cannot use app.use) for instance in a pages/bull-dashboard.js file ?

SahanAmadoruge:
i'm looking for an answer for the same issue.

noclat:
Here's how I made it work: 

```js
import Queue from 'bull';
import { setQueues, BullAdapter, router } from 'bull-board';
import nextConnect from 'next-connect';

setQueues([
  new BullAdapter(new Queue('main')),
]);

// tell Express to prepend all paths
router.use('/api/queues', router._router);

// tell Next.js to execute Express routes
// directly calling handler.use('/api/queues', router._router.handle); throws an error, idk why
const handler = nextConnect();
handler.use('/api/queues', (req, res, next) => {
  return router._router.handle(req, res, next);
});

export default handler;
```

A bit ugly but I've found a workaround. Underneath it's basically just an Express application. There's a way to execute an express route without launching the server: `app._router.handle` ([source](https://soyuka.me/call-framework-routes-programmatically/).

But it needs a bit more understanding when combining it with Next.js: we have to tell Next.js to route every subpath to that specific handler, and we have to tell the Express app to change base path for all its routes.

So to handle Next, we just need to create `/pages/api/queues/[[..path]].js`, where `[[..path]]` is the catch-all convention of Next.js. And keep in mind that as defined in the [Next.js doc](https://nextjs.org/docs/api-routes/introduction), API routes are only possible inside the `/pages/api` folder.

Then, we have to tell Express that all URLs will be prepended with `/api/queues`, the same way we'd do with Express, so using: `app.use('/path', handler)`. But the confusion here is that the whole app is exposed as `router` by the `bull-board` package. So when importing `import { router } from 'bull-board';`, `router` is actually a full Express app definition.

To make sure I don't mess with Express, I've also used [next-connect](https://www.npmjs.com/package/next-connect) to mimic the `next` argument.

**Disclaimer:** I'm still in the process of testing this. Haven't actually run a job and observed if it shows in the UI yet, but the UI is showing in full. This progress was worth sharing in itself.

jafework:
@noclat Thanks! this worked, only change is `[[..path]].js` should be `[[...path]].js`

dstotz:
[...]

PR #1224, opened 2026-06-17
feat: Next.js & Vercel support (examples, docs, regression tests)

### LINKED ISSUE #444: Next.js & Vercel Support
There's an ongoing thread #124 here about how to integrate `bull-board/express` with Nextjs. The latest solution works great, however, if you are deploying to Vercel you [run into issues](https://github.com/felixmosh/bull-board/issues/124#issuecomment-1222951173) trying to access `@bull-board/ui/package.json`. 

Here's the [main issue](https://github.com/felixmosh/bull-board/blob/master/packages/api/src/index.ts#L16), I believe
```ts
const uiBasePath = path.dirname(eval(`require.resolve('@bull-board/ui/package.json')`));
```

Vercel is not going to give access to the full path of their server... so pretty sure relative paths need to be used instead. Something like...

```ts
  const uiBasePath = "node_modules/@bull-board/ui"
  serverAdapter
    .setQueues(bullBoardQueues)
    .setViewsPath(uiBasePath + 'dist')
    .setStaticPath('/static', uiBasePath + 'dist/static')    
```

Also, it'd be great to see official Nextjs support in the documentation.

Stosiu:
Closes #444.

## The problem

Deploying bull-board inside Next.js on Vercel dies at runtime with:

```
Error: Cannot find module '@bull-board/ui/package.json'
```

`@bull-board/api` finds the UI assets with `eval(require.resolve('@bull-board/ui/package.json'))`. The `eval` hides that require from bundlers, and Next's file tracer (`@vercel/nft`) is one of them, so the UI never gets copied into the serverless function. Open since 2022 with only workarounds in the thread (#124, #310, #485 are the same root cause).

## The fix

No library change needed. Next 15+ has the config to deal with this, and `uiBasePath` already exists as an escape hatch. Two settings in `next.config.js`:

- `serverExternalPackages` — keep bull-board and bullmq out of the bundle so they resolve from `node_modules` at runtime.
- `outputFileTracingIncludes` — force `@bull-board/ui/dist` into the function, since the tracer can't follow the `eval`.

This documents the workaround rather than removing the need for it. The root cause is the `eval`; making the tracer able to follow it (package `exports`, or a resolvable UI path) would drop the config requirement entirely, but that touches how every adapter resolves the UI and is a bigger, riskier change. Happy to open a follow-up for it.

## What's in here

- `examples/with-nextjs-app` — App Router, Hono adapter.
- `examples/with-nextjs-pages` — Pages Router, Express adapter (the #124 pattern, made deployable). Both ship a worker, a docker-compose for Redis, and a README.
- A docs recipe (Next.js & Vercel), linked from the recipes and server-adapters pages.
- Tests. This question keeps resurfacing (the issue has sat open since 2022 with a recurring trickle of reports), so it's worth guarding against regressions:
  - `ui-base-path.spec.ts` — a unit test (no Redis) that the default UI path resolves to real files and that `uiBasePath` overrides it.
  - `scripts/check-nextjs-trace.mjs` plus a workflow that builds each example on Next 15 and 16 and asserts the UI assets land in the function's `.nft.json`. This catches the actual bug. `next dev` doesn't, because dev resolves straight from `node_modules` and stays green even when the deployed build is broken.

## What I checked

Both examples build and serve on Node 22 with Next 15 and 16 against a local Redis: dashboard, static assets, API, adding jobs. Pull `outputFileTracingIncludes` back out and the trace check drops to 0 files and fails, which is the whole point of having it.

## Notes

- Next 15+ only (these keys lived under `experimental` in 13/14).
- Workers can't run in serverless, so both examples run `worker.ts` as a separate process.
- The example builds use the published `@bull-board` packages, so they cover Next-version drift; the unit test covers the library's own path resolution against working-tree code.

### DIFF
--- a/.github/workflows/nextjs.yml
+++ b/.github/workflows/nextjs.yml
@@ -1,0 +1,58 @@
+name: Next.js examples
+
+# Builds the Next.js examples and checks the @bull-board/ui assets get traced into
+# the serverless function (#444). Not run per-commit — building an example is overkill
+# and the adapter has its own tests. Manual, plus once post-merge when the example
+# changes. The library's own resolution is covered by ui-base-path.spec.ts in main CI.
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+    paths:
+      - 'examples/with-nextjs-app/**'
+      - 'examples/with-nextjs-pages/**'
+      - 'scripts/check-nextjs-trace.mjs'
+      - '.github/workflows/nextjs.yml'
+
+jobs:
+  build-and-trace:
+    name: Build examples and check asset tracing
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v6
+
+      - name: Use Node 22.x
+        uses: actions/setup-node@v6
+        with:
+          node-version: 22.x
+
+      - name: Build each example on each Next.js version and check tracing
+        run: |
+          set -uo pipefail
+          failures=()
+          for example in with-nextjs-app with-nextjs-pages; do
+            for next in 15 16; do
+              echo "::group::$example on next@$next"
+              if (
+                set -e
+                cd "examples/$example"
+                npm install --no-audit --no-fund
+                npm install --no-audit --no-fund "next@$next"
+                rm -rf .next
+                npm run build
+              ) && node scripts/check-nextjs-trace.mjs "examples/$example"; then
+                echo "PASS: $example on next@$next"
+              else
+                echo "FAIL: $example on next@$next"
+                failures+=("$example@next$next")
+              fi
+              echo "::endgroup::"
+            done
+          done
+          if [ ${#failures[@]} -ne 0 ]; then
+            echo "Failed: ${failures[*]}"
+            exit 1
+          fi
--- a/examples/with-nextjs-app/.gitignore
+++ b/examples/with-nextjs-app/.gitignore
@@ -1,0 +1,9 @@
+/node_modules
+/.next
+/out
+next-env.d.ts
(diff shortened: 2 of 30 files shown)


.github/workflows/nextjs.yml
```diff
@@ -0,0 +1,57 @@
+name: Next.js examples
+
+# Builds the Next.js examples and asserts the @bull-board/ui assets are traced
+# into the serverless function (regression guard for #444). The examples install
+# the *published* @bull-board packages, so this checks the example config and
+# Next.js version compatibility; the library's own path resolution is covered by
+# packages/api/tests/api/ui-base-path.spec.ts in the main CI job.
+on:
```
felixmosh, line 7: why do you think that we need a workflow to check one of the examples?

felixmosh (changes requested):

.github/workflows/nextjs.yml
Stosiu: Fair question. It's not the adapter I wanted to guard, that's got its own tests. It's the Next config. These examples only exist for the `outputFileTracingIncludes` setup, and that breaks silently: `next dev` works because it resolves from `node_modules`, so the one way to catch a broken config is to build and check the UI assets actually landed in the trace.

I've already dropped the per-commit trigger. It's manual now, plus one run when the example changes on `master`. The always-on guard is a small unit test in the `@bull-board/api` suite, not the workflow.

Happy to drop the workflow entirely and keep just the unit test and a manual check at merge if you'd rather not carry it

felixmosh (approved):
Thank you for this PR 🙏🏼

PR #1225, opened 2026-06-18
chore: enforce formatting and linting in CI

Stosiu:
## Why

`chore: reformat code` commits keep landing manually because nothing enforces oxfmt: the `Lint` CI step runs oxlint with `--fix`, which auto-fixes and passes instead of failing, and `lint-pr` only checks the title. So both formatting and lint drift stay invisible until someone notices.

## What

- Add a **`Format check`** CI step + `format:check` script (`oxfmt --check`) that fails the PR instead of fixing formatting post-merge.
- Point CI at a new `lint:check` script (`oxlint` without `--fix`) so lint actually gates. `lint` keeps `--fix` for local use.
- Extend the format glob to `examples/`. The old trailing `.` was a no-op once an explicit glob is passed, so `yarn format` never covered examples and they'd drifted (19 `.ts` files, now reformatted).
- Remove the three NestJS examples' own prettier setup (`.prettierrc` using a conflicting `trailingComma: all`, `format` scripts, deps, `eslint-plugin-prettier`) so everything formats under the one root config.

`master` already passes both checks, so the gates are green from day one.

## Notes

- Format scope is `ts/tsx` only, matching the existing convention; example `.js` and md/json/css are untouched.
- Lint stays scoped to `packages/` (examples intentionally use `console.log`, which the `no-console` rule would flag).
- Skipped a pre-commit hook: bot/web contributors never run hooks, so only a CI gate actually catches the drift.


felixmosh (approved):
Thank you for this improvemet :]

Issue #788, opened 2024-07-26
[QUESTION] NestJS + Typescript + BullMQ + Bull Board Production Deployment with Nginx as Reverse Proxy


tirth0:
Greetings, I have deployed bullboard with bullmq and nestjs with the following configuration:
There is a local redis instance on the remote server.
This is in the imports section in app.module.ts:
```
    BullModule.forRoot({
      connection: {
        host: 'localhost',
        port: <PORT>,
      },
    }),
    //register the bull-board module forRoot in your app.module
    BullBoardModule.forRoot({
      route: "/queues",
      adapter: ExpressAdapter
    }),
```

This works absolutely fine in local setup, but once deployed on production  it gets stuck on a loading page:
![Screenshot 2024-07-26 at 7 18 20 PM](https://github.com/user-attachments/assets/6bdd7a8c-e821-4116-8f95-9c9eb481742e)

In Nginx, my app is served under a subpath `app` making the URL : https://my-app.com/app/queues

I want to know what configuration changes is required for me to set up bullboard UI properly with my configuration.
Any help is appreciated!

felixmosh:
It stuck in loading since all app files are not downloaded (your server returns 404 for .css, .js) files.
Looks like an nginx misconfig, can you share your nginx config?

tirth0:
My configuration for the location block in nginx is as follows:

```
    location /app/ {
      proxy_set_header X-Real-IP $remote_addr;
      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
      proxy_set_header X-NginX-Proxy true;
      proxy_ssl_session_reuse off;
      proxy_set_header Host $http_host;
      proxy_cache_bypass $http_upgrade;
      proxy_redirect off;
      client_max_body_size 100M;
      proxy_read_timeout 600;
      proxy_connect_timeout 600;
      proxy_send_timeout 600;
      proxy_headers_hash_max_size 512;
      proxy_headers_hash_bucket_size 128;
      proxy_set_header Upgrade $http_upgrade;
      proxy_set_header Connection "upgrade";
      proxy_http_version 1.1;
      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
      proxy_set_header Host $host;
      proxy_pass http://port
    }
```

In the dist folder for the nest app, there is only a main.js file after the build step. Where should I be looking for the static files?

felixmosh:
Do you ave any block that configures static files?

[...]

Issue #428, opened 2022-07-11
Bull stuck into 'Loading' cause missing .js file


felixa1996:
When I develop in my local machine without docker it's work seamless.
But when I move to production server, Bull-board is stuck on loading,
After I inspect, I know cause I have API Gateway inside my production server

Here I attached the screenshot below
I access with http://mydomain.com:8000/serviceName/queue
[https://pasteboard.co/HDl7oBQ67cbS.png](https://pasteboard.co/HDl7oBQ67cbS.png)
I think bull cant resolve the asset file 
[https://pasteboard.co/0YtAQp4KY4IU.png](https://pasteboard.co/0YtAQp4KY4IU.png)

Btw I am using bull-board/koa
I mean there is an API for change the static file routes
```ts
static monitor(basePath: string): KoaAdapter {
		const serverAdapter = new KoaAdapter();
		serverAdapter.setBasePath(basePath);

		createBullBoard({
			queues: Array.from(queuesList.values()).map(
				(queue) => new BullAdapter(queue)
			),
			serverAdapter: serverAdapter,
		});

		return serverAdapter;
	}
```

I already look into the source file, but cant find one that suitable for my use case
[https://pasteboard.co/Nnx5fNp3zQUz.png](https://pasteboard.co/Nnx5fNp3zQUz.png)

felixmosh:
Hi, thank you for reporting this issue.

Looks like it is related to your setup.
Let's try to debug.

1. Do you have the static assets inside the docker?
2. The html contains `<base>` tag, what its value?
3. When you call `serverAdapter.setBasePath(basePath);`, what is the value of `basePath`?
4. What is the path that the board mounted in?

felixmosh:
Ping?

yogeshattwic:
I have the same issue when using the bull-dashboard and the dashboard is behind a gateway.  I am using express adapter.
It works fine if I directly access dashboard on my local environment (http://localhost:3000/v1/xxxx/admin/queues/).
But when accessing the same through gateway, I get `Loading...` page. (http://localhost:9000/dashboard/v1/xxxx/admin/queues). I have a rule in Gateway that will replace the path /dashboard with '' (empty string).

felixmosh:
@yogeshattwic this issue is not related to bull-board, you probably didn't configured your reverse proxy properly, therefore, static assets are not loading.

yogeshattwic:
[...]

PR #1230, opened 2026-06-24
feat(nestjs): add basePath option to decouple UI base path from mount route

bueno12223:
## What

Add an optional `basePath` to `BullBoardModule.forRoot(...)` (and `forRootAsync`) for `@bull-board/nestjs`. It overrides the base path injected into the dashboard UI (used for assets, the client-side router and API calls) **without** changing the route the app mounts on.

```ts
BullBoardModule.forRoot({
  route: '/queues',          // where Nest mounts/serves the board
  basePath: '/admin/queues', // public path the browser sees
  adapter: ExpressAdapter,
});
```

When omitted, behavior is unchanged (`basePath` defaults to `route`, prefixed with the app's global prefix), so this is fully backward compatible.

## Why

Today the wrapper derives the UI base path from the mount route:

```ts
const prefix = addForwardSlash(this.applicationConfig.getGlobalPrefix() + this.options.route);
this.adapter.setBasePath(prefix);
```

There is no way to set the base path independently of the route. That breaks when the board sits behind a reverse proxy / API gateway that **rewrites the path** — the public URL differs from the route the app listens on:

- Gateway exposes the board at `https://host/admin/queues` and strips `/admin` before forwarding, so the app must mount at `/queues`.
- But the UI then emits `<base href="/queues/">`, so the browser requests assets at `/queues/static/...` instead of `/admin/queues/static/...` → 404 → the dashboard hangs on "Loading".

This is a recurring report (e.g. #428, #788). The raw adapters already expose `setBasePath()` for exactly this, but the NestJS wrapper calls it internally and gives no way to override. This PR exposes it as an option:

- `route: '/queues'` → app serves and assets are served here (after the gateway strips the prefix).
- `basePath: '/admin/queues'` → UI references its assets/API at the public path; the gateway strips `/admin` back to `/queues` where the app serves them.

The fastify mount prefix is intentionally left on the computed `route`-based prefix; only the UI base path uses the override.

## Tests

Added `packages/nestjs/tests/base-path.spec.ts`:
- injects `options.basePath` into the UI while still serving at the mount route,
- normalizes a `basePath` missing a leading slash,
- falls back to the mount route when `basePath` is not provided.

All `@bull-board/nestjs` tests pass (11/11), lint clean.

### DIFF
--- a/packages/nestjs/src/bull-board.root-module.ts
+++ b/packages/nestjs/src/bull-board.root-module.ts
@@ -48,7 +48,10 @@ export class BullBoardRootModule implements NestModule {
       ? addForwardSlash(this.options.route)
       : addForwardSlash(this.applicationConfig.getGlobalPrefix() + this.options.route);
 
-    this.adapter.setBasePath(prefix);
+    const basePath =
+      this.options.basePath != null ? addForwardSlash(this.options.basePath) : prefix;
+
+    this.adapter.setBasePath(basePath);
 
     if (isExpressAdapter(this.adapter)) {
       return consumer
--- a/packages/nestjs/README.md
+++ b/packages/nestjs/README.md
@@ -61,6 +61,7 @@ The following options are available.
 - `adapter` The routing adapter to be used, either the Express Adapter or Fastify Adapter provided by bull-board.
 - `boardOptions` options as provided by the bull-board package, such as `uiBasePath` and `uiConfig`
 - `middleware` optional middleware for the express adapter (e.g. basic authentication)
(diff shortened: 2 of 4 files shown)


packages/nestjs/src/bull-board.root-module.ts
```diff
@@ -48,7 +48,10 @@ export class BullBoardRootModule implements NestModule {
       ? addForwardSlash(this.options.route)
       : addForwardSlash(this.applicationConfig.getGlobalPrefix() + this.options.route);
 
-    this.adapter.setBasePath(prefix);
+    const basePath =
+      this.options.basePath != null ? addForwardSlash(this.options.basePath) : prefix;
```
felixmosh, line 52: Why there is an assumption that the basePath is `null` by default?

felixmosh (changes requested):

PR #1231, opened 2026-06-25
fix(ui): wrap long values in CollapsibleJSON to prevent overflow

svenickx:
## What

Long unbreakable string values (JWTs, IDs, tokens…) in the `CollapsibleJSON` job-data viewer overflowed their container instead of wrapping.

## Fix

Added `overflow-wrap: anywhere` to the value classes (`stringValue`, `numberValue`, `otherValue`) in `CollapsibleJSON.module.css`.

## Before

image

## After

image


felixmosh (changes requested):
Thank tou for the fix, 
Will you extract that style to "value"  class, and add it on component level?

svenickx:
Done 👍

PR #1238, opened 2026-06-30
feat: grouped queue overview, collapsible sidebar, and queue info panel

Stosiu:
## Why

I wanted better visibility into what's actually running. Most of my projects already name queues with delimiters, but only the sidebar turned that into a tree. The overview stayed a flat wall of long labels like `Emails.Transactional.Welcome`, which gets confusing fast. So this teaches the overview to group the same way the sidebar does, and lets you default to that grouped view when your names already have the structure.

## What's in it

**Grouped overview.** Collapsible category sections in the main view, built from the same delimiter the sidebar uses. Each category header sums the counts underneath it, so you can read a whole domain's health without expanding it. You can pause or resume a category from its header, collapse everything at once, and an active search auto-expands the groups that match. It's off by default; set `overview.groupByDelimiter: true` in `UIConfig` to make grouped the default.

Grouped overview

**Group queues setting.** The flat/grouped switch lives in Settings, under a new Queues section, with a one-line description so it's clear what it does. The `UIConfig` option is a real default: it sets the starting view, but once a user toggles it their choice is saved and wins on reload.

Group queues setting

**Collapsible sidebar.** A toggle in the header hides the sidebar for a full-width view. It's saved to localStorage, so it reopens how you left it.

Sidebar collapsed

**Search shortcut.** Search stays in the sidebar, but `⌘K` (`Ctrl K` on Windows and Linux) now focuses it from anywhere, opening the sidebar first if it's collapsed.

**Queue info panel.** An info icon next to a queue's name opens a panel with its type, paused state, global concurrency, and default job options (attempts, backoff, retention). It saves a trip to the code when you just want to check a queue's defaults.

Info icon on the queue page

Queue info panel

**Settings cleanup.** The settings modal had grown to about 15 toggles. I split it into collapsible General / Queues / Jobs sections, one open at a time.

## Notes for review

- One small API change powers the info panel: a `getQueueDefaultJobOptions()` method on the adapters (BullMQ and Bull, with an empty default on the base class), surfaced as `defaultJobOptions` on the queue payload. There are tests for it.
- The example app now models a realistic backend (Emails, Media, Payments, Notifications, Webhooks, Search) with workers producing a real spread of job states, so the grouped view isn't demoed against a single queue.
- New docs page, "Exploring the dashboard", plus the `overview.groupByDelimiter` field in the UIConfig reference.

Happy to trim the category header (counts plus pause/resume) if it feels like too much.


felixmosh:
Wow, thank you for this PR, looks awsome.

I'll need sometime to review it 😅.

Next time, it will help me to review smaller PRs.

Can you split this PR into smaller ones?

Stosiu:
Done — split into four focused PRs, each a standalone feature. Closing this one in favor of them. Suggested review order:

1. #1242 — **refactor: settings modal → collapsible sections.** Pure refactor, no behavior change. Foundation for the others.
2. #1243 — **feat: queue default job options + info panel.** Full-stack, self-contained, with tests.
3. #1244 — **feat: collapsible sidebar + ⌘K.**
4. #1245 — **feat: grouped queue overview** (the main feature). Builds on #1242, so that one first.

#1243 and #1244 are independent and can go in any order. Example-app and docs changes live in the PRs that need them. Thanks for taking the time to look.

PR #1239, opened 2026-06-30
fix(ui): improve autoscroll behavior in JobLogs component

AlexNastoldFIS:
### Summary
Fixes the play button auto-scroll feature that stopped working in v6.15.0.

### Solution
The `scrollTo()` call was targeting the outer container which has `overflow: hidden`, making it non-scrollable. Changed to use `querySelector` to target the inner `.preWrapper` element that actually has `overflow: auto`:

```typescript
const scrollableElement = logsContainer.current?.querySelector(`.${s.preWrapper}`);
scrollableElement?.scrollTo({
  top: scrollableElement.scrollHeight,
  behavior: 'smooth',
});
```

The view now correctly scrolls to the bottom when live logs are enabled.

---


felixmosh (approved):
Thank you for this fix 🙏🏼

PR #1242, opened 2026-06-30
refactor(ui): split settings modal into collapsible sections

Stosiu:
The settings modal had grown to about 15 toggles in one flat list. This splits it into three collapsible sections, General / Queues / Jobs, with one open at a time. No behavior change; every setting works the same as before.

Settings split into sections

First of the four PRs that #1238 was split into. It goes first because the others add their own settings into these sections.


felixmosh (approved):
This looks great, thank you 🙏🏼

PR #1243, opened 2026-06-30
feat: add queue default job options and info panel

Stosiu:
Adds an info panel for a queue, plus the small API change behind it.

There's a new `getQueueDefaultJobOptions()` on the adapters (BullMQ and Bull, with an empty default on the base class), exposed through a dedicated `GET /api/queues/:queueName/default-job-options` endpoint. The panel fetches it when it opens, so it stays off the queue polling payload. Tests included.

In the UI, an info icon next to the queue name opens a panel showing how the queue is set up: its type, paused state, global concurrency, and — when the queue has them — the default job options (attempts, backoff, retention). Queues without defaults simply don't render that section.

Info icon next to the queue name

Queue info panel with default job options

Queue info panel for a queue with no default job options — the section is hidden

Part of the #1238 split. The panel reuses the collapsible-section component from #1242, so it sits on top of that one and its diff includes #1242 until that merges.


packages/api/src/handlers/queues.ts
```diff
@@ -98,6 +98,7 @@ async function getAppQueues(
         type: queue.type,
         delimiter: queue.delimiter,
         globalConcurrency,
+        defaultJobOptions: queue.getQueueDefaultJobOptions(),
```
felixmosh: since this data is shown inside a modal, and it is relevant for BullMQ, I don't think that it should be added to be on Queue polling return value.

I think it is better to expose GET request to fetch it when the modal is open.

packages/ui/src/components/QueueInfoModal/QueueInfoModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+              <span className={s.muted}>—</span>
+            )}
+          </Row>
+          <Row label={t('QUEUE.INFO.TOTAL_JOBS')}>
+            <span className={s.mono}>{totalJobs}</span>
+          </Row>
+        </dl>
+      </CollapsibleSection>
+
+      <CollapsibleSection
```
felixmosh: Should not show a section if there are no values.
Showing an empty message inside an accordion, is weird :]

felixmosh (changes requested):

felixmosh:
Ping me when it is ready for review.

Stosiu:
@felixmosh ping

felixmosh (approved):

PR #1245, opened 2026-06-30
feat(ui): add grouped queue overview

Stosiu:
The overview used to be a flat wall of long labels like `Emails.Transactional.Welcome`, which gets confusing once you have a few of them. This teaches it to group by the delimiter in queue names, the same way the sidebar already does.

Grouped overview

- Collapsible category sections. Each header sums the counts of the queues under it, so you can read a whole domain's health, say all of Payments, without expanding it.
- Pause or resume a category from its header, collapse or expand everything at once, and an active search auto-expands the groups that match.
- Off by default. Flip it in Settings → Queues → Group queues by category, or set `overview.groupByDelimiter: true` in `UIConfig` to make grouped the starting view. The config is only a default: once someone toggles it, their choice is saved and wins on reload.
- The example app now models a realistic backend (Emails, Media, Payments, Notifications, Webhooks, Search) so the grouped view is demoed against real structure, and there's a new "Exploring the dashboard" docs page.

Last part of the #1238 split. Builds on #1242 (the toggle lives in its new Queues section), so that one goes first; until it merges this diff includes those commits. The docs page is a full tour, so it also covers the sidebar (#1244) and info panel (#1243).


felixmosh:
This looks soooo good... thank you.

Few things to improve:

1. The Sticky header is not sticky
2. There is `index` issue with the group header image
3. The pause button, on each group title, I think that it should follow the same convention of 3-dots menu. (I see a following requirement that, "allow to set global concarency for the entire group" 😅)

Overall, these are really good features to have 
(BTW, example.ts changes are really cool...)

felixmosh (approved):

felixmosh:
Screenshot 2026-07-01 at 17-19-07 —

Will you able to resolve conflicts?

felixmosh:
Thank you!
All pr's were released in V8.1.0

PR #1246, opened 2026-06-30
fix(nestjs): support queues with the same name but different prefix

### LINKED ISSUE #1229: `@bull-board/nestjs`: Queues with same name but different prefix are not handled correctly
When registering two Bull queues with the same `name` but different `prefix` values (e.g. to separate namespaces or Redis instances), `BullBoardModule.forFeature` resolves both to the same underlying queue object. As a result, both entries in the board display data from the same Redis queue – one of them is effectively invisible.

The issue is in `BullBoardFeatureModule.onModuleInit()`, where the queue is looked up from the NestJS DI container using only the base queue name, without considering the prefix:

https://github.com/felixmosh/bull-board/blob/0e426e9e40b6ef36eda4ccaf7b0758bb783f7528/packages/nestjs/src/bull-board.feature-module.ts#L18

Since two queues with the same `name` share the same DI token regardless of their `prefix`, NestJS returns the same queue instance for both lookups.

```ts
BullModule.registerQueue(
  { name: 'emails', prefix: 'tenant-a', redis: redisA },
  { name: 'emails', prefix: 'tenant-b', redis: redisB },
)

BullBoardModule.forFeature(
  { adapter: BullAdapter, name: 'emails', options: { prefix: 'tenant-a:' } },
  { adapter: BullAdapter, name: 'emails', options: { prefix: 'tenant-b:' } },
)
```

Both board entries show data from the same queue, even though they are configured with different prefixes and point to different Redis keyspaces.

Stosiu:
Fixes #1229

## The problem

Register two queues with the same `name` but a different `prefix` (a normal multi-tenant setup) and only one of them shows real data on the board. The other is a ghost.

It comes down to the lookup. `BullBoardFeatureModule` resolves each queue from Nest's DI container by name, and `@nestjs/bullmq` builds that token from the name alone, ignoring `prefix`. Two same-named queues collide on one token, so the lookup hands back the same instance twice. You can't fix that in the lookup; the second instance never gets a token of its own.

The `@bull-board/api` core is fine, by the way. It keys queues by `getName()` (`prefix` + name), so every other adapter already handles this. NestJS was the odd one out because it goes through DI.

## The fix

`forFeature` now takes a queue instance directly via a new `queue` field, skipping the DI lookup. `BullBoardQueueOptions` is now a union: pass `name` (DI lookup, same as before) or `queue` (your own instance). Nothing existing breaks.

## Tests

- `packages/api`: a regression test pinning the core behavior. Same name, different prefix, both stay distinct. Covers every non-Nest adapter in one shot.
- `packages/nestjs`: one test for the new `queue` path (both queues register separately), one for the old `name` path so it can't regress.

Docs updated too: `packages/nestjs/README.md` and the NestJS adapter page on the docs site.

### DIFF
--- a/packages/api/tests/api/same-name-different-prefix.spec.ts
+++ b/packages/api/tests/api/same-name-different-prefix.spec.ts
@@ -1,0 +1,52 @@
+import { createBullBoard } from '@bull-board/api';
+import { BullMQAdapter } from '@bull-board/api/bullMQAdapter';
+import { ExpressAdapter } from '@bull-board/express';
+import { Queue } from 'bullmq';
+import request from 'supertest';
+
+const connection = {
+  host: process.env.REDIS_HOST || 'localhost',
+  port: +(process.env.REDIS_PORT || 6379),
+};
+
+// Regression test for https://github.com/felixmosh/bull-board/issues/1229
+describe('queues with the same name but a different prefix', () => {
+  let serverAdapter: ExpressAdapter;
+  const queueList: Queue[] = [];
+
+  beforeEach(() => {
+    serverAdapter = new ExpressAdapter();
+    queueList.length = 0;
+  });
+
+  afterEach(async () => {
+    for (const queue of queueList) {
+      await queue.close();
+    }
+  });
+
+  it('keeps both queues as distinct board entries', async () => {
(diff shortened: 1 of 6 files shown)


felixmosh (approved):

felixmosh:
Thank you for the fix, (I'm not a Nest.js guy), looks OK :]

Issue #348, opened 2021-11-30
useInterval can cause problems if response is slower than 5000


yaro90:
Hi, 
Thanks for this very useful project, its a great idea. However I started to see a problem when my queue grew in size. Bull-board response dropped to around 30 sec to populate jobs which is still acceptable but the problem is that due to the use of setInterval at 5 sec for polling the page starts to hang after 2 min and also the backend gets stressed as requests pile up faster than they can complete. I'm not sure if I will have the time for PR, but it would be better to create the cycle using setTimeout as it would allow for sequential poll and new poll would only be made after the previous has completed.

felixmosh:
Hi, thank you for reporting the issue.

There are several possible ways to solve it:
1. Add polling timing config as suggested in #332 
2. Change interval usage to settimeout, it was like this, and interval solved some other issue, need to be checked.

yaro90:
1. Pause auto-refresh or configurable refresh time is a good feature but it would not solve properly the response being slower than request interval, we don't know how long the response would take so its a guess that you pick interval long enough.

2. I've had a look at what you had there before and it looks like people complained that the UI was re-rended after refresh. I mean if the backend state has changed there is no way to avoid re-rerender as the store would be updated. I did find the setTimeout logic somewhat odd though, but I would ignore 80% of the comments from that issue, I don't think much was solved there actually by changing to setInterval. The only issue I can see though from by brief overview if I understood it correctly, you are using useInterval inside useStore which is shared across all components (queues and status tabs) so perhaps the issue was that if you switch tabs and queues quickly and you use setTimeout (sequential requests) like you had before then when the request finishes you are already in a different tab so you end up with the wrong data in the wrong tab. Set interval does not really solve this issue, but it rather hides it and causes the bug i have mentioned above.

The real solution I believe would be to have a setTimeout sequential poll per status tab, when the user enters the tab, the poll would start and when the component is unmounted the poll would be cleared.

yaro90:
Another observation I've had which may be contributing to my slow response time is that if you have many queues, like in my case lets say 10. Every time I visit a different queue, it gets added to the update poll, so the poll grows in size as we're polling more and more queues in the same request. I believe only the queue that is being currently viewed is what should be polled for updates.

felixmosh:
I'll more than happy to review a PR which will solve it.
Instead of using multiple `setTimeout` we can add a check of the "activeQueue" and ignore response of a different one.

yaro90:
[...]

PR #1247, opened 2026-07-02
feat(ui): migrate queue/job data layer to React Query

Stosiu:
Switching status tabs or queues has no loading state, so you click and nothing visibly happens. Sometimes the previous tab's jobs even flash under the new header for a moment before the real data lands. That's the race from #348. Under the hood each view polls through a couple of hand-rolled zustand stores and a shared `useInterval`.

This moves that server state to React Query.

- Switching status, queue, or job now shows a loader instead of stale content, and the sidebar stays put while the new data loads. The old tab's jobs can't leak into the new one anymore: each view is its own query key, so the outdated request is dropped.
- `useQueues`, `useJob`, `useJobFlow`, plus `useMetrics` and the Redis stats modal, move to `useQuery`. Polling becomes `refetchInterval` on the existing interval setting. That removes both server stores, the `useInterval` race guard, `pollQueues`/`pollJob`, and the manual loading flags. zustand stays for client state (settings, menu, search, confirm).
- Query keys live in one small factory, and the `status` param is parsed to a real type instead of cast.
- `packages/ui` gets its first tests: a Jest + Testing Library setup covering the loading and transition behavior, running as its own CI job.

I went with React Query because it's easier to manage and most people already know it, so it lowers the barrier to working in here. You're not learning a bespoke polling setup before you can change anything, and caching, dedup and background refetch come from the library instead of being ours to maintain.

If you'd rather keep the current setup, I'm happy to drop the library and just fix the loading indicators and the confusing UX on what's already there. It's also a biggish diff, so I can split or trim it however you prefer.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -123,3 +123,32 @@ jobs:
 
       - name: Test Bun adapter
         run: yarn workspace @bull-board/bun test
+
+  test-ui:
+    name: Test UI
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v6
+
+      - name: Use Node 22.x
+        uses: actions/setup-node@v6
+        with:
+          node-version: 22.x
+          cache: 'yarn'
+
+      - name: Install deps
+        run: yarn install --immutable
+        env:
+          CI: true
+
+      - name: Build
+        run: yarn build
+        env:
+          CI: true
+
+      - name: Test UI
+        run: yarn workspace @bull-board/ui test
+        env:
+          CI: true
--- a/packages/ui/src/hooks/useJob.spec.tsx
+++ b/packages/ui/src/hooks/useJob.spec.tsx
@@ -1,0 +1,94 @@
+import type { AppJob } from '@bull-board/api/typings/app';
+import type { GetJobResponse } from '@bull-board/api/typings/responses';
+import { act, renderHook, waitFor } from '@testing-library/react';
+import { createMemoryHistory } from 'history';
+import { createWrapper, deferred, Deferred } from '../test/testUtils';
+import { useJob } from './useJob';
+import { useSettingsStore } from './useSettings';
+
+jest.mock('react-i18next', () => ({
+  useTranslation: () => ({ t: (key: string) => key }),
+}));
+
+function makeJobResponse(id: string): GetJobResponse {
+  return { job: { id, name: 'process' } as AppJob, status: 'completed' };
+}
+
+beforeEach(() => {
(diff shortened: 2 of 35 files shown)


felixmosh:
I'll review it ASAP, just need a bit more time :]

packages/ui/tests/hooks/useJob.spec.tsx
```diff
@@ -0,0 +1,94 @@
+import type { AppJob } from '@bull-board/api/typings/app';
```
felixmosh, line 1: Move all specs to tests folder outside of src... 🙏🏼

packages/ui/src/hooks/useJob.spec.tsx
```diff
@@ -0,0 +1,94 @@
+import type { AppJob } from '@bull-board/api/typings/app';
+import type { GetJobResponse } from '@bull-board/api/typings/responses';
+import { act, renderHook, waitFor } from '@testing-library/react';
+import { createMemoryHistory } from 'history';
+import { createWrapper, deferred, Deferred } from '../test/testUtils';
+import { useJob } from './useJob';
+import { useSettingsStore } from './useSettings';
+
+jest.mock('react-i18next', () => ({
```
felixmosh: You don't need a mock for it, you just can call
```
i18next.changeLanguage('cimode') // -> returns a Promise
```
https://www.i18next.com/overview/api#changelanguage

felixmosh (changes requested):
Thank you, 
it looks really good :]

felixmosh:
BTW,
Screenshot 2026-07-03 at 21-57-11 —

felixmosh (approved):

.github/workflows/nodejs.yml
```diff
@@ -123,3 +123,32 @@ jobs:
 
       - name: Test Bun adapter
         run: yarn workspace @bull-board/bun test
+
+  test-ui:
```
felixmosh, line 127: What is the benefit of separating ui tests into sub-workflow?

Stosiu: It mostly mirrors the existing `test-bun` split. The UI tests are jsdom/Jest, so they don't touch Redis and aren't Node-version dependent. Folding them into the `build` matrix would run them three times (20/22/24) against a Redis service they never use, and `yarn test` at the root already excludes `@bull-board/ui` (same as `bun`), so they need their own home regardless.

The downside is the duplicated checkout + install + build. If you'd rather keep the job count down, I can drop the separate job and run the UI tests as a single step inside `build`, guarded to one leg of the matrix (`if: matrix.node == '22.x'`), which reuses the build that job already does. Happy to go that way if you prefer it.

PR #1249, opened 2026-07-02
docs: refresh README screenshots to current UI

Stosiu:
Every package README hotlinks its screenshots from `master`, so they show up on the npm package pages too. They were a couple of majors old, from before the grouped overview, collapsible sidebar and environment badge landed. The two images were also different sizes (2034×1310 and 2672×1527), so they looked slightly off stacked together in the READMEs.

This recaptures both from the current UI (v8.1.0): the grouped overview and a queue detail view. The paths stay the same, so the package pages pick up the new images on their own. Both are 2880×1800 now. Docs-only change.


felixmosh:
Will you add a metrics enabled queue? (it looks cooler :]) (you can use your one of the queues at example.ts)

Stosiu:
@felixmosh done :)

felixmosh (approved):

PR #1253, opened 2026-07-06
docs: harden recipes and fix queue-adapter option docs

Stosiu:
Docs-only hardening pass. Everything is under `website/docs`, no library code changes. Closes some gaps against the actual API and covers the questions that keep coming up in issues.

Four new pages:

- Manage queues at runtime. Documents the `addQueue`, `removeQueue`, `setQueues` and `replaceQueues` functions that `createBullBoard` returns. Only `addQueue` had a passing mention before, and the difference between `setQueues` (upsert) and `replaceQueues` (full sync) wasn't written down anywhere.
- Alerting on failed jobs. bull-board is a viewer, not a monitor, so "how do I get alerted when a job fails?" comes up a lot. Shows the BullMQ `worker.on('failed')` retries-exhausted pattern, `QueueEvents` for cross-process alerting, and the dead-letter-queue approach, with pointers to the BullMQ docs for the event details.
- Troubleshooting. Collects the most common failure, asset and API 404s from a `basePath` that doesn't match the mount path, including behind a reverse proxy. Also the bundler `uiBasePath` case (links the Next.js recipe) and version mismatch.
- Set up with an AI agent. A copy-paste prompt for coding agents that bakes in the base-path gotcha and the "add auth before exposing" reminder, and points agents at the generated `llms.txt` / `llms-full.txt` so they work from the current API instead of stale memory.

A few corrections to existing pages:

- `allowCompletedRetries` was a real queue-adapter option missing from every options table. Added it (BullMQ, Bull, shared), noting Bull forces it off, plus a line in the read-only recipe.
- `externalJobUrl` now appears in the options tables. It had a recipe but wasn't listed as an option.
- The BullMQ flow-tree section showed passing `flowProducer` to `createBullBoard`, which the current signature ignores. Rewrote it: the flow tree is wired automatically from a registered BullMQ adapter's Redis connection, and only spans registered queues.
- Added an ioredis note to Installation about `maxRetriesPerRequest: null` for shared connections, a frequent setup snag.

New pages are linked from the sidebar, the recipes index and the production checklist. Built locally with `yarn docs:build`; the dead-link check passes.

### DIFF
--- a/website/docs/configuration/production-checklist.md
+++ b/website/docs/configuration/production-checklist.md
@@ -43,3 +43,7 @@ The dashboard talks to your Bull / BullMQ instances. Mismatched versions across 
 ## Logs
 
 Workers that call `job.log()` will have their lines visible in the dashboard under each job's Logs tab. If you're wondering "what did this job do before it failed", that's the answer. See [Job logs and flows](/recipes/job-logs-and-flows).
+
+## Alerting
+
+The dashboard won't page you. It only shows state while a tab is open. Wire failure alerts to BullMQ's own events, not to bull-board. See [Alerting on failed jobs](/recipes/alerting).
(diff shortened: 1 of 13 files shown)


felixmosh (approved):
Thank you 🙏🏼

PR #1255, opened 2026-07-06
feat(ui): type-check i18n keys and enforce locale completeness in CI

Stosiu:
The recent `fix: add missing queue.info messages` (ec64657) patched a set of translation keys that were used in the UI but never defined. It was found and fixed by hand, and nothing stops the same thing from happening again. This PR makes translation coverage checkable instead of relying on someone spotting raw keys in the running UI.

There are two independent guards, because they catch different classes of problem.

The first is types. `t()` is now typed against the primary `en-US` resource via `src/i18next.d.ts`, so any key used in a component that does not exist in `en-US` is a compile error, and keys autocomplete. A new `typecheck` script enforces this in CI (`tsc -p tsconfig.typecheck.json`, wired into the Test UI job). The production build only type-checks `.ts` files, so this config is what covers `.tsx`; it mirrors the build's automatic JSX runtime and relaxes `noUnusedLocals`, since many components still carry a now-unused `React` import from the classic runtime that is unrelated to key safety. A few call sites build keys from a runtime value (a status, tab, or action name) and cannot be verified statically, so they route through a small `dynamicTranslationKey` helper that keeps the escape hatch explicit and greppable rather than scattering casts.

Turning the typing on immediately surfaced two keys that the queue.info fix missed: `MENU.EXPAND_SIDEBAR` and `MENU.COLLAPSE_SIDEBAR`, used by `SidebarToggle` but present in no locale. They are added to `en-US` and propagated with `sync:locales`.

The second guard is a completeness test (`tests/i18n.spec.ts`). It asserts that every locale folder is registered in the `languages` list and vice versa, and that `sync:locales` produces no drift, so adding a key to `en-US` without syncing fails CI. This is the part that catches missing translations across locales, which typing cannot. It also flagged that `da-DK` shipped real, partly translated Danish but was never registered in `languages`, so it never loaded; it is registered now.

Non-English locales carry English placeholders for the two new keys, to be translated through the usual flow.

### DIFF
--- a/packages/ui/src/i18next.d.ts
+++ b/packages/ui/src/i18next.d.ts
@@ -1,0 +1,17 @@
+import 'i18next';
+import type messages from './static/locales/en-US/messages.json';
+
+// Makes `t()` type-checked against the primary locale (en-US). Unknown or
+// misspelled keys become compile errors, and keys autocomplete. Other locales
+// are kept in parity with en-US by the i18n completeness test, not by types.
+//
+// Kept under src/ (not typings/) so it is not published to npm — a global
+// i18next augmentation must not leak into consumers of @bull-board/ui.
+declare module 'i18next' {
+  interface CustomTypeOptions {
+    defaultNS: 'messages';
+    resources: {
+      messages: typeof messages;
+    };
+  }
+}
--- a/packages/ui/tests/i18n.spec.ts
+++ b/packages/ui/tests/i18n.spec.ts
@@ -4,18 +4,23 @@ import path from 'path';
 import { syncLocales } from 'i18next-locales-sync';
 import { languages } from '../src/constants/languages';
 
-const localesDir = path.resolve(__dirname, '../src/static/locales');
-const primaryLanguage = 'en-US';
+// Reuse the same config `yarn sync:locales` runs on, so the test can never
+// drift from the real sync behaviour.
+const syncConfig = require('../localesSync.config.js') as {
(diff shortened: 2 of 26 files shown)


felixmosh:
Great idea, thanx.

packages/ui/src/i18next.d.ts
```diff
@@ -0,0 +1,17 @@
+import 'i18next';
```
felixmosh: Move this d.ts to `/typings` folder, pls

packages/ui/tests/i18n.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  // Guards against adding/removing a key in en-US without running `yarn sync:locales`.
+  // Re-runs the sync into a throwaway folder and asserts it produces no changes,
+  // which is exactly what a synced tree looks like (plurals expanded per CLDR).
+  it('keeps every locale in sync with en-US (run `yarn sync:locales`)', () => {
+    const outputFolder = fs.mkdtempSync(path.join(os.tmpdir(), 'bb-locales-'));
+    try {
+      const secondaryLanguages = localeFolders.filter((lng) => lng !== primaryLanguage).sort();
+
+      syncLocales({
```
felixmosh, line 38: it is better to reuse real config, `localesSync.config.js`

felixmosh (changes requested):

felixmosh (approved):

Issue #1254, opened 2026-07-06
Feature: long-retention historical job metrics (opt-in)


Stosiu:
## Motivation

The metrics we ship today (#1208) are a stateless proxy over BullMQ's `getMetrics`, which is a per-minute ring buffer: opt-in per worker, per-queue, and bounded by `maxDataPoints`. There is no way to ask "how many jobs completed and failed per day over the last 90 days," and no cross-queue global view. This is the recurring ask in #258 and #618, and the Sidekiq/Horizon-style history a lot of people expect from a queue dashboard.

I would like to add long-retention history (default 90 days) of completed/failed counts, per queue and as a global rollup, reusing the existing `QueueMetrics` charts.

## Design goal: keep the core stateless

I know the core API is meant to serve only the UI, and that bull-board holds no server-side state (#759). This proposal is built around that boundary, not against it:

- All stateful behavior (recording, storage, retention) lives in a new opt-in package, `@bull-board/metrics`. `packages/api` gains no storage, no scheduler, no state.
- The core gets exactly one small, optional read-seam whose only purpose is to feed the existing UI charts.
- If a user does not opt in, nothing changes and there is zero cost.

## Approach: extend the metrics we already have

Rather than count events ourselves, snapshot the native metrics we already read. Workers with `metrics` enabled already keep an exactly-aggregated per-minute count in Redis. A small recorder in the user's own process periodically copies the new minutes out of that ring buffer into a long-retention store before they roll off.

Because the source is keyed by minute timestamp, the store is written idempotently by timestamp. Running the recorder in several processes converges to the same values instead of double-counting, so no singleton or leader election is needed. Precondition: workers must have native metrics enabled, since we are extending the retention of that same data.

## Pieces

Each piece below notes why it is shaped this way rather than the obvious alternative, since that is where I would most like your feedback.

### 1. `@bull-board/metrics` (new package), write side

Runs in the user's app/worker process:
```ts
new MetricsRecorder({ queues, connection, retentionDays: 90 }).start();
```
It snapshots `getMetrics('completed'|'failed')` into namespaced, TTL'd Redis keys (daily rollups, per queue plus a global rollup).

- Why a separate package, not core: recording is inherently stateful (it owns Redis keys, a timer, and a retention policy). Putting it in `packages/api` would break the "core serves only the UI, holds no state" boundary. A separate opt-in package keeps that boundary intact and costs nothing for users who do not enable it.
[...]

PR #1256, opened 2026-07-09
feat: opt-in historical job metrics (beta)

Stosiu:
Implements the long-retention historical metrics proposed in #1254. Shipping as **beta**: it is opt-in and safe to run (it only writes its own namespaced Redis keys), but the API and storage layout may still change in a minor release while it settles.

BullMQ's native metrics are a per-minute ring buffer, capped at `maxDataPoints`, per queue, and only kept for the buffer window. There is no way to ask "how many jobs completed and failed per day over the last 90 days", and no cross-queue total. This adds that history as an opt-in feature, reusing the existing charts, without adding any server-side state to the core.

## What it does

The core `@bull-board/api` gains a single optional read seam. `createBullBoard` accepts a `historyProvider`, and when one is present it registers one endpoint, `GET /api/metrics/history`, whose handler is a closure bound to that provider. No server adapter is touched, and with no provider configured the endpoint is absent and nothing changes. The core stores nothing and knows nothing about Redis keys; it only defines the `MetricsHistoryProvider` interface and delegates. A `hasHistoryProvider` UI flag is set so the UI can light up the history views, and the new page path is added to the SPA entry routes so it is served on a direct load or refresh.

A new opt-in package `@bull-board/metrics` holds all the state. `MetricsRecorder` runs in your always-on worker or app process and periodically snapshots BullMQ's native completed and failed per-minute metrics into long-retention Redis buckets before they roll off the ring buffer. Writes are idempotent by minute through a single atomic Lua upsert, so running the recorder in several processes, or restarting it, converges instead of double-counting. There is no singleton or leader election. It keeps daily rollups per queue plus a cross-queue global rollup, all UTC-aligned and expired by a configurable retention (90 days by default). `RedisMetricsHistoryProvider` reads those buckets back and implements the core interface. The one precondition, documented in the recipe, is that your BullMQ workers must have native metrics enabled (`metrics: { maxDataPoints }`), since there is nothing to snapshot otherwise. This iteration is BullMQ only.

In the UI, each queue's metrics chart gains a 60m / 7d / 30d / 90d range selector (60m stays the live native view, the longer ranges read the history), and can be collapsed with the preference remembered, since you often do not need it. A dedicated "Metrics history" page plus sidebar item shows the cross-queue global throughput with a per-queue breakdown table (a count and a proportional bar per queue), so you can see at a glance how much each queue processed over the window. Everything is gated on `hasHistoryProvider`, so with the feature off the UI is unchanged. The recharts markup was extracted into a shared `ThroughputAreaChart` used by all the chart surfaces.

## Screenshots

The dedicated history page: global throughput over the selected range plus a per-queue breakdown.

The Metrics history page: global completed and failed throughput chart with a per-queue breakdown table

The per-queue chart with the range selector (60m native, 7d/30d/90d from history):

A queue's metrics chart with the 60m / 7d / 30d / 90d range selector

## Testing

Everything is covered by tests running against a real Redis and real BullMQ workers. The `@bull-board/metrics` suite covers the getMetrics-to-minute mapping, the idempotent storage and rollups (including reprocessing without double-counting and the cross-queue global rollup), the recorder snapshotting completed and failed metrics, the provider read paths for day and hour granularity, and a recorder-to-provider end-to-end round trip. The core suite covers the seam (endpoint present only with a provider, input validation, delegation) and the SPA entry route. The UI has hook and page tests. The dev example (`example.ts`) wires the recorder and provider so the feature can be exercised locally, and the docs ship a recipe with the screenshots above.

## Could be split

This is one PR for reviewability of the whole feature, but it is designed to split cleanly into smaller PRs if you prefer to land it in stages: (1) the core `historyProvider` seam and endpoint in `@bull-board/api`, (2) the `@bull-board/metrics` package, (3) the UI (range selector, history page, per-queue table), and (4) the docs. Parts 1 and 2 are independently useful (a provider can be consumed without the UI), and the UI degrades to unchanged when the flag is off. Happy to break it up whichever way is easiest to review.


felixmosh:
Let me some time to review it, 🙏🏼

kfir-novella:
Very nice. I've been waiting for this for a while. Thank you. Hope it will be merged soon. 🙏🏼

packages/api/src/index.ts
```diff
@@ -29,11 +39,12 @@ export function createBullBoard({
         default: 'static/images/logo.svg',
         alternative: 'static/favicon-32x32.png',
       },
+      hasHistoryProvider: Boolean(options.historyProvider),
```
felixmosh: what will happend if there is history provider but the worker didn't enable metrics?

packages/metrics/src/MetricsRecorder.ts
```diff
@@ -0,0 +1,106 @@
...
+      this.running = false;
+    }
+  }
+
+  private async snapshotOne(
+    adapter: BaseAdapter,
+    name: string,
+    metric: MetricsType
+  ): Promise<void> {
+    const cursorKey = `${name}:${metric}`;
```
felixmosh, line 129: can u explain in words how it works?

felixmosh (commented):
looks good to me, just few clearfications...

packages/api/src/index.ts
Stosiu: Nothing breaks. `hasHistoryProvider` only tells the UI that a history endpoint exists, it doesn't promise there's data behind it. If a worker never turned metrics on, `getMetrics()` comes back empty (`data: []`), so `metricsToMinutePoints` yields nothing and the recorder stores nothing. The chart just renders empty, the same way the live metrics view already behaves when metrics are off.

There's no way to know at `createBullBoard` time whether any worker has metrics enabled without going to Redis, so I kept the flag as a plain "is a provider wired up" signal. The recipe docs already call the precondition out ("If metrics aren't enabled on the workers, `queue.getMetrics()` returns nothing, and the recorder has nothing to snapshot"), and I've now added a line there spelling out the empty-chart behavior too.

packages/metrics/src/MetricsRecorder.ts
Stosiu: Sure. It's an incremental sync, one queue+metric at a time, run every tick (60s by default).

BullMQ keeps a rolling per-minute ring buffer and hands it back newest-first from `getMetrics()`, along with `meta.prevTS`. `metricsToMinutePoints` turns that into `{ minute, value }` pairs where `minute` is an absolute index (`floor(ts / 60000)`). BullMQ never includes the in-progress current minute, so every point is a finished minute that won't change again, which is what makes it safe to persist.

`seenUpTo` (kept in memory, keyed by `queue:metric`) is a watermark: the newest minute already written. Each tick walks the points newest-first and stops the moment it hits a minute that's already stored, because everything past that point is older and stored too. Anything newer gets written with `upsertMinute` (an upsert, so overlapping windows across ticks never double-count), then the watermark moves up to the newest minute seen.

So the first tick fills in whatever the buffer holds, and every tick after that only writes the handful of minutes that showed up since the last one.

I've also added a short comment above `snapshotOne` capturing this, so it's documented in the code.

Stosiu:
Ran a self-review pass over the whole feature before this sits in front of you longer, and pushed a follow-up commit with one real fix plus the test gaps that let it hide.

The fix: `hasHistoryProvider` was spread into the UI config before `...options.uiConfig`, so a caller passing `uiConfig: { hasHistoryProvider: true }` with no provider could turn the history UI on while `/api/metrics/history` was never registered. The reverse could hide a working feature. The derived value now comes after the spread so it always wins. There's a test that pins it: caller `uiConfig` can't force the flag true without a provider.

Two endpoint test gaps went with it:

- a provider whose `getHistory` rejects now returns a structured 500 through the error handler, so a future refactor can't quietly turn it into an unhandled rejection
- `granularity: 'hour'` is asserted to reach the provider, so the accepted values can't silently regress to day-only

The two clarifications you asked about are in the same commit: the doc note that the chart renders empty when a provider is set but the worker never enabled metrics, and a comment above `snapshotOne` walking through the watermark sync.

I went over the metrics and ui sides too. Nothing else looked ship-blocking, but two things are worth flagging if you'd rather I close them now: the per-(queue, metric) totals hashes keep refreshing their TTL, so they never expire (slow Redis growth over time, not wrong data), and the history chart doesn't gap-fill idle days, so a sparse range spaces its points equidistantly. Happy to push fixes for either, or leave them for a follow-up.

felixmosh:
My biggest consern is that bloating redis with metric data... how we can make sure that it won't "bloat" the redis?

there is something that removes old values? (let's say we set it to 90 days), older metrics should be deleted, there is something that handles it?

felixmosh:
@Stosiu, can you point me into the code that shows that the metrics data retained only for the time the user requested it (90 days max) - entry points older than that should be removed.

Will you fix the test & resolve conflicts?

Stosiu:
Rebased on master, conflicts resolved, tests fixed, CI green locally. Also deleted 2.6k lines of working notes that I'd committed by mistake, which is most of why the diff looked so alarming.

## Where retention is enforced

Two mechanisms, both in [`HistoryStore.ts`](https://github.com/felixmosh/bull-board/blob/feat/historical-job-metrics/packages/metrics/src/HistoryStore.ts):

**Day-scoped keys expire on their own.** [Lines 76-81](https://github.com/felixmosh/bull-board/blob/feat/historical-job-metrics/packages/metrics/src/HistoryStore.ts#L76-L81) set an `EXPIRE` on every write. The TTL is only refreshed while that day is still being written, so a key dies its retention window after the day it covers. Nothing to clean up by hand.

**The daily totals hash is trimmed.** This is the gap you were pointing at. That hash gets written *every* day, so its TTL kept rolling forward and it never expired, growing one entry per day forever. It's now trimmed at [lines 83-84](https://github.com/felixmosh/bull-board/blob/feat/historical-job-metrics/packages/metrics/src/HistoryStore.ts#L83-L84) via the `trim` function at [line 46](https://github.com/felixmosh/bull-board/blob/feat/historical-job-metrics/packages/metrics/src/HistoryStore.ts#L46), which drops entries older than the cutoff. It runs only on the first write of a new day, so about once a day per queue, and day keys are ISO dates so the comparison is a plain string compare.

Covered by `gives each tier its own TTL` and `drops day totals older than the retention window` in `HistoryStore.spec.ts`.

## The bigger answer

Retention existed, but I measured it rather than trusting it, and the default was genuinely too fat: a busy queue cost ~15 MB at 90 days. So snapshots now go to three resolutions at once, each with its own retention:

| Tier | Default | Per busy day, per queue and metric |
| --- | --- | --- |
| Minute | 7 days | ~72 KB |
| Hour | 90 days | ~0.3 KB |
| Day (totals) | 90 days | ~15 bytes |

The hourly rollup is ~240x cheaper than the minute detail for the same day, and the charts only ever read the daily totals. So minute detail defaults to 7 days instead of 90 and **~15 MB becomes ~1.1 MB per queue** with no visible change. Idle queues still cost nothing, since a zero minute is never written. Configurable per tier via `retention: { minutes, hours, days }`.

All three tiers come from the same delta in the same atomic script, so they can't drift and there's no compaction job to schedule. One catch worth knowing: the minute window doubles as the recorder's catch-up window after downtime, and shortening it nearly introduced a double-count, so the recorder now refuses to write past it. There's a regression test for exactly that.

## Also in this push

- **Storage modal**, reached from the actions menu on the Metrics history page since it's occasional maintenance rather than daily reading. Footprint by tier, per-queue breakdown, day range. Fetched only when opened, never polled.

Actions menu with the Storage entry

Storage modal

- **Trim-to-range and clear-all**, each behind a confirmation naming the cutoff or the total size. Hidden when every queue is `readOnlyMode`.

Confirmation before clearing history

- Two endpoints for the above, registered only when the provider implements the matching optional method, so a custom read-only provider stays valid.
- **Bug fix**: the per-queue chart bailed out to "no metrics collected" whenever BullMQ's native buffer was empty, which hid the range selector and made that queue's recorded history unreachable from its own page after a worker restart. An existing test was asserting the broken behaviour, so that was corrected too. The selector is reachable again either way:

Per-queue chart with the range selector

- `stats()` measures in pipelined batches instead of a round trip per key: 4,706 keys went from 1057ms to 40ms locally.
- Redis-backed specs pinned to a scratch logical DB. They were clearing by key pattern and writing into the shared `__global__` rollup, which wiped a local dev board's history whenever you ran `yarn test`.

Tests are at 88 for metrics, 134 for api, 59 for ui. `footprint.spec.ts` writes real data through the real write path and asserts the sizing numbers above rather than taking my word for them. I also ran it against a real board end to end: seeded several days, trimmed, cleared, and confirmed the minute, hour and day tiers agree exactly.

Happy to split the storage modal out if you'd rather keep this PR to the recorder.

felixmosh (approved):

felixmosh:
@Stosiu WOW, thank you... like always, it's a real pleasure to review your PR ;]

felixmosh:
I've made some refactors here.
1. making 2 API calls always doesn't make any sense, just return both of the values (completed + failed).
2. QueueMetrics - conditions were too much for me. + some bugs, when the range in history value, the getMetrics still fetched. - I've split it into smaller contained components.

felixmosh:
Released in `v8.2.0`

PR #1264, opened 2026-07-17
feat: per-queue job data schema for the add-job form

### LINKED ISSUE #1263: Define Input Data for Queue and default input data for the add job panel
This is a QOL feature, but I think that in addition to a description it would be really nice if we could also set something like '{make:string,model:string}'
then see that in the fancy new modal with overview and description 

adding to that it would save me alot time when i need to manually add jobs, if the input data was prefilled :

{
  "make":"",
  "model":"" 
}

Then all i would have to is define that once per queue and then when i go to add a job i don't need to remember what properties it needs, and i don't have to type all the JSON I can just fill in the missing info.

again just a nice to have and if you like the idea, and i get some free time at work, I may try to implement this myself if it is even something you are interested in adding as a feature

Stosiu:
Closes #1263.

Queues can now declare an optional JSON Schema for their job `data`, and the add-job form puts it to work. The motivation is the one from the issue: when you manually enqueue a job on a queue you did not write, you should not have to go back to the code to remember what properties it needs.

A single `jobDataSchema` option on any queue adapter drives three things in the **Add job** form:

- **Prefill.** The job data editor opens with a starting value instead of an empty object. It uses the schema's `default`, otherwise its first `examples` entry, otherwise a template generated from the schema by `json-schema-library`, the same validator the editor already pulls in for linting. The template walks nested objects and arrays and seeds each leaf with its own `default` or a typed placeholder, so a schema that only lists `properties` still gives you a ready-to-fill template.
- **Autocomplete.** The editor already speaks JSON Schema for job options; wiring the schema into the job data field gives the same key autocomplete, with any `description` shown on hover.
- **Validation.** The JSON is linted against the schema inline, flagging missing required fields, wrong types and unknown keys before you submit.

The schema is author-supplied config on the adapter, exactly like `description` and `displayName`. It lives in the dashboard only and is never read from Redis, so it documents the contract for the person adding a job rather than enforcing anything at the worker. That tradeoff is called out in the docs.

The dashboard loads the schema from a dedicated `GET /api/queues/:queueName/job-data-schema` endpoint when the Add job modal opens, instead of attaching it to the queues list that the UI polls every 15 seconds. Schemas can get large, and that polled payload is already heavy, so the schema is only fetched when someone actually opens the form. This mirrors the existing `default-job-options` endpoint.

While wiring the schema into the data field, the `JsonEditor` was changed from a fixed 200px box to one that grows with its content between a min and a max height and then scrolls. Small payloads like an empty job-options object no longer sit in a large empty box, and a prefilled schema sizes naturally.

Everything is opt-in and additive. Queues without a `jobDataSchema` behave exactly as before.

### Testing

- `packages/api`: contract tests for the new `job-data-schema` endpoint (returns the configured schema, an empty object when none is set) plus a check that the schema stays out of the polled `GET /api/queues` payload.
- `packages/ui`: unit tests for the prefill derivation (`default` vs `examples` vs properties skeleton, per-property defaults).
- Full UI suite and the API queues suite pass.

The `with-express` example configures a schema on the Reset Password queue for manual verification.

### DIFF
--- a/packages/api/src/handlers/queues.ts
+++ b/packages/api/src/handlers/queues.ts
@@ -87,7 +87,6 @@ async function getAppQueues(
         name: queueName,
         displayName: queue.getDisplayName() || undefined,
         description: queue.getDescription() || undefined,
-        jobDataSchema: queue.getJobDataSchema(),
         statuses: queue.getStatuses(),
         counts: counts as Record<Status, number>,
         jobs: jobs.filter(Boolean).map((job) => formatJob(job, queue)),
--- a/packages/api/src/handlers/queues.ts
+++ b/packages/api/src/handlers/queues.ts
@@ -87,6 +87,7 @@ async function getAppQueues(
         name: queueName,
         displayName: queue.getDisplayName() || undefined,
         description: queue.getDescription() || undefined,
+        jobDataSchema: queue.getJobDataSchema(),
         statuses: queue.getStatuses(),
         counts: counts as Record<Status, number>,
         jobs: jobs.filter(Boolean).map((job) => formatJob(job, queue)),
--- a/packages/ui/src/hooks/useQueueJobDataSchema.ts
+++ b/packages/ui/src/hooks/useQueueJobDataSchema.ts
@@ -1,35 +1,17 @@
-import { GetQueueJobDataSchemaResponse } from '@bull-board/api/typings/responses';
-import { useEffect, useState } from 'react';
+import { useQuery } from '@tanstack/react-query';
+import { queryKeys } from './queryKeys';
 import { useApi } from './useApi';
 
 export function useQueueJobDataSchema(queueName: string | null, enabled: boolean) {
   const api = useApi();
-  const [jobDataSchema, setJobDataSchema] = useState<GetQueueJobDataSchemaResponse | null>(null);
-  const [loading, setLoading] = useState(false);
 
-  useEffect(() => {
-    if (!enabled || !queueName) return;
+  const { data, isPending } = useQuery({
+    queryKey: queryKeys.jobDataSchema(queueName),
+    queryFn: () => api.getQueueJobDataSchema(queueName as string),
+    enabled: enabled && !!queueName,
+    // Author-supplied config, it can't change while the dashboard is open.
+    staleTime: Infinity,
+  });
 
-    let cancelled = false;
-    setLoading(true);
-    setJobDataSchema(null);
-    api
-      .getQueueJobDataSchema(queueName)
-      .then((data) => {
-        if (!cancelled) setJobDataSchema(data);
-      })
-      .catch((error) => {
-        // eslint-disable-next-line no-console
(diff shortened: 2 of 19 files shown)


mwalkersigma:
Oh this is way better then what I was doing lol

mwalkersigma:
@Stosiu Is there any faster way, to see changes when running the dev command in the repo ? Ts-node-dev never seems to hot reload the page so i have to restart and wait for it to build

felixmosh (approved):
Looks great, thank you :]

can you fix the formatting error?

mwalkersigma:
@felixmosh when I'm running the development command and save a file is there any faster way to see change ? 

I'm ctrl + c then re running the build and dev command

felixmosh:
> @felixmosh when I'm running the development command and save a file is there any faster way to see change ?
> 
> I'm ctrl + c then re running the build and dev command

Yarn dev, should hot reload server & ui

packages/api/src/handlers/queues.ts
```diff
@@ -87,6 +87,7 @@ async function getAppQueues(
         name: queueName,
         displayName: queue.getDisplayName() || undefined,
         description: queue.getDescription() || undefined,
+        jobDataSchema: queue.getJobDataSchema(),
```
felixmosh: since this data is relevant only on addJobModal, I think that it is really wastfull to put it here, jsonSchemas tend to be quite large.
This endpoint is polled every 15 sec....

Let's create a new endpoint that will be loaded when the modal is opened.

WDYT?

felixmosh (changes requested):

packages/api/src/handlers/queues.ts
Stosiu: Done. The schema is off the polled queues list now: it moved to a dedicated `GET /api/queues/:queueName/job-data-schema` endpoint that the Add job form fetches when it opens, following the same pattern as `default-job-options`. Also fixed the formatting error from the earlier review.

packages/ui/src/hooks/useQueueJobDataSchema.ts
```diff
@@ -0,0 +1,35 @@
+import { GetQueueJobDataSchemaResponse } from '@bull-board/api/typings/responses';
+import { useEffect, useState } from 'react';
+import { useApi } from './useApi';
+
+export function useQueueJobDataSchema(queueName: string | null, enabled: boolean) {
```
felixmosh, line 5: why you didn't used `react-query` here?

mwalkersigma: @Stosiu Should be pretty simple swap to move this to useQuery

Stosiu: my bad @felixmosh fixed it (fixed the two other ones that I missed ealier too)

felixmosh (approved):
Thank you 🙏🏼

PR #1266, opened 2026-07-19
fix(api): declare bull queue libraries as optional peers

### LINKED ISSUE #1265: @bull-board/api does not declare bullmq as a dependency or peerDependency
\`@bull-board/api\`'s \`bullMQAdapter\` requires \`bullmq\` at runtime, but \`bullmq\`
only appears in \`@bull-board/api\`'s \`devDependencies\` — not in
\`dependencies\` or \`peerDependencies\`:

    "devDependencies": {
      ...
      "bullmq": "^5.79.2",
      ...
    },
    "peerDependencies": {
      "@bull-board/ui": "8.1.2"
    }

This works today only because most consumers already depend on \`bullmq\`
directly, and npm/yarn's flat \`node_modules\` happens to hoist it somewhere
resolvable. Under pnpm's isolated \`node_modules\` layout, resolution is not
guaranteed — bumping the host project's own \`bullmq\` version (even a patch,
e.g. 5.80.6 -> 5.80.8) can shift where pnpm's virtual store places the
package and break \`@bull-board/api\`'s undeclared \`require('bullmq')\`,
producing \`Cannot find module 'bullmq'\` at runtime with no build-time
warning.

**Repro (pnpm workspace):**
1. Install \`@bull-board/api@8.1.2\` and \`bullmq@5.80.6\` in a pnpm workspace.
2. Confirm resolution succeeds:
   \`node -e "console.log(require.resolve('bullmq', {paths:['<path to @bull-board/api's install dir>']}))"\`
3. Bump only \`bullmq\` to \`5.80.8\` (\`pnpm add bullmq@5.80.8\`).
4. Any code path exercising \`@bull-board/api\`'s BullMQ adapter throws
   \`Cannot find module 'bullmq'\`.

**Suggested fix:** declare \`bullmq\` as an (optional) peerDependency —
mirroring how \`@bull-board/ui\` is already a peerDependency — so package
managers can verify/warn on version compatibility instead of silently
relying on hoisting.

alex-morozov84:
Closes #1265.

`@bull-board/api` imports `bull` / `bullmq` from its queue adapters at runtime,
but currently only declares them as dev dependencies. Under package managers with
strict or isolated dependency layouts, this can fail with `Cannot find module`
when the consumer uses the Bull or BullMQ adapter.

This declares both queue libraries as optional peer dependencies, since consumers
normally use either Bull or BullMQ, not necessarily both.

The peer ranges match the versions currently used by this package's own
devDependencies/tests.

Checks:
- yarn lint:check
- yarn format:check
- yarn test

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -57,6 +57,16 @@
     "ts-jest": "^29.4.11"
   },
   "peerDependencies": {
-    "@bull-board/ui": "8.1.2"
+    "@bull-board/ui": "8.1.2",
+    "bull": "^4.16.5",
+    "bullmq": "^5.79.2"
+  },
+  "peerDependenciesMeta": {
+    "bull": {
+      "optional": true
+    },
+    "bullmq": {
+      "optional": true
+    }
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -480,6 +480,13 @@ __metadata:
     ts-jest: "npm:^29.4.11"
   peerDependencies:
     "@bull-board/ui": 8.1.2
+    bull: ^4.16.5
+    bullmq: ^5.79.2
+  peerDependenciesMeta:
+    bull:
+      optional: true
+    bullmq:
+      optional: true
   languageName: unknown
   linkType: soft
 


felixmosh:

PR #1278, opened 2026-07-23
feat(ui): retry all failed jobs from the overview screen

### LINKED ISSUE #1262: Feature request: Be able to retry all failed jobs from main queue screen / queue cards
Hi! I would love to be able to "retry all failed jobs". 

It would be great if I could retry all failed jobs from all queues from the global context dropdown:

Screenshot of main queue UI with a red circle around the global contextual actions dropdown that lets you pause all queues or sort them

And/or it would also be just as helpful to be able to retry all failed jobs for a specific queue from each queue's dropdown:

Screenshot of main queue UI with a red circle around the contextual actions dropdown for a specific queue that lets you add jobs, pause jobs, set the queue concurrency, empty the queue, or obliterate the queue

Thanks!

Stosiu:
Closes #1262.

Retrying failed jobs meant opening every queue one at a time and hitting "Retry all" on the failed tab. This adds the action to the three dropdowns on the overview screen, using the existing `PUT /api/queues/:name/retry/failed` endpoint. The cross-queue variants just loop over the queue list, as you suggested on the issue.

Each item carries the number of jobs it will act on, so you know the size of what you are about to trigger before the confirm dialog appears:

**Global dropdown**, the one from the first screenshot in the issue. Counts every writable queue that has failed jobs.

Global dropdown

**Queue card dropdown**, the second one from the issue.

Queue card dropdown

**Group dropdown**, added for consistency since the overview groups queues and already offers a group-level pause.

Group dropdown

The cross-queue confirm names both numbers:

Confirm dialog

## Feedback while it runs

A retry across fifteen queues takes a moment, and until now clicking the button looked like nothing happened. Retries now raise a pending toast that resolves into a success message, so the action always reports back.

Pending toast

Success toast

This also covers the existing "Retry all" button on the queue page, which goes through the same `retryAll` action. The vendored `toastify.css` was missing the spinner and toast-icon rules, so those are added.

## Behaviour

An item only renders when there is something to do: the queue is writable, its adapter allows retries, and it has at least one failed job. The cross-queue variants filter the same way, so a read-only queue or one with an empty failed set is never included in the request fan-out. When the API rejects a request it already raises its own error toast, so the pending toast is dismissed rather than reporting a false success.

## Tests

`packages/ui/tests/utils/failedRetries.spec.ts` covers the eligibility rules and the count rollup, `tests/utils/actionToast.spec.ts` covers the pending/success/failure paths including a partially failed batch, and `tests/components/RetryFailedActions.spec.tsx` opens both dropdowns and asserts the click reaches the action with the right queues. 76 tests pass in `packages/ui`.

## Translations

All new copy is translated across the eleven locales rather than left as English fallbacks.

The second commit is a small unrelated cleanup I noticed while auditing the locale files for untranslated strings: the metrics range chips (`60m`, `7d`, `30d`, `90d`) were still Latin abbreviations in de, fr, ja, ko and zh, `Peak / min` and `Peak hukommelsesforbrug` were untranslated in Danish, and pt-BR used `NA` instead of `N/D`. Happy to drop that commit if you would rather keep this PR to one thing.


packages/ui/src/toastify.css
```diff
@@ -120,6 +120,34 @@
...
+    animation: Toastify__spin 0.65s linear infinite;
+}
+
+@keyframes Toastify__spin {
+    from {
+        transform: rotate(0deg);
+    }
+    to {
+        transform: rotate(360deg);
+    }
```
felixmosh, line 150: Can you add a Dark mode colors to the toast?

felixmosh (approved):

felixmosh:
Thank you

bamadesigner:
Thank you!

PR #1280, opened 2026-07-23
refactor(ui): share the queue test fixture across specs

Stosiu:
Self-review follow-up to #1278, no behaviour change.

`makeQueue` had been copy-pasted into five specs, byte for byte the same each time, and #1278 added two more copies of it. It now lives in `tests/testUtils` alongside a `makeQueueWithFailed` variant for the specs that care about the failed count, which takes about 160 lines out of the suite.

Also drops two comments that only restated the code underneath them, and a defensive `queue.counts?.failed` in `canRetryFailedJobs` where `counts` is required on `AppQueue` and the function next to it already reads it directly.

The dead `toastify.css` rules and the toast dark mode colours are deliberately left out of this one, since #1281 replaces that layer wholesale.

### DIFF
--- a/packages/ui/src/utils/actionToast.ts
+++ b/packages/ui/src/utils/actionToast.ts
@@ -8,10 +8,6 @@ function hasError(result: unknown): boolean {
   return !!result && typeof result === 'object' && 'error' in result;
 }
 
-/**
- * Runs a queue action behind a pending toast so a long request never looks like a no-op,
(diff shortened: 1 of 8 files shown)


felixmosh (approved):

felixmosh:
can i merge it? :]

Stosiu:
@felixmosh yeah now it's good to go :)

PR #1281, opened 2026-07-23
refactor(ui): replace react-toastify with the base-ui toast

### LINKED ISSUE #1279: refactor(ui): Replace react-toastify with base-ui toast
UI package is using both of the libs, I don't see any benefit of adding additional dependecy.

With this refactor, we should review dark mode colors of the toast as well.

Stosiu:
Closes #1279.

The UI already depends on `@base-ui/react`, so `react-toastify` was a second library doing a job the first one already does. This moves the toast layer over and drops the dependency along with 282 lines of vendored CSS.

That vendored `packages/ui/src/toastify.css` was dead weight before this PR. It is a copy of the v9 stylesheet, but react-toastify 11 injects its own sheet at runtime, and since that sheet is appended later it wins every tie on specificity. None of the vendored rules were reaching the toasts. It is also the reason the toast never matched the dashboard: it rendered in the library's own light theme (`--toastify-color-light: #fff`, text `#757575`) whatever the app was set to, and nothing in our file could override it.

## What it does

`services/toastManager.ts` holds one `Toast.createToastManager()` instance, created outside React and handed to `Toast.Provider` as a prop. That indirection is the point: `Api` is a plain class and cannot call `useToastManager`, so its error interceptor needs to push onto the same manager the React tree renders from. `components/Toaster` renders the provider, portal, viewport and toast list.

The three states map onto the manager directly. `add({ type: 'loading', timeout: 0 })` raises the pending toast, `update(id, { type: 'success', timeout })` swaps it once the request settles, and `close(id)` drops it when the request failed, since `Api` has already raised its own error toast in that case. Type only drives an icon and is exposed as `data-type`, so every toast is the same surface with a different mark rather than a repainted panel.

Styling reuses the variables the cards and dropdowns already use (`--card-bg`, `--text-color`, `--dropdown-border-color`, `--completed`, `--failed`), so both themes come out of the existing tokens and there is no theme prop to keep in sync with the dark mode setting.

Two behaviour changes worth calling out. The viewport moved to the bottom right, because the top right sat on top of the overview action buttons, including the dropdown you just used to open the toast. Below 480px it goes full width with small gutters.

One gotcha for anyone touching the styles later: `Toast.Title` renders an `h2`, and the global heading reset in `index.css` normalises weight, margin and line height but not `font-size`, so the title needs an explicit `font-size: inherit` or it renders at the user agent's `1.5em`.

## Screenshots

All four show the pending state, which is the one that used to look like nothing had happened.

| | Desktop | Mobile |
|---|---|---|
| **Dark** |  |  |
| **Light** |  |  |

The mobile light shot has #1282 applied on top, otherwise the header wordmark is invisible and distracts from the toast. That fix is not part of this branch.

## Testing

`tests/components/Toaster.spec.tsx` renders the real component and drives the shared manager through the success, error and loading-then-success paths, asserting the `data-type` the styling hangs off. `tests/utils/actionToast.spec.ts` moves onto the new manager and keeps its coverage of the partially failed batch. 79 tests pass in `packages/ui`.

Checked by hand in the example app in both themes at 1440px and 390px, on the pending and success states, with the retry request held open so the loading toast stays up long enough to look at.

## Note

Not fixed here: the counts read "Retry 1 failed jobs" when exactly one job has failed. That is a pluralisation bug from #1278 and wants `_one` keys across the locales, plus nested plural fragments for the confirm text, since it interpolates two independent counts and i18next can only pluralise on one. Happy to send it separately.


felixmosh:
Man you are fast ⚡
How is looks on mobile?

Stosiu:
@felixmosh moved to draft to correct mobile issues and perfect styling, as icons in notifications were slightly misaligned. I'll mark it as ready for review once I've addressed these fixes

Stosiu:
Should be good to go @felixmosh

felixmosh (approved):
Thank you 🙏🏼

PR #1283, opened 2026-07-23
feat(demo): show historical metrics in the demo, document @bull-board/metrics

Stosiu:
The [live demo](https://felixmosh.github.io/bull-board/demo/) doesn't show the historical metrics feature at all, and `@bull-board/metrics` is missing from the readme, so someone landing on either has no way to discover it. This adds both.

Opening as a draft while I get a second opinion on how faithful the demo's fake data should be.

## Demo

The demo runs entirely in the browser on MSW, so there is no Redis and no recorder to snapshot anything. `MockMetricsHistoryProvider` stands in for `RedisMetricsHistoryProvider`: it synthesises 90 days of hourly throughput per queue at startup and then serves, purges and measures it like stored data. The series has a daily curve peaking in the afternoon, quieter weekends, a mild upward trend towards today, and rare hours where failures spike, all seeded off the queue name so the charts are stable across reloads. Queue baselines come from the job counts already in the fixtures, so the history table ranks queues the same way the dashboard does.

It implements the full `MetricsHistoryProvider` surface, including `getUsage` and `purge`, so the storage panel is real rather than decorative. Byte costs per bucket come from the storage table in the metrics readme, so the numbers land in the same ballpark a real recorder would produce. Purging mutates the in-memory series for real and a reload brings everything back.

Two smaller things came out of wiring it up:

- `handlers.ts` used to call `setApiRoutes(appRoutes.api.map(...))` after `createBullBoard` to swap in the browser-safe flow handler, which silently dropped any route `createBullBoard` had registered conditionally. Replaced with a `mapApiRoutes` shim on `MSWServerAdapter` that rewrites the routes already registered.
- The demo's uiConfig is injected statically by rsbuild rather than by the entry route, so `hasHistoryProvider`, `hasHistoryUsage` and `canPurgeHistory` have to be repeated there. Noted in a comment next to both copies.

`hashStr`/`mulberry32` moved out of `MockAdapter` into `prng.ts` since both files now need them.

### Metrics history page, 90d

Metrics history page

### Storage panel

Storage panel

### Longer ranges on a queue chart

Queue chart at 90d

## Readme

Adds `@bull-board/metrics` to the packages table and a short section explaining what it is for, why BullMQ alone can't do it, and that it is opt-in, linking to the existing recipe and the demo.

## Verification

- `tsc --noEmit` and `oxlint` clean on the demo workspace
- `yarn build` in `website/demo` succeeds
- Walked the demo in a browser: history page at 7d/30d/90d, per-queue ranges, storage panel, and a trim purge (recorded range narrowed from 2026-04-24..2026-07-23 to 2026-07-16..2026-07-23, hourly tier dropped from 506 kB to 43.1 kB, trim button correctly disabled afterwards). No console errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,12 +65,23 @@ That's it! Now you can access the `/admin/queues` route, and you will be able to
 
 See the [docs](https://felixmosh.github.io/bull-board/) for queue adapter options (read-only, retries, formatters, visibility guard), BullMQ Pro setup, board UI config, and more.
 
+## Historical metrics
+
+BullMQ keeps only a short ring buffer of per-minute metrics, so the throughput chart can't look back further than an hour or so. The optional `@bull-board/metrics` package (beta) snapshots those metrics into long-retention Redis buckets and feeds them back to the board, which adds a Metrics history page and 7/30/90 day ranges on every queue chart. It is entirely opt-in: without it the core stays stateless and writes nothing.
+
+```sh
+npm install @bull-board/metrics
+```
+
+See the [historical metrics recipe](https://felixmosh.github.io/bull-board/recipes/historical-metrics) for the recorder setup and storage sizing, or try it on the [live demo](https://felixmosh.github.io/bull-board/demo/).
+
 ## Packages
 
 | Name                                                                     | Version                                                  | Downloads                                                                         |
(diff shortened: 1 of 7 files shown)


felixmosh (approved):
👑

PR #1284, opened 2026-07-23
fix(api): remove job scheduler runs without destroying the schedule

### LINKED ISSUE #908: bug: Removing repeatable jobs added via job scheduler fails.

Stosiu:
Fixes #908.

Cleaning a failed run of a job scheduler either deleted the whole schedule and returned a lying 204, or returned a 500 when the scheduler was already gone.

`cleanJob` treated any job carrying a `repeatJobKey` as a scheduler job. BullMQ is much narrower: it blocks removal only for `repeat:<key>:<current millis>`, the single run a scheduler is waiting on. Past failed and completed runs remove normally.

So `cleanJob` now tries `job.remove()` first and only special cases the job when BullMQ rejects it with `ErrorCode.JobBelongsToJobScheduler` (`-8`, on `error.code`). The scheduler check is the first statement in the `removeJob` lua, before any mutation, so the attempt is side effect free. Failed cron runs delete cleanly and the schedule keeps running.

Removing just the pending run is not possible: it leaves the scheduler listed in `getJobSchedulers()` but never firing again. `clean` therefore returns 400 with `code: "JOB_BELONGS_TO_JOB_SCHEDULER"` and `jobSchedulerId`, and a new `PUT /api/queues/:queueName/job-schedulers/:schedulerId/remove` does the destructive part. The UI turns that 400 into a confirm dialog naming the schedule, shown regardless of the "confirm job actions" setting.

`-8` turned out not to be the only outcome worth handling. `removeJob` also answers `0` when a worker holds the job, which `Job.remove()` raises as an `Error` with no code, so cleaning an active job returned `500 Internal server error`. That is a transient conflict rather than a server fault, and it is reachable whenever a worker picks up a job between the list rendering and the click. It is now a 409 naming the job, keyed off the job state rather than an error string, since BullMQ gives no code for it.

Both of those failures also showed that the error toast only ever displayed the generic `error` field and dropped `message`, which is why the original report in #908 had to be read out of devtools. The toast now fills in the `Toast.Description` slot that the toaster already renders.

## Errors carry translation keys now

Review follow-up, taken repo wide rather than applied to the new handler alone. All 17 error sites in `packages/api` are built by `errorResponse()` (`packages/api/src/errors.ts`) and return a key:

```ts
errorResponse(404, 'ERRORS.QUEUE_NOT_FOUND');
errorResponse(400, { key: 'ERRORS.STATUS_NOT_RETRIABLE', options: { status: queueStatus } });
```

`error` is always `{ key, options? }`. `message` is `string | { key, options? }`, the string being for runtime text with no key, which is only the thrown error's message in `handlers/error.ts`. All keys translated in the ten non-English locales.

Two guards keep it that way. `ErrorTranslationKey` is a closed union, and the UI resolves it through a `t()` typed against `en-US/messages.json`, so a missing key fails the type check by name; a locale left behind fails the existing `i18n.spec.ts`. What types cannot catch is a handler skipping `errorResponse()` and hardcoding a string, since `body` is `Record<string, any>`, so `error-responses.spec.ts` scans the source for that shape. Say the word if you would rather not have a source-scanning test.

## Breaking changes

Cleaning the pending run of a schedule used to return 204 and delete the schedule. It now returns 400 and deletes nothing, so anyone driving the API directly needs the new endpoint. `HTTPStatus` gains `409`.

The `error` field was an English string and is now `{ key, options? }`. Consumers displaying it verbatim need to resolve the key, or read `code`.

## Tests

Both scheduler failure modes, the 400, the 409, the new endpoint including its 404 and read only cases, and the dialog. For the error contract: `errorResponse()` units, the source scan, and `packages/ui/tests/services/Api.spec.ts` for the rendering path. Verified across every adapter suite and by hand against real Redis. `CONTRIBUTING.md` documents adding a key.


packages/api/src/handlers/removeJobScheduler.ts
```diff
@@ -0,0 +1,26 @@
...
+  queue: BaseAdapter
+): Promise<ControllerHandlerReturnType> {
+  const { schedulerId } = req.params;
+
+  const removed = await queue.removeJobScheduler(schedulerId);
+
+  if (!removed) {
+    return {
+      status: 404,
+      body: { error: 'Job scheduler not found' },
```
felixmosh: Errors should return translation keys... not hardcoded english text

felixmosh: `message` should be of type, `string | {key: string, options: Record<string, any>}}

felixmosh (changes requested):
maybe it is not related to this PR, but it is a good opportunity to handle this.

packages/api/src/handlers/removeJobScheduler.ts
Stosiu: Done, and taken repo wide as you suggested. All 17 error sites in `packages/api` now go through `errorResponse()` and return a key instead of prose, including the pre-existing ones in the providers, `retryAll`, `retryJob`, `obliterateQueue`, `setGlobalConcurrency` and the metrics history handlers.

Stosiu: `message` is now `string | { key: string; options?: Record<string, any> }`. Two deviations worth calling out: `options` is optional, otherwise the eight sites that interpolate nothing carry `options: {}`; and `error` is stricter than the union, always `{ key, options? }`, so English cannot be hardcoded in the headline. The string escape hatch survives only on `message`, used in one place, `handlers/error.ts`, where the detail is a thrown error's own message. Happy to widen `error` too if you prefer the fields symmetrical.

Stosiu:
@felixmosh both points are fixed in 1e71369 and 82baa93, and taken repo wide as you suggested: all 17 error sites in `packages/api` now return translation keys, and `message` is `string | { key, options? }`.

The part worth a look is what stops it regressing. `ErrorTranslationKey` is a closed union, and the UI resolves it through a `t()` typed against `en-US/messages.json`, so a key the locale file lacks fails the type check by name. A locale left behind fails the existing `i18n.spec.ts`. What types cannot catch is a handler skipping `errorResponse()` and hardcoding a string, since `body` is `Record<string, any>`, so a test scans the source for that shape. Happy to drop that one if it feels out of place.

Details and the second breaking change (`error` is now `{ key, options? }`) are in the updated description.

felixmosh (approved):
Thank you for this fix

PR #1289, opened 2026-07-27
feat(ui): show completed vs failed split in the metrics history queue table

Stosiu:
The bar in the "By queue" table only encoded `completed / maxCompleted`, so a queue that completed 42k jobs and failed 9.7k of them drew the same green bar as one with no failures at all. The failure count sat in the next column and the bar never acknowledged it.

Now the bar spans total runs and splits into completed and failed. Length still compares queues, the split compares outcomes within one queue, and hovering gives exact counts. The failure rate sits beside the failed count, because "9,724 failed" means something quite different once you know it's 15.6% of that queue's work rather than a rounding error. Non-zero segments clamp at 1% of the track, otherwise a sub-percent failure rate draws nothing at all. Sorting moved from completed to total runs so bar length stays monotonic down the table.

The bar also used to share a cell with the completed count, which let the count's digit width decide where each track started: "333,492" pushed its bar 13px right of "267". Separate columns now, fixed 240px, same x on every row. Chasing that turned up two more. The failure-rate slot was sized in `ch`, which resolves against the header's font rather than the cell's and left the FAILED header 44px off its own numbers. And the header offset wasn't applying at all, since `.table th`'s padding shorthand outranks a bare class.

Three smaller things while I was in here. Chart tooltip values go through `toLocaleString()`, so a busy day reads 403,775 rather than 403775. Page `max-width` goes from 1120px to 1600px, because the card was stopping well short of the space on a wide display. And the segments size with `translateX`/`scaleX` instead of animating `width`, so a totals refresh doesn't relayout every row.

Before:

After:

The three new keys are translated in all eleven locales rather than English-filled. `Tooltip` takes an optional `className` now, existing call sites unchanged. Docs screenshot and the paragraph describing the table are updated to match.


felixmosh (approved):
LGTM

PR #1294, opened 2026-07-28
feat: show connected workers per queue

### LINKED ISSUE #1292: Feature: show connected workers per queue
bull-board currently can't tell you whether anything is consuming a queue. A queue with a growing waiting count looks exactly the same whether it is overloaded or whether all its workers died, and the second case is the one you want to catch quickly.

Both Bull and BullMQ expose `getWorkers()`, and it is cheap (a `CLIENT LIST` filter). Proposal, read-only:

- List connected workers per queue: client name, address, connection age.
- Show a "no workers connected" hint on the queue card and queue page when the list is empty and the queue is not paused.
- Fetch on the existing polling interval, or on demand from a modal like the Redis stats one, whichever fits the current data flow better.

No worker actions in the first version. BullMQ has no remote worker control anyway, so there is nothing meaningful to offer beyond visibility.

Every comparable dashboard already has this: Sidekiq's Busy tab, Flower's workers page, Asynqmon's Servers view, Taskforce's workers overview. bull-board is the odd one out here.

I can pick this up.

Stosiu:
Closes #1292.

A queue with a growing waiting count looks identical whether it is overloaded or whether every worker consuming it has died, and the second case is the one you want to catch quickly.

The dashboard shows nothing at all while a queue is healthy. A badge appears only once a queue has no workers and is not paused, on its overview card and beside the status tabs on its own page, so the only room the feature ever takes is for the case worth acting on. A paused queue is supposed to have nothing consuming it, so it never warns.

On the queue page it sits with the status tabs, next to the 73 waiting jobs and the flatlined throughput chart it explains.

Clicking the badge opens the queue info panel on its **Connected workers** section. Each row leads with whatever identifies the worker, which is the name when you gave it one and the address when you did not, both of which are in the shot below. No worker actions, as the issue proposed: BullMQ has no remote worker control.

The list lives only there, as a section rather than a window of its own, with the count beside global concurrency in the overview. The badge only exists while something is wrong, so the panel is how you check on a queue that looks fine.

`BaseAdapter.getWorkers()` delegates to Bull's and BullMQ's own `getWorkers()`, with Pro inheriting BullMQ's. The base returns `null` rather than an empty list, which keeps subclasses outside this repo working and, more importantly, keeps "could not ask" distinct from "asked, nobody is there". Only the empty list warns; `null` shows nothing at all. That covers an adapter without the method, a queue whose Redis is unreachable (caught per queue, so one bad connection does not take down the response for the rest of the board), and providers that block `CLIENT LIST`, where Bull resolves `undefined` and BullMQ returns a placeholder entry. Both normalize to `null`, so Memorystore never renders as every queue having lost its workers.

Naming is BullMQ's alone, so the `:w:` separator it appends to a named worker's connection name is passed into `normalizeWorkers` from the BullMQ adapter rather than sitting in the base. Bull passes nothing, which is correct for it.

Whether a queue has workers rides along with the queue listing as `hasWorkers`, following the `jobSchedulerCount` refactor in c0f4d901, so the warning costs no request of its own. It is typed `boolean | null` rather than a plain boolean because `null` has to keep meaning "could not ask", for the reasons above. Only `false` warns.

The full list, which only the info panel shows, comes from `GET /api/queues/:queueName/workers` and is fetched once when the panel opens, mirroring `default-job-options`. Nothing polls it.

What that costs is one `CLIENT LIST` per queue per poll, cheap on a handful of queues and less so on dozens, so `showWorkers: false` in `uiConfig` turns it off the way `showMetrics` gates its chart. It stops the command being run rather than hiding the badge: nothing is asked of Redis during the poll, and the per-queue route answers 403. I defaulted it to `true` on the grounds that a warning shipped off never fires for the people who would most benefit, but say the word if you would rather it match `showMetrics`.

I had connection idle time in the payload at first and dropped it. Workers register a blocking connection whose idle time keeps climbing while a job is actually running, so it reports "idle" for the busiest worker on the board.

The 403 goes through `errorResponse` with a new `ERRORS.WORKERS_DISABLED` key instead of the hardcoded `'Forbidden'` it started as, so it passes the guard test that scans handlers for untranslatable English. The badge uses the `className` prop `Tooltip` gained, which let me drop a wrapper span I had added to stop the button riding the tooltip's line box a pixel off centre.

The demo runs on mock adapters, which inherit the `null` default and so would have said nothing about workers at all. `MockAdapter` now generates deterministic ones: a mix of named and unnamed across the thirteen queues, and `reports:export` deliberately left with none so the warning state is on show rather than only the healthy one.

Tests cover both adapters against real Redis (named and unnamed workers, nothing consuming, prefixed queues, the separator being read only when passed, visibility, unreachable queue, blocked `CLIENT LIST`, the 403 and the 404), the `hasWorkers` flag on the listing including that opting out asks Redis nothing and that one unreachable queue does not cost the others their listing, and the UI surfaces, including the paused queue that must not warn, that the badge issues no request, and that the panel asks once on open and nothing while closed. Eleven new keys, translated across all eleven locales rather than English-filled. Docs get a section in the dashboard guide, a UIConfig row, and a line in the alerting recipe; the queue info screenshot is refreshed since that panel gained a row.

Rebased onto master, which squashed the branch to keep the conflict resolution reviewable, so the history here is the feature plus one commit for this review round.


felixmosh:
Do you think that this info should be on the page itself?
I don't see how this info will be used by me, it takes real estate on the page.

Suggestion, just add a workers section on Queue info modal. WDYT?

Stosiu:
@felixmosh yeah, that's in already. Connected workers is a section in the queue info modal now, and I deleted the standalone modal I'd built, so the badge just opens the panel on that section.

The real estate point is fair too. The overview doesn't spend any anymore: a card shows nothing while a queue has workers. All that's left on the page is a warning when a queue has none and isn't paused.

That warning is the one part I'd argue for keeping. A queue whose workers all died looks exactly like a busy one, so if it only lives in the modal you have to already suspect it to go find it. The count in the queue page header I don't care about, happy to drop that.

Want me to pull the warning as well, or leave it?

felixmosh:
Keep only the warning 🙏🏼

Stosiu:
@felixmosh already done, pushed a few minutes before your comment landed. Page only shows the warning now, the count sits in the queue info panel.

Last open thing, if you have a view: `showWorkers` defaults to `true` at the moment. `showMetrics` defaults to `false`, so happy to flip it to match.

packages/api/src/handlers/queueWorkers.ts
```diff
@@ -0,0 +1,33 @@
...
+      }
+
+      // A queue whose Redis is unreachable should not take the whole board down,
+      // it is reported the same way as one that cannot answer.
+      const queueWorkers = await queue.getWorkers().catch(() => null);
+      return [queueName, queueWorkers] as const;
+    })
+  );
+
+  for (const entry of entries) {
```
felixmosh: Looks better.
```
const workers: QueueWorkers = Object.fromEntries(entries.filter(Boolean));
```

packages/api/src/queueAdapters/base.ts
```diff
@@ -11,9 +11,15 @@ import {
   QueueJobOptions,
   QueueMetrics,
   QueueType,
+  QueueWorker,
   Status,
 } from '../../typings/app';
 
+/** The `:w:<name>` suffix BullMQ appends to the connection name of a named worker. */
```
felixmosh: not belongs here, pass if from the BullMQ adapter

packages/ui/src/components/WorkersBadge/WorkersBadge.tsx
```diff
@@ -0,0 +1,61 @@
...
+);
+
+/**
+ * Warns when nothing is consuming a queue, and shows nothing at all otherwise.
+ * The count itself belongs in the queue info panel: a healthy number repeated across the
+ * board is real estate spent on the case you never need to act on.
+ */
+export const WorkersBadge = ({ queue }: { queue: AppQueue }) => {
+  const { t } = useTranslation();
+  const { workers } = useQueueWorkers(queue.name);
```
felixmosh: I don't think that this badge requires a separate polling, can we add boolean value on queue poll for this badge?

packages/ui/src/components/QueueInfoModal/QueueInfoModal.tsx
```diff
@@ -45,17 +49,25 @@ function formatOptionValue(value: unknown): string {
...
-  const [openSection, setOpenSection] = useState<InfoSection | ''>('overview');
+  const [openSection, setOpenSection] = useState<InfoSection | ''>(initialSection);
   const toggleSection = (section: InfoSection) =>
     setOpenSection((current) => (current === section ? '' : section));
 
   const totalJobs = queue.statuses.reduce((sum, status) => sum + (queue.counts[status] || 0), 0);
   const { defaultJobOptions } = useQueueDefaultJobOptions(queue.name, open);
   const optionEntries = Object.entries(defaultJobOptions || {});
+  // `null` means the queue cannot report workers at all, so the panel says nothing about them.
+  const { workers } = useQueueWorkers(queue.name);
```
felixmosh: here, there is no need for polling, just fetch the list once, no?

felixmosh (changes requested):

packages/ui/src/components/WorkersBadge/WorkersBadge.tsx
Stosiu: Done, `hasWorkers` is on the queue listing now and the badge makes no request of its own.

One thing I want to check with you though, since it is not quite the boolean you asked for: I typed it `boolean | null`. The `null` is for "could not ask" as opposed to "asked, nobody is there". Memorystore and friends block `CLIENT LIST`, and an adapter that does not implement `getWorkers` returns nothing either, so if both collapsed into `false` every queue on those setups would show the warning as though its workers had just died. Only `false` warns, `null` renders nothing.

If you would rather keep the payload a plain boolean I can, but then the unknown case needs to go somewhere, and everything I can think of costs more than the union does. Happy either way, just did not want to quietly widen the type on you.

felixmosh: it can be `boolean | null`, no problem :]

felixmosh (approved):

PR #1295, opened 2026-07-28
feat(metrics): job latency histograms with wait, run and queue-age

Stosiu:
`@bull-board/metrics` can tell you how much work a queue did. It cannot tell you how long any of it took. This adds that: two latency histograms per queue, wait time and run time, plus a backlog gauge, collected on the recorder tick that already exists and needing no changes to your worker code.

Wait time is `processedOn - timestamp`, how long a job sat before a worker picked it up, and it is the signal that says scale out. Run time is `finishedOn - processedOn`, how long the handler took, and it is the signal that says the handler regressed. They stay separate because they point at different fixes.

## Why no worker instrumentation

BullMQ's `moveToFinished` does `ZADD targetSet, timestamp, jobId`, so the completed and failed sets are sorted sets scored by finish time. Scanning past a stored watermark returns exactly the jobs that finished since the last tick, with no gaps, which is the same pattern the counter recorder already uses. The main BullMQ Prometheus exporter goes the other way, `QueueEvents` plus a `getJob()` per completion, which costs a dedicated blocking Redis connection per queue and silently drops jobs already trimmed. Polling avoids both.

Each scan stops at `now - 5s` and the watermark advances to that bound rather than to the highest score seen. `ZADD` and the scan are not atomic, so a job finishing microseconds before a scan can be missing from it, and advancing to the highest observed score would skip it permanently.

## The gauge exists because the histogram lies when it matters

A completion-derived wait histogram only sees jobs that finished. When a queue is genuinely backed up, jobs stop finishing, so the histogram goes quiet exactly when the number matters most and a stalled queue looks identical to an idle one. So the recorder also samples the age of the oldest job still waiting, and the UI draws it on the same axis as the wait chart. This is the same reason Sidekiq's `Queue#latency` measures queue age rather than job latency.

That gauge reads the `wait` list, the `paused` list and both ends of `prioritized`. A paused queue has its list renamed, and a queue where every job carries a priority has a permanently empty `wait` list, so reading only `wait` would report a reassuring zero for a badly backed-up queue.

It rolls up across queues as a MAX rather than a sum, since the useful cross-queue number is the oldest job waiting anywhere on the board and adding ages together would mean nothing.

## Storage, measured

18 fixed bucket boundaries, packed as a CSV vector into one hash field per time bucket rather than one field per bucket. Redis hash entry overhead dwarfs a two-byte count, so packing is roughly 8x cheaper and makes the bucket count nearly free. Hour and day tiers only; percentiles over a single minute are noise.

Measured against real Redis rather than estimated:

| | |
|---|---|
| One queue, 90 day retention, both histograms plus the gauge | 254.5 KB |
| Same, pathological, all 18 buckets populated every hour | 574.6 KB |
| Cross-queue rollup | about 224 KB once for the whole board, not per queue |
| `sample()` at 1000 finished jobs in a tick | 9.11 ms, about 12 Redis round trips, flat in job count |
| Subsampling cap | 5000 jobs 30.39 ms, 10000 jobs 29.94 ms |

A 200 queue board is therefore roughly 50MB plus the one shared 224KB. An earlier estimate in this branch said 700KB per queue; measuring showed that was overstated about 2.75x, and the figure was corrected before merge.

## Reading it back

Percentiles are computed server-side, because percentiles do not merge but histograms do. A weekly p95 cannot be averaged from 168 hourly p95s; the bucket vectors are summed across the range and the percentile computed once. That is why storage holds buckets rather than precomputed percentiles.

`getLatency` is a fourth optional method on `MetricsHistoryProvider`, alongside `getUsage` and `purge`, so a custom read-only provider stays valid. Its route is registered only when a provider implements it, and `hasLatencyHistory` gates the UI the same way `hasHistoryProvider` already does.

## In the UI

One chart at a time, behind a Throughput / Latency tab, on the queue page and the Metrics history page alike, each reusing the range selector its page already has. The totals above the chart follow the tab: completed and failed counts on Throughput, p95 run and p95 wait for the range on Latency.

Run time, wait time and queue age share one chart because they share a unit. Throughput does not, and a stacked or side-by-side arrangement made that mismatch visible: the two Y axes have different label widths, so their plot areas started at different offsets and the date ticks did not line up. A tab removes the comparison rather than trying to force the axes to agree.

The latency chart carries seven possible series. Run p95, wait p95 and the queue-age gauge are on by default, since those answer the two questions people arrive with: is the handler slow, and is the backlog growing. p50 and p99 for each are one click away in the legend. Hue distinguishes the metric and line weight distinguishes the percentile, because six unrelated colours is how a chart like this becomes unreadable.

Axes are logarithmic and labelled as such. Latency spans orders of magnitude, and on a linear axis a p99 measured in minutes flattens p50 and p95 into a single line along the bottom.

The tab stays in place at every range. Latency is only stored hourly and daily, deliberately, because percentiles over a single minute are noise, so selecting Latency on the 60m range explains that rather than showing an empty chart or making the control disappear.

The final segment of every series is drawn dashed when the newest bucket is still forming, with a tooltip saying so. Today's bucket only covers the hours elapsed so far, so against complete prior days it otherwise renders as a cliff at exactly the point a reader looks first.

Queue page, Latency tab. This queue's handler regressed a few days ago: run p95 climbs, and the dashed queue-age line climbs above the wait percentiles as the backlog builds. That divergence is why the gauge shares the axis.

Metrics history, the cross-queue view, with the p95 run-time column on the right of the by-queue table. The tiles above the chart have switched to p95 run and p95 wait for the range.

## Notes

On by default, `latency: false` turns it off. `onLatencyError` is available because a latency tick that fails is swallowed so it cannot take the counter snapshot down with it, which means a collector broken since startup otherwise looks exactly like an idle board.

`removeOnComplete: true` leaves nothing to scan, so no latency data will ever appear for such a queue. Documented, since it is the most likely reason someone sees an empty chart.

Rebased onto master after #1284 landed. Reusing that PR's exact error-key names paid off: the two `ErrorTranslationKey` unions merged by adding a single new entry, `ERRORS.INVALID_METRIC`, and `errors.ts` resolved to master's copy unchanged. The metrics history handler's plain-English errors are already fixed on master by that PR, so this branch no longer carries that change.

`retry-job.spec.ts` had its timeout raised to 30s, matching the eight other workspaces. BullMQ wraps its worker loop in `retryIfFailed` with a 15s retry delay, so one transient Redis error parks the worker past the old 5s jest timeout. Not a race, and unrelated to this change, but it was reddening CI intermittently.

The demo at `example.ts` now backfills 30 days of plausible history on a fresh Redis, counters and latency both, so the charts show something immediately instead of staying empty until the recorder has run for days. It skips entirely if any of those days already hold data, and `DEMO_BACKFILL=0` opts out.


Stosiu:
@felixmosh a note on the size of this one, since 66 files is a lot to land at once.

It can be split, and I will if you prefer it. The seam is clean: roughly 2400 lines in `packages/metrics` and `packages/api` for collection, storage, the provider method and the route, then roughly 2400 in `packages/ui` for the charts, of which 380 is locale files. Two PRs, backend then UI.

I kept it whole because the backend half is not really reviewable on its own. It ships a route nothing calls and a storage format nothing reads, so the only way to judge whether the design is right is to look at the charts it produces. That is not a hypothetical concern here. Four decisions I would have got wrong on paper only became obvious once it was rendered against real data:

The Y axis had to become logarithmic. Latency spans orders of magnitude, and on a linear axis a p99 measured in minutes flattened p50 and p95 into a single line along the bottom, so two of the three series said nothing.

The newest bucket needed marking as partial. Today's day only covers the hours elapsed so far, so against complete prior days it drew as a near vertical cliff at exactly the point a reader looks first.

The queue-age gauge had no cross-queue rollup, so on the Metrics history page that series was a legend entry with nothing behind it. `recordQueueAge` only ever wrote queue-scoped keys.

And the gauge itself was reading only the `wait` list, which meant a paused queue reported an age of zero. On the demo Redis that turned out to be hiding two paused queues holding 4,565 and 4,640 jobs whose oldest was ten days old.

None of those are visible in a backend diff.

On prior art, the collection approach is deliberately different from what is already out there. `ron96G/bullmq-exporter`, the main Prometheus exporter for BullMQ, uses `QueueEvents` plus a `getJob()` per completion and silently drops jobs already trimmed. That costs a dedicated blocking Redis connection per queue, since `QueueEvents` duplicates the connection and blocks on `XREAD`. This reads the finished sorted sets instead: `moveToFinished` does `ZADD targetSet, timestamp, jobId`, so scanning past a stored watermark on the tick that already exists returns exactly the jobs finished since the last one, with no extra connections and no changes to anyone's worker code.

The queue-age gauge is Sidekiq's `Queue#latency` idea. A completion-derived histogram only sees jobs that finished, so when a queue is genuinely backed up and jobs stop finishing, it goes quiet exactly when it matters and a stalled queue looks identical to an idle one. Sidekiq measures the age of the oldest job instead for that reason, and drawing both on the same axis makes the gap visible.

Bucket layout follows Prometheus convention rather than anything exotic. Fixed boundaries, no sketches, percentiles computed by summing bucket vectors across the range and computing once, because percentiles do not merge but histograms do.

Why it earns its place: the package can already tell you how much work a queue did and cannot tell you how long any of it took, which is the question you actually have during an incident. Wait time says scale out, run time says look at the handler, and those point at different fixes. Storage measured against real Redis is about 254KB per queue at 90 day retention, plus one shared 224KB for the cross-queue rollup rather than per queue, and about 9ms per tick at 1000 finished jobs.

Happy to split it into two if that reviews better, or to drop the demo backfill and screenshots into a separate change to shrink the diff.

Stosiu:
A map of the diff, since 66 files is hard to navigate cold. This is the path one job's duration takes, with the file that owns each step.

Nothing in your worker code is touched. The dashed box is BullMQ's own keyspace, which is only ever read.

```mermaid
flowchart TD
    subgraph bullmq ["BullMQ keyspace (read only, never written)"]
        direction LR
        W["Worker finishes a job"] --> Z["moveToFinished:<br/>ZADD completed, finishedOn, jobId"]
        Z --> H["job hash:<br/>timestamp, processedOn,<br/>finishedOn, atm"]
        WAIT["wait / paused / prioritized"]
    end

    subgraph collect ["Collection, on the recorder tick that already exists"]
        REC["MetricsRecorder.ts<br/>existing timer, latency on by default"]
        SAMP["LatencySampler.ts<br/>scan past watermark, bounded at now-5s<br/>lease per queue, retry and skew handling"]
        HIST["histogram.ts<br/>18 fixed buckets, pack, merge, quantile"]
    end

    subgraph store ["Storage"]
        LS["LatencyStore.ts<br/>Lua: merge packed vectors, max gauge"]
        R[("Redis<br/>hour + day tiers<br/>per queue and __global__")]
    end

    subgraph read ["Read path"]
        PROV["RedisMetricsHistoryProvider.ts<br/>getLatency, sum vectors then quantile once"]
        API["handlers/metricsLatency.ts<br/>GET /api/metrics/latency"]
    end

    subgraph ui ["UI"]
        HOOK["useLatencyMetrics.ts"]
        TABS["MetricsChartTabs.tsx<br/>Throughput / Latency"]
        CHART["LatencyChart.tsx + latencySeries.ts<br/>log axis, 7 series, partial bucket"]
    end

    Z -.->|ZRANGEBYSCORE| SAMP
    H -.->|pipelined HMGET| SAMP
    WAIT -.->|oldest job| SAMP
    REC --> SAMP
    SAMP --> HIST
    HIST --> LS
    LS --> R
    R --> PROV
    PROV --> API
    API --> HOOK
    HOOK --> TABS
    TABS --> CHART
```

Two things the diagram is meant to make obvious.

The scan is bounded at `now - 5s` and the watermark advances to that bound rather than to the highest score seen. `ZADD` and the scan are not atomic, so a job finishing microseconds before a scan can be absent from it, and advancing to the highest observed score would skip it permanently.

The quantile is computed once, at the provider, after summing bucket vectors across the range. Percentiles do not merge, so a weekly p95 cannot be averaged from seven daily p95s. That single property is why storage holds buckets rather than precomputed percentiles, and it is what the `range` granularity used by the table column exists for.

Storage shape, for the sizing claim:

```
bull-board:metrics:<queue>:runtime:hour:<YYYY-MM-DD>   field <hour>  -> "0,0,12,45,3,..."   18 counts, packed
bull-board:metrics:<queue>:runtime:totals              field <day>   -> "0,0,c,d,e,..."
bull-board:metrics:<queue>:waittime:...                same shape
bull-board:metrics:<queue>:queueage:...                scalar ms, stored as a max not a sum
bull-board:metrics:__global__:...                      same four, rolled up once for the whole board
```

Packing all 18 counts into one hash field rather than one field per bucket is the difference between roughly 254KB and roughly 2MB per queue at 90 day retention, because Redis hash entry overhead dwarfs a two byte count. It also makes the bucket count nearly free, which is what pays for the long tail out to 6h that wait time needs.

Suggested reading order if it helps: `histogram.ts` first since it is pure and everything rests on it, then `LatencyStore.ts` for the two Lua scripts, then `LatencySampler.ts` which is where the concurrency lives, then the provider and route, then the UI.

felixmosh:
Can you fix the merge conflicts? Is it OK to trust you with `metrics` package, so each bug there, I will forward to you? :]

Stosiu:
Yes on both. I'll fix the conflicts soon, and happy to own `metrics`, so forward anything that comes in there and I'll take it.

Worth saying it's beta and I'd treat it that way for a while. I'm testing it as thoroughly as I can before adding anything new, and I'm also rolling it out to my own client projects so it gets exercised against real queues and real Redis rather than just the demo. Anything that shakes out there comes back here as fixes.

Stosiu:
@felixmosh fixed

felixmosh:
can you fix the merge?

Stosiu:
@felixmosh fixed 🙋🏻

PR #1304, opened 2026-07-28
feat: schedulers view with per-scheduler edit and removal

### LINKED ISSUE #1290: Feature: schedulers view (list all schedulers, edit cron, remove a single one)
Right now the only way to see schedulers is indirectly, through the repeat info on delayed jobs. There is no place that shows all schedulers, when each one last ran, or when it fires next, and no way to act on a single scheduler from the UI. #827 asked for this before the stale bot closed it, and #908 (fixed in #1284) showed how confusing the current indirection is: cleaning a delayed job could silently remove the whole schedule behind it.

I'd like to add a schedulers view. Concretely:

- A global "Schedulers" page (next to the overview) listing schedulers across all queues via `queue.getJobSchedulers()`: queue, scheduler id, cron pattern or `every` interval, next run, last run, template name and data. Filterable by queue, the same way the sidebar filters queues today.
- The queue page links into it pre-filtered, so "schedulers of this queue" is one click without duplicating a whole tab.
- Remove a single scheduler, using the `removeJobScheduler` adapter method that landed in #1284.
- Edit the cron expression or interval through `queue.upsertJobScheduler()`, so a schedule can be fixed without redeploying the app. Guarded the same way as other destructive actions (readOnlyMode and friends).
- Legacy Bull: list via `getRepeatableJobs()`, remove via `removeRepeatableByKey()`, and hide the edit action since there is no upsert equivalent.

One thing I'd deliberately leave out of the first version: pausing/resuming a single scheduler. BullMQ has no API for that, so it would mean bull-board persisting the scheduler definition itself and re-upserting it later, which gets messy if the app re-upserts the same scheduler on deploy while it is "paused". Worth a separate discussion if there's interest.

Prior art, for what it's worth: Sidekiq, Asynqmon and Taskforce all have a dedicated cron/scheduled view with last run and next run. It's a standard part of these dashboards at this point.

I'll implement this if you're on board.

Stosiu:
Closes #1290.

Adds the schedulers view we discussed there: a board-wide page listing every job scheduler across all queues, with removal and cron editing, and a sidebar entry that only appears once there is a scheduler to look at.

## Last run, and why it is trustworthy

The issue asked for a last-run column. BullMQ stores no such field: the `repeat:<id>` hash holds `name, ic, limit, startDate, endDate, tz, pattern, every, offset, data, opts` and nothing about previous runs. Sidekiq-cron and asynq show one only because their scheduler layer persists it.

It is still derivable. The worker creates the next delayed job at the moment the previous iteration moves to active, stamping it with `timestamp` and the id `repeat:<schedulerId>:<scheduled millis>`. So the pending job's timestamp is when the previous run started. Measured against a live queue, the derived value and the run's own `processedOn` were 1 ms apart. A scheduler that has not fired yet (`iterationCount` of 1) and one whose pending job is gone both report nothing rather than guessing, and legacy Bull reports nothing at all.

The times link to the jobs behind them. The next run always links, since that delayed job is what will run, so you can open it, read the payload or promote it. The last run links when the job can be named and still exists, which means interval schedules whose previous run has not been trimmed by `removeOnComplete`. Naming the previous run of a cron schedule would mean parsing the pattern backwards, so those show the time alone rather than a link that is dead more often than not.

## API

- `BaseAdapter` gains `getJobSchedulers`, `getJobSchedulersCount` and `updateJobScheduler`, plus a `supportsJobSchedulerUpdate` capability flag. **This is the one thing worth your opinion**: they are abstract, following the precedent `removeJobScheduler` set in #1284, so anyone with a custom adapter outside this repo has to implement them. The alternative is non-abstract defaults returning `[]` and `0`, which keeps third-party adapters compiling at the cost of quietly reporting "no schedulers" for an adapter that has them. Happy to switch if you prefer the softer option.
- `GET /api/job-schedulers`, optionally narrowed with `?queueName=`.
- `GET /api/job-schedulers/count`, one `ZCARD` per queue, returning `{ total, byQueue }` with the empty queues left out. This is what the sidebar and the queue page gate on, so neither has to pull the full listing on every page load, and the polled `/api/queues` payload is untouched.
- `PATCH /api/queues/:queueName/job-schedulers/:schedulerId` rewrites the schedule only. The stored template is re-sent verbatim, so the job a scheduler produces keeps the name, data and options the application registered. A cron BullMQ cannot parse comes back as a 400 with the stored schedule untouched, since BullMQ works out the next fire time before it writes anything.
- Legacy Bull lists and counts repeatables, and `removeJobScheduler` now actually removes one through `removeRepeatableByKey`; the stub added in #1284 always returned false. Editing answers 405, since Bull has no upsert.

## UI

The page shows queue, scheduler id, job name, schedule, next run, last run, run count against its limit, and the job template behind an expander. Editing and removal respect `readOnlyMode`, and editing is hidden for Bull queues. A queue that has schedulers links into the view pre-filtered.

Both actions are deliberately described as temporary, because for most applications they are. Schedulers registered at boot come back on the next deploy, and `upsertJobScheduler` means the application's definition overwrites an edit made here. The modal and the confirmation say so, and the guide has a section on it, since the honest framing is "stop a misbehaving cron until the job is fixed" rather than "change the schedule".

## Drive-by fix: 204 responses on Elysia

Extending the contract battery surfaced a pre-existing bug. The Elysia adapter passed a body to the `Response` constructor along with a 204, which throws, so **every route answering 204 returned a 500 on Elysia**: retry a job, promote a job, clean a job, and the scheduler removal from #1284. One conditional in `ElysiaAdapter` fixes all of them, and the new contract cases now hold it in place.

## Tests

- 26 API tests covering listing, per-queue filtering, counting, the last-run derivation against a real worker, run-id naming in all four of its states, every validation branch of the edit, read-only and 404 handling, and the legacy Bull paths.
- The adapter contract battery grew four cases, so the listing, the count, the `PATCH` edit and the removal are proven on Express (4 and 5), Fastify, Hono, Koa, Hapi, H3, Elysia, NestJS and Bun. The `PATCH` case matters most: it is the only route in the API carrying a body on that verb.
- UI tests for the page, the sidebar gating, the run links, the remove flow, the form staying open on a refused schedule, plus unit tests for the interval formatting and the relative-time helper.

Screenshot in the guide, and the demo's mock adapter implements the new methods.


felixmosh (approved):
Great addition, thank you 🙏🏼

Issue #1293, opened 2026-07-27
Feature: design-token theming and Base UI component migration (whitelabel support)


Stosiu:
In #1112 you mentioned it would be cool to redesign the UI using base-ui / shadcn style components, without Tailwind. I'd like to actually pick that up, and I think there's a bonus hiding in it: whitelabel theming.

The plan I have in mind:

1. **Token layer first, non-breaking.** Consolidate `theme.css` into a documented set of semantic CSS variables using shadcn's naming, since that's become the vocabulary everyone knows: `--background`, `--foreground`, `--card`, `--primary`, `--muted`, `--destructive`, `--radius`, and so on. Light and dark become two value sets for the same tokens. Dark mode already works roughly this way, this step just makes the contract explicit.
2. **Expose overrides.** A `uiConfig.theme` option where integrators pass their own values for those variables. That makes bull-board whitelabel-able (company colors, radius, fonts) without forking, and the docs get a theming recipe with a couple of ready-made palettes.
3. **Migrate components to `@base-ui/react` incrementally.** The toast already runs on Base UI (#1281). Modals, dropdowns, tabs, tooltips and switches are the natural next candidates. Styling stays CSS modules referencing the tokens, so no Tailwind anywhere.
4. **The visual refresh rides along per component** instead of landing as one big redesign PR.

Things I'd want your call on before starting: adopt shadcn token names verbatim or define bull-board's own, should `uiConfig.theme` accept the full token map or a curated subset, and which component to migrate first.

All of it in small PRs, one step or one component each.

Stosiu:
One more angle I want to put on the table before we settle the plan above: using shadcn/ui itself instead of hand-styling Base UI components. I know you said no Tailwind in #1112, so feel free to shoot this down. I'd just rather ask once with the full picture than wonder later.

What we'd actually get:

- The components are the expensive part, not the tokens. Restyling dialogs, dropdowns, tabs, comboboxes and tooltips by hand on top of Base UI is real work per component, plus getting a11y and keyboard behaviour right every time. shadcn hands us finished, widely battle-tested implementations with the exact look we're after.
- It is not a dependency. shadcn vendors the component source into the repo, so we own and edit every line, same as the current components. No runtime lock-in, and if Tailwind ever becomes annoying the code is ours to restyle.
[...]

PR #1306, opened 2026-07-29
feat(ui): design token theming with whitelabel overrides and a visual refresh

Stosiu:
Picking up #1293. It follows what you asked for in the issue: shadcn's token names with our own values, no Tailwind anywhere, Base UI where a primitive actually helps, and the migration done lazily rather than as one big bang.

## The token layer

`packages/ui/src/theme.css` is now the single source of truth for the whole board. It defines the shadcn contract verbatim (`--background`, `--foreground`, `--card`, `--popover`, `--primary`, `--secondary`, `--muted`, `--accent`, `--destructive`, `--border`, `--input`, `--ring`, `--radius`), plus the sidebar and chart sets, and adds one bull-board namespace for the eight job statuses (`--status-failed`, `--status-completed`, and so on). Light and dark are two value sets for the same names.

Every one of the ~50 CSS modules now reads only from these, so the old ad-hoc variables (`--card-bg`, `--accent-color-d1`, `--button-primary-hover-bg`, and about eighty others) are gone. The highlight.js rules that used to live in `theme.css` moved to `hljs.css`, so `theme.css` is purely tokens and is the one file you edit to retune the board.

## On the "everything built with shadcn looks the same" worry

Fair, and I tried to design against it. The default palette is deliberately not zinc plus Inter plus soft shadows. Light is a cool paper, dark is blue ink rather than neutral grey, and the accent is a petrol teal, which is the one hue family the eight status colours leave free.

The bigger lever is type: IBM Plex Sans, with IBM Plex Mono and tabular figures for everything numeric (counts, ids, durations, timestamps). That is what most changes the feel. It reads like an instrument panel rather than a generic admin template. Surfaces are 1px borders instead of drop shadows, and the radius is 6px rather than the shadcn default.

Overview, light theme

Overview, dark theme

## Whitelabel

`uiConfig.theme` takes `{ light, dark }` maps of those token names and writes them into the page at boot:

```ts
createBullBoard({
  queues,
  serverAdapter,
  options: {
    uiConfig: {
      theme: {
        light: { primary: '#6d28d9', ring: '#6d28d9', radius: '0.75rem' },
        dark: { primary: '#a78bfa', ring: '#a78bfa' },
      },
    },
  },
});
```

Keys are typed as `ThemeTokenName`, unknown names are dropped, and values containing `;`, `{`, `}`, `<` or `>` are rejected, so a config can never inject arbitrary CSS or markup into the page. Because the names match shadcn's, palettes from the usual theme editors drop straight in. Documented in `website/docs/configuration/ui-config.md`.

Below is the board with nothing changed except a handful of those variables:

The same board with a violet brand theme applied through uiConfig.theme

Note the status colours stay put. They are separate tokens you can override too, but they keep their meaning by default.

## Queue cards

The card carries the queue name, a proportional composition strip, and at most two figures: active and failed, each only when non-zero, each linking to that filtered view. Failures are the only thing on the card allowed to shout.

You were right that five or six statuses will not fit there. I measured it against the narrowest card, a 320px grid column with 288px of content width: three labelled figures need 240px, five need 400px, six need 463px, and all eight need 678px, with "waiting children" alone taking 109px. Dropping the labels for a coloured dot plus number does not rescue it either, since six statuses still want 306px.

So the full breakdown moved to a hover panel instead, styled like the existing chart tooltips. It is built on Base UI's `PreviewCard` rather than `Tooltip`, because a tooltip popup closes as soon as the pointer enters it and its rows could not be links. Each row links to that status view, which also replaces the old 6px-tall bar segments as click targets.

Hovering a queue's strip reveals every status with its count

## Status counts on the overview filters

The overview status filters now carry board-wide counts per status. They were already links that filter the view, so the counts sit where you are about to act rather than in a separate read-only panel.

Queue page

## Base UI

Less to do than the issue assumed, since Dialog, Menu, Switch and Toast were already migrated. I moved `CollapsibleSection` onto `Collapsible` (it was a hand-rolled `grid-template-rows` animation), added the `PreviewCard` above, and restyled everything else onto the tokens. Tooltip is still the CSS one, now with keyboard focus support and an optional multiline mode for sentence-length text.

Buttons stay a styled native `<button>`. Base UI does ship a `Button`, and swapping to it would mainly buy `focusableWhenDisabled`, so it is worth doing at some point, but it is orthogonal to the theming work and I left it out to keep this diff about tokens.

Settings, showing the Base UI collapsible sections and switches

Overview at mobile width

## The example app

`example.ts` used to drain to all-completed within a minute of booting, so the board never showed active, waiting, prioritized or waiting-children jobs and the UI was hard to evaluate. It now seeds prioritized jobs and a few flows, and keeps a trickle of new work coming, so every status is represented. The screenshots above are that data.

## Scope, and where this should land

You said this will be a major, so the new look stays the shipped default rather than reproducing the current one.

Two things I would rather you decided than have me guess at.

**Whether to point this at `master` at all.** As it stands, merging this puts a full visual rewrite on the default branch, which means `master` is not really releasable again until everything that rides along is finished: the 19 docs screenshots, the Select migration, and whatever falls out of review. If you would rather keep `master` free for patch releases in the meantime, I am happy to retarget this at a long-lived branch, something like `next` or `v9`, and keep stacking the rest of the rewrite on top until you want to cut the major. Then the churn lands in one deliberate merge on your schedule instead of dribbling into the default branch.

**Whether to split it.** Your issue asked for small PRs, one step or one component each, and this is not that. The natural seam is the token layer, `uiConfig.theme`, the docs and the restyle as one change, with the card rework and the overview counts as a second. That splits cleanly either way, onto `master` or onto a feature branch.

## Not in here

Tailwind, any IA or navigation change, migrating the native selects to Base UI Select, and reshooting the docs site screenshots. Those 19 images still show the old UI and want redoing once the look is settled, which feels like its own PR.

A per-queue sparkline on the cards when `showMetrics` is on would be nice, but both metrics endpoints are per-queue, so 23 queues on a 5 second poll means 23 requests per tick. That needs a bulk endpoint first, so I left it alone.

## Testing

UI suite is 111 tests across 20 suites, api is 141, plus the express contract suite, a `tsc --noEmit` pass, oxlint and oxfmt, all green locally. New specs cover the card figures, the pulse segments, the empty-queue case and the per-status aggregation behind the overview counts, including the Bull case where a queue reports fewer statuses than it carries counts for.

Screenshots are hosted on a branch of my fork so they stay out of this diff.


felixmosh:
This will be Major version :]

felixmosh:
Screenshot 2026-07-29 at 16-48-40 —
I don't think that there will be much room to place 5-6 statuses here.

Stosiu:
Pushed the changes from your review. What moved since you looked:

**The cards.** You were right that five or six statuses do not fit there. Measured against the narrowest card, a 320px grid column leaving 288px of content: three labelled figures need 240px, five need 400px, six need 463px and all eight need 678px, with "waiting children" alone taking 109px. Dropping the labels for a coloured dot plus number does not rescue it either, six statuses still want 306px.

So the card now shows the composition strip plus at most two figures, active and failed, each only when non-zero. The full per-status breakdown moved into a hover panel on the strip, styled like the chart tooltips. It is built on `PreviewCard` rather than `Tooltip`, because a tooltip popup closes the moment the pointer enters it and the rows needed to be links. Those rows also replace the old 6px-tall bar segments as click targets, which were never really hittable.

**The overview summary strip is gone.** It was a read-only row of board totals above the grid. The counts now sit on the status filters instead, which were already links that filter the view, so the number is where you act on it rather than in a panel of its own. Removing it also let the `DASHBOARD.SUMMARY` keys I had added come back out of all eleven locale files.

**`example.ts` now looks like a running system.** Every queue used to drain to all-completed within a minute of boot, so the board never showed active, waiting, prioritized or waiting-children jobs, which made the UI genuinely hard to judge. It now seeds prioritized jobs and a few flows and keeps a trickle of work coming, so every status is populated. All the screenshots in the description are reshot against that data, including one of the hover panel with six statuses in it.

**A correction to something I wrote earlier.** The description used to claim Base UI has no button primitive. That is wrong, `@base-ui/react/button` exists. Moving our `Button` onto it would mainly buy `focusableWhenDisabled`. I have left it out as orthogonal to the theming, but it is a small change if you want it folded in.

**Tests.** 111 across 20 suites now. The new ones cover the card figures, the pulse segments, the empty-queue case and the aggregation behind the overview counts, including the Bull case where a queue reports fewer entries in `statuses` than it carries values in `counts`.

Two open questions in the description under "Scope, and where this should land". The one I would most like your read on: do you want this pointed at `master`, or would you rather I retarget it at a long-lived branch like `next` so `master` stays releasable while the rest of the rewrite settles? Merging a full visual rewrite into the default branch means it is not really shippable again until the docs screenshots and the remaining migrations are done too, and I would rather not park your release train for that.

felixmosh:
Some answers, 
1. Let's open a v9 branch.
2. We will need to work on it with small PR's as possible, so I'll be able to review it quickly.


Some insights regarding the design, first of all it looks slick!

1. 
Screenshot 2026-07-30 at 13-57-52 —
hard to see which state is active.
2. 
Screenshot 2026-07-30 at 14-00-15 —
the top shadow is cut by the sticky header

3. 
Screenshot 2026-07-30 at 14-04-00 —
hard to read

4. 
Screenshot 2026-07-30 at 14-05-16 —
there is no "connection" between the active state & hover state color wise. there are 2 separate colors

Stosiu:
Pushed fixes for all four.

Your 1 and 4 turned out to be one bug. Hover was a neutral wash and selection was either a different neutral or a very faint 8% primary tint, so the two states had nothing in common and hover often ended up louder than the thing that was actually selected. Everything selectable now derives both states from `--primary`: hover is a 7-8% wash, selected is 14-16% plus primary coloured text, and hovering something already selected goes to 22-24%. That covers the job detail tabs, which were the worst case since they were plain `--muted` on a card of nearly the same value, plus the status filters and the pagination. The active filter's count badge picks up the tint too.

The clipped card in 2 is not a shadow. The first card's top edge sat at y=96.5 while the sticky bar's box ended at y=102.9, so the bar's opaque background was slicing the card's top border off. There is a 12px gradient scrim under the bar now, so content dissolves into the page instead of meeting a hard edge. Its height matches the gap the bar already leaves below itself, so at scroll top it lands on empty background and tints nothing.

For 3, `--status-paused` was the weakest token in the set: 4.1:1 against the hovered sidebar row in dark and 4.45:1 in light, both under AA. It is `#a29e96` and `#6b6560` now, which gives 5.5 to 6.8:1 in dark and 5.0 to 5.3:1 in light. Same hue, just recalibrated.

Unrelated, but I also caught a dev only edit that had leaked into the branch: `rsbuild.config.mts` was proxying to my local port instead of 3000. Fixed in the same commit.

On the v9 branch, that one is on you. I only have pull rights here, and a PR base has to exist in this repo, so I cannot create it. `git push origin master:v9` is all it takes. Once it is there I will retarget this PR at it, which keeps the thread and the review history, and since v9 would come off master the diff is unchanged.

On the small PRs, I would rather not split what is already here. The token layer and the restyle are one change in practice, since every module had to move onto the tokens for the tokens to mean anything, and cutting it in two would mostly give you an intermediate state that is half migrated rather than two independently reviewable pieces. So this one lands as it is, just retargeted, and everything after it comes as small granular PRs, one step or one component each, working through what is left: the Select migration, the Button move onto Base UI, the docs screenshots and the rest.

felixmosh:
v9 branch is open

Stosiu:
@felixmosh should be good to mere 🎉

felixmosh:
@Stosiu I'll review it ASAP.

There is something that is needed to be changed in order to support bullmq v6, maybe we can add it to the major

Stosiu:
@felixmosh, just FYI: I have five more PRs queued up after this one to migrate the remaining smaller items, completing the full migration of all elements

felixmosh:
Screenshot 2026-08-04 at 18-02-00 —
Screenshot 2026-08-04 at 18-02-29 —
when there are actions, the "fade-out" overlaps the first card

missing hover on dropdown (wrong mapping).
image

button hover colors are not consistant
image
Screenshot 2026-08-04 at 18-05-05 —
Screenshot 2026-08-04 at 18-05-34 —
image


How to open it on mobile?
Screenshot 2026-08-04 at 18-08-05 —


Not related, but mobile menu is missing `Metrics history`
image

Stosiu:
All 5 fixed, fast-forward on top of what you reviewed.

1. **Fade over the first card.** It was hung below the sticky bar, so with an actions row it landed on the card. It is the bar's own bottom padding now.

Sticky bar with an actions row, first card intact

2. **Dropdown hover.** Wrong mapping, as you said: in dark, `--accent` (#20293a) is near identical to `--popover` (#202834), so there was nothing to see.

Menu item hover

3. **Button hovers.** Same root cause, plus half the board hovered on that neutral while pagination and the tabs hovered on a primary tint. Everything selectable now uses `--state-hover` / `--state-selected` / `--state-selected-hover`, 8/16/24% of `--primary`, with `--sidebar-state-*` off `--sidebar-primary`. Sidebar included, its hover and active were the same colour too. `--accent` stays in the contract: shadcn ships it solid, so a pasted palette would bring the bug straight back.

Button hover next to the selected tab, page and job tab

4. **Opening it on mobile.** Base UI's PreviewCard is `mouseOnly`. A touch press toggles it now, mouse presses left alone so a click does not close it under the pointer.

Breakdown opened by tapping the strip

5. **Metrics history in the mobile menu.** Added, plus Schedulers, which was missing for the same reason. Same gating as the sidebar.

Mobile menu with the nav entries

One extra, not from your list: the status tab fade was always on, so the first tab sat dimmed even at scroll 0. Each edge fades only while there are tabs past it.

Status tabs at the start of their travel

210 tests across 31 suites green, tsc + oxlint + oxfmt clean, both themes checked by hand at desktop and 430px.

#1316 (bullmq v6) is pointed at master. Retarget it at v9 so it rides the major?

Ready for another round when you have time.

felixmosh:
This looks much more consistant.

One gap, remain 

Screenshot 2026-08-07 at 12-18-17 —
can use this tooltip design
Screenshot 2026-08-07 at 12-20-08 —

I've tried to change colors, currently same value `accent` is spread across many other values, such as `ring`, `sidebar-ring` etc...
I think that it is better to map them to the same value with `var`, and who ever what to change will set the specific theme entry.

so by default, 

```
--ring: var(--primary);
--sidebar-primary: var(--primary);
 --sidebar-ring: var(--primary);
 
 etc
```

so, changing the `primary` will reflect on all places... WDYT?
If someone wants to change `--ring` to something other that primary, it is possible.
The idea is to make theming as simple as possible, and derive other values from a small set of tokens.

I don't see anyone customizing 20 + entries.

Regarding #1316, since it is backward compatible, I've marged (& released) it to master.

Stosiu:
@felixmosh, I have a busy week ahead, but I'll review this later in the week

Stosiu:
Both points done, fast-forward on top of what you reviewed.

**The tooltip.** It was an inverted pill, `--foreground` fill with `--background` text, so it was the one floating thing on the board that ignored the palette. It sits on the chart tooltip's surface now, and the two shadow recipes the board had grown collapsed into one `--shadow-popover`, so menus, hover panels, toasts and both chart tooltips are a single layer.

The plain tooltip next to the chart tooltip, same surface

**Deriving from a small set.** `theme.css` is three blocks now: light values, dark values, and a derived block that restates the rest.

```css
--ring: var(--primary);
--sidebar-primary: var(--primary);
--sidebar-ring: var(--primary);
--card-foreground: var(--foreground);
--chart-1: var(--status-completed);
/* ... */
```

Twenty repeated literals gone. Each name stays individually overridable and an override wins, since `uiConfig.theme` writes into `:root` and `body.dark-mode` and both outrank that rule. Below is the board with nothing set but `primary`, one value per theme. Status colours stay put.

The same board with only primary overridden

One subtlety, since it is the way this file breaks silently: a custom property that references another is substituted where it is declared, and `.dark-mode` sits on `body`, so `--ring: var(--primary)` under `:root` alone would freeze the light primary. The derived block is on `:root, .dark-mode` for that reason, with a test covering it.

**Not from your list:** the example app seeds four job schedulers on boot, so the Schedulers view is visible on a fresh checkout instead of only after hitting `/add-scheduled-job` by hand.

Schedulers view with the seeded entries

Ready for another look when you have time.

Stosiu:
@felixmosh ^^

felixmosh:
Can you explain this issue a bit?

> One subtlety, since it is the way this file breaks silently: a custom property that references another is substituted where it is declared, and .dark-mode sits on body, so --ring: var(--primary) under :root alone would freeze the light primary. The derived block is on :root, .dark-mode for that reason, with a test covering it.

Stosiu:
`var()` is substituted where the declaration lands, and only the result inherits. Dark mode is a class on `body`, one level below `html`, so a derived token declared on `:root` alone resolves against the light `--primary` and inherits that literal down. `body.dark-mode` redefining `--primary` afterwards changes nothing, since `--ring` was already substituted upstream.

```css
:root      { --primary: #0f6f64; --ring: var(--primary); }  /* resolves here, on html */
.dark-mode { --primary: #3ecfb2; }                          /* too late for --ring */
```

Result would be half the palette (`--ring`, `--sidebar-primary`, `--chart-1..5`, the `--state-*` mixes) frozen light on a dark board. Declaring on `:root, .dark-mode` makes it apply again on `body`, where the dark `--primary` is in scope, so it re-resolves per theme.

Worth a comment because it fails silently: no invalid value, no warning, just wrong colours until you toggle. The test walks the parsed stylesheet and fails any declaration containing `var()`/`color-mix()` that isn't declared for both themes, so adding one to the wrong block gets caught.

Moving `.dark-mode` to `html` would fix it at the source, but that changes the public class contract for anyone targeting `body.dark-mode`. Say the word if you'd rather have that.

felixmosh:
Ha, thank you for elaborating regarding this.

It should be 
```css
:root      { --primary: #0f6f64; --ring: var(--primary); }  /* resolves here, on html */
.dark-mode { --primary: #3ecfb2; --ring: var(--primary); }   
```

felixmosh:
I'll release this major version, it looks AWSOME, thank you.

We will need to update images + docs.

PR #1043, opened 2025-12-05
Add obliterate functionality for queue management

IgnisDa:
Implement the obliterate feature for queue management, allowing complete deletion of queues. Update translations for the new functionality in multiple languages and add comprehensive tests to ensure reliability.

Fixes #1042.

Issue #1042, opened 2025-12-05
Allow obliterating a queue


IgnisDa:
We have a lot of old queues in our application that need to be cleaned up. I was wondering if a button to obliterate a queue could be added to the dashboard. Reference: https://github.com/OptimalBits/bull/issues/1437#issuecomment-879134768

I suppose it should also ask for password confirmation since this will be a destructive action.

felixmosh:
It can be done, will you make a PR for it?
I see that bullmq have the same function.

felixmosh:
It should be inside the queue 3 dots actions.

Regarding password protection... We sould check how it can be implemented globally, or maybe it is not the board responsibility....
Like login, access control etc

IgnisDa:
@felixmosh When can we expect a new release for this?

PR #1307, opened 2026-07-31
feat: force option for obliterating a queue with active jobs

madicnikola:
## Problem

Obliterating a queue that still has active jobs fails. Pausing a queue (which the board requires before obliterating) stops new jobs from being picked up, but the jobs a worker already holds keep running, and both Bull and BullMQ refuse to obliterate while any job is active unless `force` is passed.

The board always called `obliterate({ force: false })`, so there was no way to get past it from the UI, and the refusal surfaced as a bare `Internal server error` toast that said nothing about active jobs.

> Internal server error
> Cannot obliterate queue with active jobs

## Change

- **UI** — the obliterate confirm now carries a **Force obliterate** checkbox, unchecked by default. Its state is sent as `{ force }` in the obliterate request body.
- **Confirm plumbing** — `openConfirm` accepts an optional `checkbox` (`{ label, description, defaultChecked }`) and resolves with `{ checked }`, which `getConfirmFor` passes to the action. Every existing caller keeps its current behaviour: no checkbox, `checked: false`. The checkbox state is created fresh per open, so ticking it and cancelling does not leak into the next confirm.
- **API** — `BaseAdapter.obliterate` now takes `ObliterateOptions` (`{ force?: boolean }`), forwarded by the Bull, BullMQ and BullMQ Pro adapters. The handler reads `force` from the body and still requires the queue to be paused.
- **Error** — the "there are active jobs" refusal is now a `409` with `ERRORS.QUEUE_HAS_ACTIVE_JOBS` plus a detail key telling the user about the checkbox, instead of a 500. This follows the same shape as `ERRORS.JOB_IS_ACTIVE` from the job-clean path.

New translation keys (`QUEUE.ACTIONS.CONFIRM.OBLITERATE_FORCE`, `..._DESCRIPTION`, `ERRORS.QUEUE_HAS_ACTIVE_JOBS`, `..._DETAILS`) are translated in all eleven locales.

## Tests

`packages/api/tests/api/obliterate-queue.spec.ts` — two new cases driven by a real worker holding a job active: a `409` with the translated keys when force is not passed, and a successful obliterate with `{ force: true }`.

`packages/ui/tests/hooks/useQueues.spec.tsx` — the confirm offers the force checkbox; unticked sends `force: false`, ticked sends `force: true`, cancelling sends nothing.

```
yarn workspace @bull-board/ui test        # 25 suites, 179 tests pass
cd packages/api && yarn test              # obliterate-queue.spec.ts: 8/8 pass
```

(The `entry-point`, `ui-base-path` and `metrics-*` suites fail identically on a clean `master` in my environment — unbuilt `ui/dist` — so they are untouched by this change.)

Related: #1042, #1043.


packages/ui/src/utils/getConfirmFor.ts
```diff
@@ -1,18 +1,26 @@
...
+    description: string;
+    checkbox?: ConfirmCheckbox;
+  }) => Promise<ConfirmResult>
 ) {
   return function withConfirmAndFn(
-    action: () => Promise<any>,
+    action: (result: ConfirmResult) => Promise<any>,
     description: string,
-    shouldConfirm: boolean
+    shouldConfirm: boolean,
```
felixmosh: Please group these options into an object.
put the `description` into the same new options param.

packages/ui/src/components/ConfirmModal/ConfirmModal.tsx
```diff
@@ -1,20 +1,42 @@
...
+export interface ConfirmResult {
+  /** State of the confirm's checkbox, `false` whenever it did not render one. */
+  checked: boolean;
+}
+
 export interface ConfirmProps {
   open: boolean;
   title: string;
   description: string;
+  checkbox?: ConfirmCheckbox;
```
felixmosh, line 26: Instead of using a checkbox to mark the `force`, maybe it is better to open a new confirm (in the case that there is a running job), which when confirmed, it will make the same request with `force`.

I don't see any good reason to add a checkbox for this case, inside a Confirm modal.

WDYT?

felixmosh (changes requested):
Thank you for this PR, 🙏🏼

packages/ui/src/utils/getConfirmFor.ts
madicnikola: Done — `withConfirmAndFn` now takes a single `ConfirmActionOptions` object (`description`, `shouldConfirm`, `checkbox`), and every call site in `useQueues`, `useJob` and `useJobSchedulers` was updated.

packages/ui/src/components/ConfirmModal/ConfirmModal.tsx
madicnikola: I'd like to make the case for keeping the checkbox, but I've changed it so it only shows up when it is actually relevant.

`force` is a first-class option of the underlying API — `queue.obliterate({ force })` in both Bull and BullMQ — so surfacing it as an option on the obliterate action keeps the board 1:1 with what people already know from the SDK. The practical difference is for someone who already knows the queue is junk and just wants it gone: with the checkbox that is one dialog and one click, where the escalation flow costs a failed request, a second dialog and a second request for a user who was never in doubt.

Your point about it being pointless noise is fair for the common case though, so the checkbox is now conditional — it only renders when the queue actually has active jobs, which the UI already knows from `counts.active`:

```ts
const activeJobs = data?.find((queue) => queue.name === queueName)?.counts.active ?? 0;

return withConfirmAndUpdate(({ checked }) => api.obliterateQueue(queueName, checked), {
  description: t('QUEUE.ACTIONS.CONFIRM.OBLITERATE_QUEUE'),
  shouldConfirm: true,
  checkbox: activeJobs
    ? {
        label: t('QUEUE.ACTIONS.CONFIRM.OBLITERATE_FORCE'),
        description: t('QUEUE.ACTIONS.CONFIRM.OBLITERATE_FORCE_DESCRIPTION', { jobs: activeJobs }),
      }
    : undefined,
});
```

So an ordinary obliterate keeps exactly the confirmation it has today, and when the checkbox does appear it carries the same information the second confirm would have ("3 jobs are still being processed…") without the failed round trip.

The 409 `ERRORS.QUEUE_HAS_ACTIVE_JOBS` stays either way, since that is what any client that does not send `force` gets instead of a bare 500.

If you still prefer the two-step confirm, say the word and I'll switch it — the `cleanJob` → `removeJobScheduler` flow is a good precedent for it and it is your call.

madicnikola:
@felixmosh ready for another look 🙏

- **Options object** — `withConfirmAndFn` now takes `{ description, shouldConfirm, checkbox }` instead of positional args; all call sites in `useQueues`, `useJob` and `useJobSchedulers` updated.
- **Checkbox** — kept, but now only rendered when the queue actually has active jobs, so a normal obliterate keeps the plain confirmation. Reasoning in the thread above; happy to switch to the two-step confirm if you still prefer it.
- **Rebased** onto current `master` (the `BullMQProAdapter` conflict was the `withSnapshotReset` refactor — resolved in favour of the wrapper) and added the new `ru-RU` strings.

`yarn workspace @bull-board/ui test` → 26 suites / 186 tests pass; `obliterate-queue.spec.ts` → 8/8.

felixmosh:
Thank you for this fix & patience , ☺️

PR #1309, opened 2026-08-01
chore: support ioredis v6

Venatum:
## Support ioredis v6

Widens the `ioredis` range to allow **v6** alongside v5. v6 has only two breaking changes: Node.js 20+ (already met by CI) and RESP3-by-default. RESP3 keeps `replyMappingMode: "legacy"`, so `hgetall` / `hmget` / `pipeline` / `multi` / `scan` return the same RESP2-compatible shapes as v5 — no reply-parsing changes needed. All direct ioredis usage lives in `packages/metrics`.

### Changes

- **Versions** — `@bull-board/metrics`: peer `^5.0.0` → `^5.0.0 || ^6.0.0`, devDep → `^6.0.0`, `engines.node >=20`. `@bull-board/api`: devDep → `^6.0.0` (tests only; no runtime import, so no `engines` floor on the published lib).
- **RESP2 guardrail** — the three metrics classes that open their own connection default the options path to `new Redis({ protocol: 2, ...opts.connection })`, for exact v5 wire parity and Redis < 6 support. Applied only when the caller passes options (never an injected `Redis`); an explicit caller `protocol` still wins.
- **v6 typing fix** — `LatencySampler` passes `zrange` range args as strings (`'0'`/`'-1'`); v6 types the `stop` arg as string-only. Redis coerces either way, behavior unchanged.
- **Tests / docs** — new `protocolGuardrail.spec.ts` (default, override, injected passthrough); README note that a shared client must come from the same `ioredis` install.

### Compatibility

Fully backward compatible: the metrics peer range still accepts ioredis v5, and the `protocol: 2` default preserves existing v5 wire behavior. Consumers pick a single ioredis major via the peer dependency.

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -51,7 +51,7 @@
     "@types/supertest": "^2.0.16",
     "bull": "^4.16.5",
     "bullmq": "^5.80.9",
-    "ioredis": "^5.11.1",
+    "ioredis": "^6.0.0",
     "jest": "^30.4.2",
     "supertest": "^7.2.2",
     "ts-jest": "^29.4.11"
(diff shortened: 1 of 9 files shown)


felixmosh (approved):
Thank you for this PR 🙏🏼

PR #1316, opened 2026-08-03
feat(api): support BullMQ v6

### LINKED ISSUE #1315: BullMQ v6
Hi, is there expected support for v6 of BullMQ? Installing results in:

```sh
npm warn using --force Recommended protections disabled.
npm warn ERESOLVE overriding peer dependency
npm warn While resolving: @bull-board/api@8.5.0
npm warn Found: bullmq@6.0.5
npm warn node_modules/bullmq
npm warn   bullmq@"^6.0.5" from the root project
npm warn
npm warn Could not resolve dependency:
npm warn peerOptional bullmq@"^5.79.2" from @bull-board/api@8.5.0
npm warn node_modules/@bull-board/api
npm warn   @bull-board/api@"8.5.0" from @bull-board/fastify@8.5.0
npm warn   node_modules/@bull-board/fastify
npm warn   2 more (@bull-board/ui, the root project)
```

Stosiu:
Closes #1315.

The peer range was the reported symptom, but widening it alone would have shipped a broken board. Three things behind it break on v6, and the worst one breaks quietly.

## What v6 changed

`Queue#client` is gone. Raw Redis access moved onto the backend, `queue.getBackend().client`, and only the Redis backend has one. So `getRedisInfo()` threw and `/api/redis/stats` returned a 500.

The same removal took the flow tab down with it. `getClient()` returned undefined, so `providers/flow.ts` called `WeakMap.set(undefined, producer)` and threw. Every job, 500. Had the WeakMap tolerated it, the next line would have built a `FlowProducer` on `connection: undefined` and quietly connected to localhost, which is worse.

Then `paused`, which is the one that worried me. It is gone from `JobType` and from `getJobCounts()`, and a paused queue's jobs are stored as `waiting`. Nothing throws. The adapter kept advertising the status, so you got a tab that could only ever be empty, and `handlers/queues.ts` reading `counts[statuses[0]]` was one click from `NaN`.

That mix is probably why the board looks fine on a casual upgrade. The queue list and job views genuinely are fine.

## The fix

Both client bugs turn out to be the same bug, so they collapse into one resolver. It probes for the capability rather than sniffing a version, the way `promoteAll` and `getGlobalConcurrency` already do in that file:

```ts
private async resolveRedisClient(): Promise<RedisClient | null> {
  const queue = this.queue as unknown as VersionedQueue;
  if (typeof queue.getBackend === 'function') return (await queue.getBackend()?.client) ?? null;
  return (await queue.client) ?? null;
}
```

`getStatuses()` and `getJobStatuses()` gate `paused` on the same probe. The tab disappears on v6 and those jobs show under Waiting, where they actually are. The queue-level banner still tells you the queue is paused. BullMQ Pro is on v5, so it keeps the tab.

## PostgreSQL

v6 can also run on Postgres, and I did not want that to mean an error panel. The backend exposes its `pg` pool, so the adapter asks it what it can answer honestly:

| | |
|---|---|
| Version | 17.10 |
| Up time | 3 days |
| Connected clients | 6 |
| Blocked clients | 0 |
| Port | 5432 |

Memory usage, peak memory, fragmentation ratio and mode are dropped rather than filled in. `pg_database_size` measures disk, and putting that under a row labelled peak memory would be worse than showing nothing. The modal retitles itself to match.

`getRedisInfo()` now returns `string | null`, with a `getDatastoreStats()` beside it that defaults to null, so no existing adapter changes. A future datastore that can answer neither gets a translatable 404 rather than a 500. New keys are translated across all eleven locales.

Not full Postgres support, to be clear. `@bull-board/metrics` still drives raw ioredis commands and wants its own PR. Everything else on the board works on Postgres: I checked add, counts, listing, pause, resume, promote, clean, drain, flows, schedulers, workers, global concurrency and metrics against a real database.

## Docs

A [PostgreSQL backend recipe](website/docs/recipes/postgres-backend.md), a supported-versions section on the BullMQ adapter page, a troubleshooting entry for the Paused tab going missing, and version notes on the adapters overview, production checklist, introduction and README. I ran the recipe end to end against a packed build of this branch in a real single-version v6 project, not just under the test matrix.

## Tests

`jest.config.js` in `packages/api` becomes three projects: the existing suite on the plain `bullmq` devDependency, plus `tests/bullmq-matrix/` run once per major, with `moduleNameMapper` swapping the bare `bullmq` specifier for an npm alias. Same trick the Express adapter uses for express@4 and 5. Keeping the matrix in one directory means running that twice instead of all 24 api specs twice.

`assertResolvedMajor()` reads the major back out of `bullmq/package.json` and fails if a mapping stops applying. Without it a broken alias leaves both projects on one version and every version-specific expectation still passes, which is a nasty way to be wrong.

Postgres cases run against a real database and skip visibly without `POSTGRES_URL`. CI gains a postgres service and a `yarn typecheck:bullmq` step, on the theory that a peer major breaks types first and runtime second.

One thing I caused and then fixed: three concurrent projects oversubscribed the machine and started surfacing `Connection is closed` teardown races in unrelated specs, about one run in two. The default project alone never did it. `maxWorkers: '50%'` puts concurrency back where the suite expects it, and five consecutive full runs are clean.

## Your call

The Paused tab vanishing on v6 is user-visible. I picked that over aliasing `paused` to `waiting`, which would have preserved the v5 experience at the cost of misreporting where jobs live. Easy to switch if you disagree.

The peer widening is additive; the tab disappearing arguably is not, so I have no strong feeling about whether this is a minor.


packages/api/src/providers/flow.ts
```diff
@@ -19,7 +19,12 @@ async function getFlowProducer(queues: BullBoardQueues): Promise<FlowProducer |
   const adapter = findBullMQAdapter(queues);
   if (!adapter) return null;
 
+  // A queue on a non-Redis backend has no client to build a producer from. Passing null through
+  // would key the WeakMap on a non-object and let BullMQ fall back to a default localhost
+  // connection.
   const client = await adapter.getClient();
+  if (!client) return null;
```
felixmosh: Is that mean that Bull on postgress doesn't shows Flows?

packages/api/src/queueAdapters/bullMQ.ts
```diff
@@ -34,9 +70,51 @@ export class BullMQAdapter extends BaseAdapter {
...
+    return {
+      backend: DATASTORES.postgres,
+      version: String(row.version),
+      port: Number(row.port),
+      uptime: Number(row.uptime),
+      clients: { connected: Number(row.connected), blocked: Number(row.blocked) },
+    };
+  }
+
+  /**
```
felixmosh: Please, clean these all comments 🙏🏼

felixmosh (changes requested):

packages/api/src/providers/flow.ts
Stosiu: It didn't 😅 Fixed: on v6 the FlowProducer is now built from the queue's own backend, so flows render on Postgres too. Tests in postgres.spec.ts cover a cross-queue pg flow and a mixed redis+postgres board. Also merged master to pick up #1327, the prefix logic moved into the adapter with it.

packages/api/src/queueAdapters/bullMQ.ts
Stosiu: Cleaned 🧹

Stosiu:
Ready for another look 🙏🏼 Both threads addressed: comments trimmed, and the Postgres question turned into a feature, flows now work there. One producer per connection is kept, same as before, just keyed on the backend when there is no Redis client.

felixmosh (approved):
Thank you @Stosiu 🙏🏼

PR #1327, opened 2026-08-04
fix(api): pass queue prefix to FlowProducer in flow provider

kasp470f:
The Job Flow tree never renders when queues use a custom Redis prefix.

`getFlowProducer()` in `packages/api/src/providers/flow.ts` creates `new FlowProducer({ connection: client })` without forwarding the queue's prefix. BullMQ queues registered with a custom prefix (e.g. `queue-local` via NestJS `BullModule.forRootAsync`) store job data under `<prefix>:<queue>:<jobId>`, but the FlowProducer defaults to `bull:` and looks there instead. `getFlow()` finds nothing and returns null, so the "Job Flow" section is absent for every flow job.

The fix reads the prefix from the adapter's underlying BullMQ queue (`adapter.queue.opts.prefix`) via a new `getQueuePrefix()` on `BullMQAdapter`, and passes it to `new FlowProducer({ connection: client, prefix })`. When no custom prefix is set, BullMQ fills in its own default, so existing behaviour is unchanged.

**To reproduce**: register two BullMQ queues with a non-default prefix, create a flow with a matching `FlowProducer`, open a child job in bull-board — the flow section is missing. After this change it renders the full parent-child tree.

## Tests

Added a test case in `job-flow.spec.ts` that creates queues and a `FlowProducer` with `prefix: 'custom-prefix'`, adds a flow, and asserts that `getFlowTree` resolves the full tree through the API. The three existing default-prefix tests continue to pass.

### DIFF
--- a/packages/api/src/providers/flow.ts
+++ b/packages/api/src/providers/flow.ts
@@ -23,7 +23,8 @@ async function getFlowProducer(queues: BullBoardQueues): Promise<FlowProducer | 
   const cached = flowProducerCache.get(client);
   if (cached) return cached;
 
-  const producer = new FlowProducer({ connection: client });
+  const prefix = adapter.getQueuePrefix();
+  const producer = new FlowProducer({ connection: client, ...(prefix ? { prefix } : {}) });
   flowProducerCache.set(client, producer);
   return producer;
 }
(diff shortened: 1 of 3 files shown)


felixmosh (approved):
Thank you for this FIX 🙏🏼

PR #1347, opened 2026-08-11
fix(api): count grouped jobs, not groups, in BullMQProAdapter

### LINKED ISSUE #1346: `BullMQProAdapter.getJobCounts()` adds a count of groups to a count of jobs
**Package:** `@bull-board/api@7.1.5` — `packages/api/src/queueAdapters/bullMQPro.ts`

### What happens

On a queue using BullMQ Pro groups, the counts shown in the UI are the number of *groups*, not the number of *jobs*. A queue with 3 groups holding 45 queued jobs displays `Waiting: 3`.

### Reproduce

```js
const queue = new QueuePro('demo', { connection });
for (const groupId of ['a', 'b', 'c']) {
  for (let i = 0; i < 15; i++) await queue.add('job', {}, { group: { id: groupId } });
}
// 45 jobs across 3 groups
await new BullMQProAdapter(queue).getJobCounts();
// => { waiting: 3, ... }   expected 45
```

### Suggested fix

`getGroupsByStatus(status, start, end)` returns each group's id together with its job count, so summing those gives the job total:

```ts
const groups = await this.proQueue.getGroupsByStatus(status, 0, -1);
const jobs = groups.reduce((total, group) => total + group.count, 0);
```

Note this requires **bullmq-pro >= 7.46.3**. 

### Environment

- `@bull-board/api` 7.1.5
- `@taskforcesh/bullmq-pro` 7.46.3
- `bullmq` 5.76.6

kibertoad:
getGroupsCountByStatus() reports how many *groups* sit in each group status, so folding it into the job counts reported one job per group: a queue with 45 jobs spread over 3 groups showed "Waiting: 3". The counts also drive getPagination(), so the listing was capped at a single page.

getGroupsByStatus() returns each group's job count next to its id, so the per-status job total is the sum of those. The four group statuses are disjoint in bullmq-pro (a group is removed from the waiting set when it becomes limited, maxed or paused), so nothing is counted twice.

Counting and listing now share one briefly cached group snapshot instead of each making its own Redis round trip, which keeps a page of jobs consistent with the counts its pagination was computed from.

bullmq-pro only started returning `count` in 7.46.3; older versions fall back to counting a group as a single job rather than yielding NaN.

Fixes #1346

### DIFF
--- a/packages/api/bullMQProAdapter.d.ts
+++ b/packages/api/bullMQProAdapter.d.ts
@@ -1,6 +1,5 @@
 export { BullMQProAdapter } from './dist/queueAdapters/bullMQPro';
 export type {
-  GroupJobCountsByStatus,
   GroupStatusName,
   GroupSummary,
   GroupSummaryWithCount,
--- a/packages/api/bullMQProAdapter.d.ts
+++ b/packages/api/bullMQProAdapter.d.ts
@@ -1,5 +1,6 @@
 export { BullMQProAdapter } from './dist/queueAdapters/bullMQPro';
 export type {
+  GroupJobCountsByStatus,
   GroupStatusName,
   GroupSummary,
   GroupSummaryWithCount,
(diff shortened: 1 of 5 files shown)


felixmosh (approved):

felixmosh:
Thank you for this quick fix 🙏🏼

PR #1350, opened 2026-08-14
feat(ui): replace custom SVG icons with Lucide

andreynaz4renko:
## Summary

- Swap the hand-drawn UI glyphs for Lucide via a thin `createIcon` wrapper so icons stay a single import surface and inherit theme color through CSS `color` instead of `fill`
- Keep Redis and GitHub as custom SVGs — Lucide no longer ships brand logos
- Drop `fill: currentColor` from icon CSS modules so stroke-based Lucide glyphs pick up the existing token colors

## Test plan

- [ ] `yarn workspace @bull-board/ui test`
- [ ] Open the board in light and dark themes and scan header, sidebar, queue/job actions, menus, pagination, and toasts — icons should match surrounding text color
- [ ] Confirm Redis (stats) and GitHub (header) logos still render
- [ ] Hover/focus buttons and dropdown items and check icon color still follows the hover token


felixmosh:
Hi, @andreynaz4renko thank you for this PR...
We will release a major version with redesign + themes & token change.

Can rebase this PR on v9 branch?

andreynaz4renko:
Hi! Sure! I can rebase this PR onto the v9 branch tomorrow. @felixmosh

andreynaz4renko:
Rebased onto `v9` — this PR now targets that branch instead of `master`.

andreynaz4renko:
This change still uses the current icon token `--button-icon-fill`.
That's the token set that #1306 replaces with the new theme. Once that lands on v9, I will rebase again and point the icons at the new tokens so they can match theme colors.

andreynaz4renko:
Hi, @felixmosh! I want to make 2 more PRs: pinned queues and Schedulers section improvements (search, sorting, ended schedulers, editor). Should we do this for v9?

felixmosh:
Is it related to ui changes? If so, then yes

andreynaz4renko:
I don't know actually. So for pinned queues I want to add pin button on a Q card. When active - Q will be in "Pinned" group at the top. This is more of a new UX feature. Same thoughts for the Schedulers section.
We can make it a minor version for v9 after release

andreynaz4renko:
Something like this

CleanShot 2026-08-15 at 16 02 52@2x

felixmosh:
How would you implement it with pagination support?

andreynaz4renko:
Pins stay as an ordered list of names always at the top on the client. The Pinned group is a small set, so it would not be paginated - only the remaining queues would. Additionally, we can add a limit on the number of pins: 10 or something, but I'm not sure it's necessary.
GET /api/queues would take the pin names and return them as a separate pinned slice (always complete, in pin order), then page only the rest. Resolving pins against the full queues array is enough

felixmosh:
Not sure I understood it 🤔 

Ha, you've talked about queues not jobs...so it just a matter of ordering, no pagination at queue level.

Pay attention, 
1. currently queue list is shared between the menu & dashboard page.
2. What will happened with groups? (Maybe using grouping + queue prefix is all you need)

felixmosh:
Can you resolve conflicts? (v9 - redesign was released 🎉)

andreynaz4renko:
> Not sure I understood it 🤔
> 
> Ha, you've talked about queues not jobs...so it just a matter of ordering, no pagination at queue level.
> 
> Pay attention,
> 
> 1. currently queue list is shared between the menu & dashboard page.
> 2. What will happened with groups? (Maybe using grouping + queue prefix is all you need)

Yes, queues, not jobs. Sorry for the confusion.
No pagination, pins are just client-side ordering on the same queue list.
Pins live in browser settings (same store as the rest of the dashboard prefs), so the same ordered list drives both the sidebar and the overview.
Grouping by delimiter/prefix stays as-is. Pinned queues get a virtual Pinned group at the top of both views. Queues that already belong to a prefix group stay there as well (Emails.Welcome remains under Emails). There's a setting to also lift them out of their category if you want. Ungrouped queues always move into Pinned

andreynaz4renko:
Should we rebase this branch back to master?

felixmosh:

PR #1357, opened 2026-08-17
feat(ui): move SelectField onto Base UI Select

Stosiu:
## Problem

`SelectField` is the last native form control on the board. Everything else is a Base UI primitive already, and #1306 left the Select migration as the next step.

A native `<select>` cannot be styled past its own box, so v9 works around it with a chevron baked into a base64 data URI in `index.css`:

```css
background-image: url('data:image/svg+xml;base64,...stroke="#6b7280"...');
```

That stroke is a literal grey and stays that grey in both themes, so the one arrow on the board ignores the palette everything else reads from. The list it opens is the operating system's: unthemed, and on the schedulers page that is 24 queue names rendered as OS chrome next to a fully themed table.

Two smaller things ride along on the native element. `defaultJobTab` is typed `TabsType`, but the settings modal has always offered a `'default'` entry alongside the real tabs, so the stored value can be a string the type says is impossible; the native select swallows it. And the language setting starts as `''`, which matches no option, so `<select>` computes `selectedIndex: -1` and renders the control blank until you pick something.

## The fix

`SelectField` is Base UI's `Select`. The trigger repeats the shared control recipe from `index.css` rather than inventing a second one, so it sits next to an `InputField` at the same height, radius, padding and focus treatment.

The schedule type select above the inputs it shares a form with

The popup is the `--popover` surface with `--shadow-popover`, the same layer as the menus and the hover panels, and the rows read from `--state-hover`, `--state-selected` and `--state-selected-hover`, so hover and selection are the one hue at three strengths we settled on in round 3.

Language select open in the dark theme, selected row and hovered row

The same select in the light theme

`defaultJobTab` becomes `TabsType | 'default'`, with `useDetailsTabs` matching on it instead of `Array.includes`. Base UI's `onValueChange` hands back a `string`, which is what forced the question. The language select falls back to `i18n.language`, which `initI18n` has already narrowed to a supported language, so it shows what the board is actually rendering in.

The base64 chevron, `select` and `.select` all come out of `index.css`. Nothing else uses them: this was the last native select in the tree.

## Three things Base UI defaults got wrong here

**The popup laid itself over the trigger.** That is `alignItemWithTrigger`, and on the twelve entry language list it produces a popup taller than the modal it opens in. Off, so the popup anchors below the trigger and scrolls inside its own `max-height`. The schedulers filter, at 24 queues, is the case that needs it:

Queue filter on the schedulers page, 24 entries scrolling inside the popup

**`Select.ItemIndicator` renders only for the selected row.** Every other row lost its first grid cell and its label wrapped mid word. `keepMounted` plus `visibility: hidden` keeps the column and hides the glyph.

**The highlighted row got two focus indicators.** Base UI moves focus onto the row, so the global `:focus-visible` ring drew a box around something `data-highlighted` already marks. The ring is off on items.

One thing that is not a default: the schedulers filter carries a `min-width`. The trigger is as wide as the value it shows, where the native select was as wide as its longest option, so without a floor the filter resized every time you picked a different queue.

## Tests

229 across 35 suites. Three cover the select (keyboard selection, form submission, and the `aria-label` only case the schedulers filter uses) and two the language fallback; both of those are red against `value={language}`.

`tsc --noEmit`, oxlint and oxfmt clean, both themes checked by hand.

Still on the list from #1306: `Button` onto Base UI, and the docs site screenshots.

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -55,6 +55,7 @@
     "@tanstack/react-query": "^5.101.2",
     "@testing-library/dom": "^10.4.1",
     "@testing-library/react": "^16.3.2",
+    "@testing-library/user-event": "^14.6.1",
     "@types/identity-obj-proxy": "^3.0.2",
     "@types/jest": "^30.0.0",
     "@types/react": "^19.2.17",
--- a/packages/ui/src/components/AddJobModal/AddJobModal.tsx
+++ b/packages/ui/src/components/AddJobModal/AddJobModal.tsx
@@ -82,7 +82,7 @@ export const AddJobModal = ({ open, onClose, job, queue: queueProp }: AddJobModa
           }))}
           name="queueName"
           value={selectedQueue.name || ''}
-          onChange={(event) => setSelectedQueue(queues.find((q) => q.name === event.target.value)!)}
+          onChange={(value) => setSelectedQueue(queues.find((q) => q.name === value)!)}
         />
         <InputField
           label={t('ADD_JOB.JOB_NAME')}
--- a/packages/ui/src/components/Form/Field/Field.module.css
+++ b/packages/ui/src/components/Form/Field/Field.module.css
@@ -8,8 +8,7 @@
   margin-top: 1rem;
 }
 
-.field > input,
-.field > select {
+.field > input {
   font-size: 0.875rem !important;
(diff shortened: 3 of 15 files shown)


felixmosh (approved):
LGTM

PR #1358, opened 2026-08-17
feat(ui): move Tooltip onto Base UI

### LINKED ISSUE #1356: Queue description tooltip is clipped by the status bar, queue header, or a neighboring queue
When hovering the info icon next to a queue name, the description tooltip is clipped and cannot be read in full.
It is cut off by:
- the status bar above the queue list
- the queue card header
- a neighboring queue card
## Steps to reproduce
1. Open the dashboard with queues that have descriptions.
2. Hover the info icon next to a queue name.
3. Observe the tooltip being cut off at the top.
## Expected
The tooltip is fully visible above other UI.
## Actual
The tooltip is clipped by surrounding layout.

Image
Image

Stosiu:
Closes #1356.

## Problem

The queue description tooltip is unreadable: the sticky status bar, the group header and the next queue card all paint over it.

Tooltip is the last hand-rolled primitive on the board. It is a `::before` on a wrapper span, with the text in a `data-title` attribute, positioned by hand and drawn at `z-index: 2` inside the card it belongs to. `QueueCard`'s `.infoTip` is also `z-index: 2`, so the bubble is stacked inside the card rather than above the page. Everything that comes later in the document and stacks at the same level wins: `StickyHeader` is `position: sticky; z-index: 2`, and a neighbouring card is simply a later sibling. Nothing is clipping the bubble, it is being covered, which is why it looks cut off at whichever edge it meets first.

The `multiline` prop exists for the same reason. A hand-positioned bubble needs `nowrap` and centring for a short label, and wrapping, a `max-width` and left anchoring for a sentence, because a centred sentence on an icon near the left of a card runs off the viewport. The call site has to know which one it is passing.

## The fix

`Tooltip` is Base UI's `Tooltip`. The popup is portalled to the body, so it is no longer inside the card's stacking context, and the `z-index` on the positioner puts it above the header (99), the sticky bar (2), the menus (100) and a modal (1000/1100), and below the toaster (9999) so a notification is never hidden behind a hover hint.

Queue description tooltip clearing the group header and the neighbouring card

A job action tooltip painting over the sticky action bar

The description tooltip in the light theme

The surface is unchanged from round 3 of #1306: `--popover` fill, hairline border, `--shadow-popover`, the same layer as the chart tooltips.

`multiline` is gone. A real positioner handles the collision itself, so the two cases collapse into one `max-width: 20rem`: a short label still sits on one line, a sentence wraps. `QueueCard` drops its `multiline={true}` and nothing else changes.

## What it costs

`Tooltip.Provider` goes around the app. That is what gives every tooltip the shared 400ms open delay and a 100ms grace period, so moving between two adjacent action icons does not replay the delay.

The trigger keeps `position: relative`. `.infoTip` raises itself above `.link:after`, the overlay that makes the whole card clickable, and that `z-index` does nothing on a static element.

Escape closes an open tooltip now, which the CSS one had no way to do.

## Tests

227 across 34 suites, three of them new: opens on hover, opens on keyboard focus, closes on escape. All three are red against the old component, which put the text in an attribute rather than in the document.

`tsc --noEmit`, oxlint and oxfmt clean, both themes checked by hand.

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -55,6 +55,7 @@
     "@tanstack/react-query": "^5.101.2",
     "@testing-library/dom": "^10.4.1",
     "@testing-library/react": "^16.3.2",
+    "@testing-library/user-event": "^14.6.1",
     "@types/identity-obj-proxy": "^3.0.2",
     "@types/jest": "^30.0.0",
     "@types/react": "^19.2.17",
--- a/packages/ui/src/App.tsx
+++ b/packages/ui/src/App.tsx
@@ -1,3 +1,4 @@
+import { Tooltip } from '@base-ui/react/tooltip';
 import React, { Suspense, useEffect } from 'react';
 import { Route, Switch } from 'react-router-dom';
 import { ConfirmModal } from './components/ConfirmModal/ConfirmModal';
@@ -60,7 +61,7 @@ export const App = () => {
   }, [isMobile, sidebarCollapsed]);
 
   return (
-    <>
+    <Tooltip.Provider delay={400} closeDelay={100}>
       <Header>
         <div className="header-title-group">
           {!isMobile && <SidebarToggle />}
@@ -85,6 +86,6 @@ export const App = () => {
         </div>
       </main>
       <Toaster />
-    </>
+    </Tooltip.Provider>
   );
 };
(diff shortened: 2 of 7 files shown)


felixmosh (approved):
LGTM

PR #1359, opened 2026-08-17
fix(ui): restore the checkbox styling the token rename dropped

Stosiu:
## Problem

The **Force obliterate** checkbox from #1307 has no box. Checked, it is a bare tick in the inherited text colour, which in dark mode is nearly invisible against the modal.

`CheckboxField.module.css` never made it onto the token layer in #1306. It still reads five variables that PR deleted:

```css
border: 1px solid var(--input-border);                                        /* gone */
.checkbox[data-checked] { background-color: var(--button-default-hover-bg); } /* gone */
.checkbox:focus-visible { border-color: var(--input-focus-border);            /* gone */
                          box-shadow: var(--input-focus-shadow); }            /* gone */
.indicator svg { fill: var(--accent-color-d1); }                              /* gone */
```

A `var()` naming a property nothing declares is not an error anywhere. The declaration is dropped at computed-value time, and a shorthand takes its longhands with it, so `border: 1px solid var(--input-border)` computes to no border at all. No box, no fill when checked, no focus styling beyond the global ring. Nothing in the build, the linter or the type checker says a word about it.

Force obliterate checked, rendering as a bare tick with no box

## The fix

The replacements are the obvious ones: `--input` for the resting border, `--primary` for the checked fill and its border, `--primary-foreground` for the tick, and the focus treatment `index.css` already gives the inputs, since the two appear in the same forms.

The same dialog with the checkbox on the token layer

## A test, so it cannot come back

`themeTokens.spec.ts` walks every stylesheet under `src`, collects what they declare, and fails on any `var(--x)` without a fallback that nothing defines. Runtime-set properties are allowlisted by name, and that list is short: Base UI's `--anchor-width` and `--collapsible-panel-height`, plus the four values components measure and write onto an element themselves.

Reverting the border line above turns it red with the file and the property named.

## Three colours the scan found still hardcoded

None of them can follow a theme:

- `Modal.module.css` hardcoded `rgba(0, 0, 0, 0.85)` for the backdrop. It is `--overlay` now, same value.
- `Details.module.css` tinted the job data panel with `rgba(0, 0, 0, 0.06)`, which in dark mode is a darker patch on an already dark card. It is `--muted`, which is a sunken panel in both themes.
- The checkbox and the switch thumb each carried their own drop shadow, at 0.05 and 0.25 alpha. Both are `--shadow-control` now, one value, next to `--shadow-popover` in `theme.css`.

The switch thumb keeps its literal `#fff`, with a comment saying why: it reads against the track, which is `--input` when off and `--primary` when on, and both are dark enough in either theme.

## Whitelabel

`shadow-popover`, `shadow-control` and `overlay` are in `ThemeTokenName` and the runtime allowlist. Elevation was the one part of the board a deployment could not touch, and `shadow-popover: 'none'` is a reasonable thing to want. Documented in `ui-config.md`.

## Tests

225 across 33 suites. `themeTokens.spec.ts` gains the scan, and its existing dark-coverage assertion picks up the two new achromatic tokens. Full build, `typecheck:bullmq`, oxlint and oxfmt clean.


packages/ui/src/components/Form/CheckboxField/CheckboxField.module.css
```diff
@@ -19,12 +19,15 @@
...
+  border-color: var(--primary);
 }
 
+/* The same focus treatment the inputs get from index.css, since the two sit in the same forms. */
 .checkbox:focus-visible {
-  border-color: var(--input-focus-border);
-  box-shadow: var(--input-focus-shadow);
+  outline: none;
+  border-color: var(--ring);
+  box-shadow: 0 0 0 3px color-mix(in srgb, var(--ring) 20%, transparent);
```
felixmosh: Should we extravt this input shadow into a token (derived from --ring?

packages/ui/tests/utils/themeTokens.spec.ts
```diff
@@ -27,6 +27,40 @@ function parseRules(css: string): Rule[] {
   return rules;
 }
 
+/**
```
felixmosh: Great addition 🙏🏼

felixmosh (changes requested):
Please, remove all redundant comments, it makes me itchy 😅, if there is any important comment, put it in the docs

packages/ui/src/components/Form/CheckboxField/CheckboxField.module.css
Stosiu: Done. It is `--shadow-ring` now, derived from `--ring` in the derived-tokens block, and the inputs in index.css use it too instead of repeating the recipe. It is overridable like the other shadows and documented with them.

Stosiu:
Both points done.

The comments are out of the whole PR, the stylesheets and the spec included. Where something was worth keeping, it went into the whitelabel token docs instead.

The focus shadow is a token now: `--shadow-ring`, derived from `--ring` in the derived-tokens block, so retinting the board moves it too. The inputs in index.css use it instead of repeating the recipe, and so does the checkbox. It is overridable through `uiConfig.theme` like the other shadows and documented alongside them. Ready for another look.

PR #1360, opened 2026-08-17
docs: reshoot every asset at v9 and add a whitelabel guide

Stosiu:
Picking up "We will need to update images + docs" from the v9 release. Screenshots and docs only; the code the screenshots needed is split out into the two PRs below.

Depends on #1357, #1358, #1359, #1364 and #1365. The shots are taken against `master` with all of them applied, because each changes what a screenshot shows: Base UI selects in the settings modal, the tooltip from #1356, the checkbox in the obliterate dialog, the header no longer clipping queue titles, and the demo finally able to render schedulers, latency and a job data schema.

## Problem

Every image in the docs and the READMEs predates v9.

Two matter more than the rest. `screenshots/overview.png` and `screenshots/dashboard.png` are embedded by absolute raw URL in all eleven per-package READMEs, so they are what every `@bull-board/*` page on npm shows, and they were the oldest images in the repo. Replacing the files in place refreshes all eleven without touching a README.

There was also no whitelabel guide. Theming is the headline v9 feature and its only documentation was a `## Theming` section inside the UIConfig reference, which is where you look once you already know the feature exists.

## Assets

Every referenced screenshot reshot at 2880x1800 from the demo app, whose fixtures are seeded so the numbers hold still. 19 replaced, 6 new, 5 deleted.

The deleted ones were unreferenced: `cron-queue`, `job-failed`, `queue-detail`, `queue-info-modal-no-defaults`, and a stray `docs/public/dashboard.png`. A check that every screenshot is referenced and every reference resolves passes both ways now, which it did not before.

The new ones cover things the docs described but never showed: the default job options section, the settings modal, the overview at phone width, the Add job form driven by a schema, and the light and dark halves of the theming example.

## README

The hero is a `<picture>` that swaps on `prefers-color-scheme`, so the headline feature demonstrates itself: read the README in GitHub's dark mode and the board is dark. Below it, four linked thumbnails for schedulers, historical metrics, flows and logs, and theming.

There is a one-line table of contents under the badges now, since the feature table pushes install further down the page.

No animation. A GIF of this board runs 5 to 15 MB committed permanently to git history, cannot be theme-aware, and goes stale on the next UI change. The `<picture>` costs nothing and says more.

## The whitelabel guide

`recipes/whitelabel-theming.md`, in the sidebar under Recipes. It opens on the smallest useful change, `primary` and `radius`, shows the result in both themes, then covers setting light and dark separately, the token table, how the derivations work and how to override one, the chrome knobs that are not theming, and what the sanitiser drops.

The board with a violet primary

The `## Theming` section in `ui-config.md` now opens with a link to it and stays a reference.

## Testing

`yarn build`, `yarn lint:check`, `yarn format:check` and the docs site build all pass, and the new page renders. Every screenshot was taken from a real build rather than a mock, and both themes were checked by hand.


screenshots/dashboard.png
felixmosh, line 1: after the fix of "weird gap" at #1364, we need to re-shot this image

felixmosh (changes requested):

screenshots/dashboard.png
Stosiu: Reshot, along with its twin: `screenshots/dashboard.png` and `screenshots/overview.png` (the pair every adapter README points npm at) and the docs hero pair `dashboard-overview-dark.png` / `dashboard-overview.png`. All four come from the fixed build, so the band above the logo is gone.

Every other screenshot in this PR carries the same 18px band, since they were all shot from the same build. I would rather redo that whole set in one pass once #1364 is settled, so they come from master rather than from a branch whose approach may still change. Say the word if you want them now instead.

PR #1361, opened 2026-08-17
fix(api): keep bull repeatables armed when cleaning delayed jobs

### LINKED ISSUE #294: clean all delayed jobs can finish on inconsistent state
Hi,

Maybe I'm wrong but I notice that if you clean all delayed jobs this doesn't remove repeatable jobs. This is not an expected behavior because if you ask to bull for repeatable jobs it returns your jobs but that repeatable will never run again because the delayed counterparts are deleted. I think this is an inconsistent state for bull,
I'm not sure here what is the best things to to, maybe one of these:

- Create a new tab for repeatable jobs and remove from the delayed the repeatable counterpart, threat all repeatable at the same manner
- When delete delayed job if there is a repeatable counterpart remove that too (maybe notify the user that some delayed job are chained to a repeatable job)

This make sense for you or just I don't understand well what happen inside bull ?

Stosiu:
Closes #294.

Cleaning delayed jobs on a Bull queue deletes the run a repeatable is waiting on. The schedule stays registered and never fires again. Removing that job on its own does the same and answers 204.

BullMQ fixed this in 5.16.0: `cleanSet.lua` calls `isJobSchedulerJob` and skips the armed run, and single removal raises `JobBelongsToJobScheduler`, which `cleanJob` already turns into a 400. I bisected it, 5.15.0 is broken and 5.16.0 is not. Bull has nothing equivalent and 4.16.5 is the last release, so the guard has to live here.

`BullAdapter.clean` delegates to `queue.clean` untouched when the queue has no repeatables. Otherwise it walks the set itself with Bull's own grace rule, first of `finishedOn`, `processedOn`, `timestamp`, locked jobs skipped, and leaves the armed run behind. Identifying that run needs both halves of the pairing: `opts.repeat.key` registered, and `opts.prevMillis` equal to its `next`. The key alone also matches past runs, which are ordinary jobs and should still clean. The md5 in the job id is not the repeatable key.

`BaseAdapter.getArmedJobSchedulerId` returns null by default, so BullMQ keeps relying on its own error. Bull overrides it, `cleanJob` asks before removing, and you get the same 400 with the same `jobSchedulerId`. No new translation keys.

## Tests

`bull-scheduled-job-removal.spec.ts`, the Bull sibling of the BullMQ suite. Both central cases fail on master, the bulk clean leaving the delayed set empty and the single one answering 204. Two controls in there guard the other direction: a queue with no repeatables cleans exactly as before, and the grace window still holds.

### DIFF
--- a/packages/api/src/handlers/cleanJob.ts
+++ b/packages/api/src/handlers/cleanJob.ts
@@ -19,9 +19,24 @@ function isJobSchedulerRun(error: unknown): boolean {
 async function cleanJob(
   _req: BullBoardRequest,
   job: QueueJob,
-  _queue: BaseAdapter
+  queue: BaseAdapter
 ): Promise<ControllerHandlerReturnType> {
   const jobId = job.toJSON().id ?? 'unknown id';
+
+  // Bull deletes the run a repeatable is waiting on without complaint, stranding the schedule, so
+  // the adapter is asked up front. BullMQ answers null here and raises the error below instead.
+  const armedSchedulerId = await queue.getArmedJobSchedulerId(job);
+
+  if (armedSchedulerId) {
+    return errorResponse(400, 'ERRORS.JOB_BELONGS_TO_JOB_SCHEDULER', {
+      message: {
+        key: 'ERRORS.JOB_BELONGS_TO_JOB_SCHEDULER_DETAILS',
+        options: { jobId, jobSchedulerId: armedSchedulerId },
+      },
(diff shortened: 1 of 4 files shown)


Stosiu:
Gave this one the same comment cleanup as the others before you get to it, so the diff is code only now.

felixmosh (approved):
LGTM, thank you @Stosiu

PR #1362, opened 2026-08-17
feat(ui): let uiConfig set the default job details tab

### LINKED ISSUE #264: Feature: set default tab in BullAdapter settings

Stosiu:
Closes #264.

Viewers can already pick a job tab in Settings. What is missing is the board-wide default, for the people who never open Settings at all, which is what the issue is really about.

`uiConfig.jobDetails.defaultTab` adds it, with the same semantics as `overview.groupByDelimiter`: only a default, and a viewer who picks a tab still wins. The order is viewer preference, then the configured default, then the existing tab list, whose first entry is already status aware. A configured tab that does not apply to the job in front of you, `Timeline` outside mobile being the obvious one, falls through rather than blanking the details panel.

The catch is the persisted preference. `defaultJobTab` was non-optional and initialised to `'Data'`, and persist has no `partialize`, so everyone who has ever loaded the board is carrying `'Data'` whether they chose it or not. Read as a deliberate preference it outranks the operator's default forever, and the feature does nothing for exactly the users it is meant to help. So `'default'` becomes the initial value, and a persist migration rewrites the legacy `'Data'` while leaving any tab someone actually picked alone.

The modal has always offered a `'default'` entry and it has always worked at runtime, since `tabs.includes('default')` is false and it falls through. Only the type disagreed, and that compiled because `SelectField` extends `SelectHTMLAttributes<any>`.

**One thing worth your call**: after the migration, viewers who never touched the setting will see failed jobs open on Error rather than Data. That is what "default" means everywhere else in the tab logic, but it is a visible change. Happy to pin it to Data instead.

## Tests

`useDetailsTabs.spec.tsx` covers the resolution order both ways and the migration, including the cases it must not touch. Dropping the configured-default branch fails 3 of them, disabling the migration fails 2.


packages/api/typings/app.d.ts
```diff
@@ -520,6 +520,12 @@ export type IMiscLink = {
   url: string;
 };
 
+/**
```
felixmosh: Remove this comment

packages/ui/src/hooks/useDetailsTabs.tsx
```diff
@@ -10,16 +11,43 @@ export const availableJobTabs = [
...
 
 export type TabsType = (typeof availableJobTabs)[number];
 
+/** The viewer has expressed no preference and wants whatever suits the job in front of them. */
+export const DEFAULT_JOB_TAB = 'default';
+
+export type JobTabPreference = TabsType | typeof DEFAULT_JOB_TAB;
+
+/**
+ * The viewer's own choice wins, then the board-wide default an operator configured, then the tab
```
felixmosh: Remove this comment

packages/ui/src/hooks/useSettings.ts
```diff
@@ -36,6 +36,25 @@ interface SettingsState {
...
+ * `defaultJobTab` used to start out as `'Data'`, so every viewer who has ever loaded the board
+ * carries that value whether or not they chose it. Left alone it would read as a deliberate
+ * preference and outrank anything an operator configures through `uiConfig`, so the one value
+ * that was the old default is rewritten to mean "no preference". A tab the viewer actually
+ * picked, `'Data'` included once they have upgraded, is never touched.
+ */
+export function migrateSettings(persisted: unknown, version: number): SettingsState {
+  const settings = persisted as Partial<SettingsState> | undefined;
+
+  if (version === 0 && settings?.defaultJobTab === 'Data') {
```
felixmosh, line 44: Great solution 👍

felixmosh (changes requested):

packages/api/typings/app.d.ts
Stosiu: Removed.

packages/ui/src/hooks/useDetailsTabs.tsx
Stosiu: Removed, along with the rest of them in this PR.

Stosiu:
Comments removed, in the typings, the hooks and the spec. Ready for another look.

felixmosh (approved):

felixmosh:
can you fix the CI?

Stosiu:
Fixed. Your merge of master brought in the Base UI SelectField from #1357, whose `onChange` hands over the value rather than an event, and the settings select in this PR was still reading `event.target.value`. It now takes the value directly. Green across all seven jobs.

PR #1363, opened 2026-08-17
feat(ui): support a custom icon per miscLinks entry

### LINKED ISSUE #1276: feat(ui): support custom icon for miscLinks in setUIConfig
### Feature request

It would be great if `miscLinks` in `setUIConfig` could support a custom icon per link. I want add button with icon https://github.com/Diluka/bull-board-docker/blob/cdeb1822ee6cdc0fa3f54430a9e1b881982198d0/src/app.ts#L95

for now I have to inject html https://github.com/Diluka/bull-board-docker/blob/cdeb1822ee6cdc0fa3f54430a9e1b881982198d0/src/app.ts#L101

### Current behavior

`miscLinks` allows adding custom links (text + URL), but there is no way to define the icon shown for each link.

### Proposed API

Allow an optional `icon` field in each `miscLinks` item, for example:

```ts
setUIConfig({
  miscLinks: [
    {
      text: 'My Dashboard',
      url: 'https://example.com/dashboard',
      icon: 'chart', // or custom icon key / URL
    },
  ],
});
```

### Why this is useful

- Better visual distinction between links
- Improved UX when multiple links are configured
- Helps branding/integration with internal tools

### Backward compatibility

This can be fully backward-compatible by making `icon` optional.  
If not provided, keep the current default icon behavior.

Stosiu:
Closes #1276.

`miscLinks` entries render as bare text today, which is why the reporter ended up injecting HTML through middleware. This adds the optional `icon` you suggested in the thread:

```ts
miscLinks: [{ text: 'Logout', url: '/logout', icon: '/static/logout.svg' }],
```

A URL rather than a key, since the board's own icons are inline SVGs, and because `boardLogo.path` and `favIcon` already take a caller string straight into `src` and `href`. The dropdown has a leading icon slot already, the one the queue actions menu uses, so the image drops into it. `option.url` was an `href` before this, and an `img src` cannot run a `javascript:` URL, so nothing new opens up.

One thing worth knowing. The built-in icons take their colour from `fill: var(--muted-foreground)`, which an `img` cannot inherit, and the dropdown background follows the theme, so a dark monochrome icon disappears in dark mode. A CSS filter would fix that and wreck a colour logo, so instead it is sized to match the built-ins and left to carry its own colour, with the docs saying as much. Tell me if you would rather it sat in a neutral chip.

`alt=""`, since the link text beside it already names the destination. No new translation keys.

The component had no tests before; it has three now, and the icon one fails when the render is reverted.

**Worth deciding before this merges**: the workaround in the linked repo injects a header button rather than a menu item, so what they actually want may be a top-level entry next to Redis and Docs. This does what the issue title asks for. Say the word and I will build the other one instead.

### DIFF
--- a/packages/api/typings/app.d.ts
+++ b/packages/api/typings/app.d.ts
@@ -518,11 +518,6 @@ export type BoardOptions = {
 export type IMiscLink = {
   text: string;
   url: string;
-  /**
-   * URL or static path to an image shown before the link text. The board's own icons are inline
-   * SVGs coloured by CSS, which an external image cannot inherit, so pick one that reads on both
-   * the light and dark dropdown background.
-   */
   icon?: string;
 };
 
--- a/packages/api/typings/app.d.ts
+++ b/packages/api/typings/app.d.ts
@@ -518,6 +518,12 @@ export type BoardOptions = {
 export type IMiscLink = {
   text: string;
   url: string;
+  /**
+   * URL or static path to an image shown before the link text. The board's own icons are inline
+   * SVGs coloured by CSS, which an external image cannot inherit, so pick one that reads on both
+   * the light and dark dropdown background.
+   */
+  icon?: string;
 };
 
 /**
--- a/packages/ui/src/components/DropdownContent/DropdownContent.module.css
+++ b/packages/ui/src/components/DropdownContent/DropdownContent.module.css
@@ -33,6 +33,16 @@
   fill: var(--muted-foreground);
 }
 
+/* Caller-supplied link icons. Sized to match the built-in svgs above, but left uncoloured:
+   `fill` does nothing to an `img`, so the image has to carry its own colour. */
+.content > [role='menuitem'] > img {
+  margin-right: 0.5rem;
+  width: 1.18em;
+  height: 1em;
+  object-fit: contain;
+  vertical-align: middle;
+}
+
 .content > [role='menuitem']:hover,
 .content > [role='menuitem'][data-highlighted] {
   color: inherit;
--- a/packages/ui/src/components/DropdownContent/DropdownContent.module.css
+++ b/packages/ui/src/components/DropdownContent/DropdownContent.module.css
@@ -33,8 +33,6 @@
   fill: var(--muted-foreground);
 }
 
-/* Caller-supplied link icons. Sized to match the built-in svgs above, but left uncoloured:
-   `fill` does nothing to an `img`, so the image has to carry its own colour. */
 .content > [role='menuitem'] > img {
   margin-right: 0.5rem;
   width: 1.18em;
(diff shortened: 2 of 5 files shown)


Stosiu:
The red CodeQL check here is a GitHub blip rather than anything in the diff. Autobuild passed and the analysis step failed while uploading its results:

> An unexpected error occurred when sending a status report: No server is currently available to service your request.

The API was throwing 503s intermittently while I was opening these. Everything else is green, and CodeQL passed on the two sibling PRs from the same session. Could you re-run that job when you get a moment? I do not have the rights to.

packages/api/typings/app.d.ts
```diff
@@ -518,6 +518,12 @@ export type BoardOptions = {
 export type IMiscLink = {
   text: string;
   url: string;
+  /**
+   * URL or static path to an image shown before the link text. The board's own icons are inline
```
felixmosh: Remove this comment

packages/ui/src/components/DropdownContent/DropdownContent.module.css
```diff
@@ -33,6 +33,16 @@
   fill: var(--muted-foreground);
 }
 
+/* Caller-supplied link icons. Sized to match the built-in svgs above, but left uncoloured:
```
felixmosh: Remove this comment

website/docs/configuration/ui-config.md
```diff
@@ -46,7 +46,7 @@ All fields are optional. Defaults are applied by `createBullBoard` where noted.
 | `boardLogo.path` | `string` | — | URL or static path to the logo image (required when `boardLogo` is set). |
 | `boardLogo.width` | `number \| string` | — | Logo width (px number or CSS length). |
 | `boardLogo.height` | `number \| string` | — | Logo height (px number or CSS length). |
-| `miscLinks` | `Array<{ text: string; url: string }>` | `[]` | Extra links in the header menu (logout, etc.). |
+| `miscLinks` | `Array<{ text: string; url: string; icon?: string }>` | `[]` | Extra links in the header menu (logout, etc.). `icon` is an optional URL or static path to an image shown before the link text; the board's own icons are inline SVGs coloured by CSS, which an external image cannot inherit, so pick one that reads on both the light and dark dropdown background. |
```
felixmosh: The fact the the board is using inline svg is irrelevant here

felixmosh (changes requested):

packages/api/typings/app.d.ts
Stosiu: Removed.

packages/ui/src/components/DropdownContent/DropdownContent.module.css
Stosiu: Removed.

website/docs/configuration/ui-config.md
Stosiu: Dropped it. The docs row now just says the icon is rendered as-is, so pick one that reads on both dropdown backgrounds.

Stosiu:
Comments removed, and the docs row no longer mentions how the built-in icons are drawn. It now just says the image is rendered as-is, so pick one that reads on both dropdown backgrounds. Ready for another look.

The CodeQL job is still the upload blip from yesterday rather than anything in the diff, so it will need a re-run when you get a moment.

felixmosh (approved):

PR #1364, opened 2026-08-17
fix(ui): stop the environment badge clipping the queue title

Stosiu:
## Problem

With `uiConfig.environment` set, the badge covers the top of the queue page title.

The Charges title with its top clipped by the Demo badge

`.envBadge` is `position: absolute; top: 0` inside the fixed header, so it takes no space, and the header's vertically centred content runs underneath it. Measured on a queue page, the badge occupies y 0 to 18 and the `h1` starts at 7.7, losing the top 10.3px of a 28.8px title.

Nothing appears wrong on the overview, because the header there carries no title for the badge to land on. It only bites on a queue page.

This is a v9 regression, and mine. The badge CSS is byte-identical to v8.6.1; what changed is `--header-height`, which #1306 took from 80px to 64px. The content is centred, so removing 16px moved it up 8px, and v8's 8px gap became an 8px overlap.

## The fix

The header grows by the badge's own height when there is one, rather than squeezing a title and description that already need most of 64px.

That growth goes on `--header-offset`, a new derived token, not on the header alone. Five things offset themselves by the header height: the header, `main`, the menu, the sticky status bar and the overview's group headers. Padding the header by itself would have left the other four tucked underneath it, which is exactly the bug one level down.

`Header.tsx` sets `--env-badge-height` on the root element when a badge is configured and removes it otherwise, so `--header-offset` is plain `--header-height` on a board without one. It reads the configured `environment.fontSize`, so a larger badge reserves more.

The same header with the title clear of the badge

Header 82px, badge 0 to 18, title 25.7 to 54.5, and the menu and `main` both start at 82. The gap is 7.7px, which is what v8 had.

## Tests

Three, covering the class that hangs the reservation off, its absence without an environment, and the badge variables landing on the header where the content can read them. The first goes red against the old markup.

236 across 37 suites overall, `tsc --noEmit`, oxlint and oxfmt clean. Checked by hand at 1440px and 430px, with and without a badge.


packages/ui/src/components/Header/Header.tsx
```diff
@@ -12,23 +12,45 @@ export const Header = ({ children }: PropsWithChildren<any>) => {
...
-              '--badge-bg': uiConfig.environment.color,
-              '--badge-color': uiConfig.environment.textColor,
-              '--badge-font-size': uiConfig.environment.fontSize,
-            } as React.CSSProperties
-          }
-        >
-          {uiConfig.environment.label}
-        </div>
-      )}
+    // The badge variables live on the header rather than the badge, because the header's
```
felixmosh: Remove comments 🙏🏼

Stosiu: Gone, here and in Header.module.css, theme.css and the spec.

Stosiu:
Comments gone, in Header.tsx, Header.module.css, theme.css and the spec alike. Ready for another look.

felixmosh:
There is weird gap at the top here.
Screenshot 2026-08-18 at 10-01-08 —

packages/ui/src/theme.css
```diff
@@ -177,6 +177,8 @@
  */
 :root,
 .dark-mode {
+  --header-offset: calc(var(--header-height) + var(--env-badge-height, 0px));
```
felixmosh: this is a weird way of calculating it.

Maybe if there is a badge, it should modify the header-offset instead of setting the badge height?

WDYT? is it better approach?

felixmosh (changes requested):

packages/ui/src/theme.css
Stosiu: Agreed, and switched to that. `--header-offset` is now plain `var(--header-height)` in the derived block, and the header sets it to `calc(var(--header-height) + <badge>)` when a badge is configured, so the chained calc is gone.

One catch worth naming, since it is what the first version was working around: it has to be set on `body`, not on the root element. `--header-offset` is declared on `:root, .dark-mode`, and `.dark-mode` sits on body, so a value written to the root element is shadowed by that declaration for everything inside body. A test covers it: put it back on the root element and two assertions fail.

Stosiu:
Good catch, and it was mine: the reserved room was padding the whole header, so the logo block was pushed down and the header background showed above it.

The badge only spans the content area (it starts after the sidebar), so only the content reserves room now. On mobile it is centred over everything instead, so there the header keeps the padding and the content resets it. Measured at both widths: desktop 82px header (64 + 18) with the logo filling all of it, mobile 126px (108 + 18) with nothing clipped.

after, desktop

after, mobile

PR #1366, opened 2026-08-18
feat(ui): finish the Base UI component migration

Stosiu:
Finishes step 3 of #1293. Master already ran Dialog, Menu, Select, Switch, Checkbox, Collapsible,
Tooltip and Toast on Base UI; this is the rest of the list. Five commits, one component each.

- **Button** onto `@base-ui/react/button`. Same props, same classes, plus the `render` prop the
  tab strip below needs.
- **Job detail tabs** onto `Tabs`. They were a `<ul>` of buttons: no tab roles, no
  `aria-selected`, no arrow keys, no link from the panel back to its tab. Each tab is still the
  same `Button`, passed through `Tabs.Tab`'s `render`.
- **Metrics range control** onto `ToggleGroup`. It claimed `role="tablist"` with no tab panels
  behind it, and in 60m mode there is no latency panel to point at, so it is a segmented
  single-select rather than tabs. Gains roving tabindex and arrow keys.
- **Form fields** onto `Field`, which replaces the `htmlFor` we threaded by hand with Base UI's
  own label wiring, so every control now resolves to its label.
- **Job progress ring** wrapped in `Progress.Root`, so it reports `role="progressbar"` with
  `aria-valuenow` instead of only rendering the percentage as text.

Making the tabs controlled surfaced a bug in `useDetailsTabs`: the list was built in an effect
starting from `[]`, so `selectedTab` was `undefined` on the first render. It is a pure function of
job status and timeline visibility, so it is derived during render now.

Left alone on purpose: `StatusTabs` (NavLinks driving the URL, so Tabs is the wrong primitive),
`SidebarToggle` (a disclosure with `aria-expanded`, which `aria-pressed` would fight), and
`Pagination` (no Base UI equivalent).

Nothing changes visually. Tabs, both metrics controls and the progress ring:

Queue page

Form fields:

Settings form fields

260 tests across 41 suites, typecheck, oxlint and oxfmt green. New specs cover the tab roles,
panel wiring, arrow keys and the progressbar values, and both fail against the pre-migration
components.

### DIFF
--- a/packages/ui/src/components/Button/Button.tsx
+++ b/packages/ui/src/components/Button/Button.tsx
@@ -1,22 +1,21 @@
+import { Button as BaseButton, type ButtonProps as BaseButtonProps } from '@base-ui/react/button';
 import cn from 'clsx';
 import React from 'react';
 import s from './Button.module.css';
 
-interface ButtonProps extends React.DetailedHTMLProps<
-  React.ButtonHTMLAttributes<HTMLButtonElement>,
-  HTMLButtonElement
-> {
+interface ButtonProps extends Omit<BaseButtonProps, 'className'> {
+  className?: string;
   isActive?: boolean;
   theme?: 'basic' | 'primary' | 'default';
   compact?: boolean;
 }
 
-export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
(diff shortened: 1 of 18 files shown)


felixmosh (approved):
LGTM

felixmosh:
Thank you @Stosiu 🙏🏼

PR #1372, opened 2026-08-19
chore(deps): bump zustand to v5

Stosiu:
Bumps `zustand` from 4.5.7 to 5.0.15 in `@bull-board/ui`.

v5 dropped the cached-snapshot layer, so selectors returning a fresh object loop. Nine call sites did that; eight now use `useShallow` and `useJobFlow` just selects the scalar.

`useShallow` is not enough for `Menu`, `MenuTree` and `OverviewControls`. They selected only the store's actions, which are stable references, and called `isMenuOpen(path)` during render. v4 re-rendered them on every store update because the selector object was always new; with `useShallow` the slice is always equal, so collapsing a group did nothing. They now select the derived booleans instead. There was no coverage for expand/collapse at all, so the suite stayed green on a broken menu; the added tests fail without the fix.

Last commit widens the `bullmq` peer range in the private `test-utils` workspace to `^5.79.2 || ^6.0.0`, matching `@bull-board/api`. No install changes today, it just stops the range going stale.

### DIFF
--- a/packages/test-utils/package.json
+++ b/packages/test-utils/package.json
@@ -9,6 +9,6 @@
     "@bull-board/api": "9.3.1"
   },
   "peerDependencies": {
-    "bullmq": "^5.0.0"
+    "bullmq": "^5.79.2 || ^6.0.0"
   }
 }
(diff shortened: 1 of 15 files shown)


felixmosh (approved):

felixmosh:
Thank you 👍

PR #1376, opened 2026-08-22
feat(ui): let settings hide the environment badge

andreynaz4renko:
The environment badge is useful, but I wanted a way to hide it without changing server config. This adds a toggle in Settings that stores the preference in the browser.

The toggle only shows up when an environment is actually configured.


felixmosh:
But u can remove it from uiConfig, no?

andreynaz4renko:
> But u can remove it from uiConfig, no?

Yeah, you can drop it from uiConfig, but that needs a redeploy and applies to everyone who uses the board. The toggle is just a per-browser preference.

felixmosh (approved):
Thank you

PR #1377, opened 2026-08-22
feat(ui): follow system theme from a settings dropdown

andreynaz4renko:
Replaces the dark mode switch with a theme dropdown: light, dark, and "as in system".

If you pick the system option, the board follows the OS color scheme and updates when it changes. An existing dark mode preference is kept as light or dark.


felixmosh (approved):
Thank you 👍

Issue #780, opened 2024-06-26
Feature Request: Granual permission job system


renarsvilnis:
**Problem**
There is [`readOnlyMode`](https://github.com/felixmosh/bull-board?tab=readme-ov-file#queue-options) but that is for the whole adapter. But it applies to the particular job on creation rather than runtime. Especially as the bull-board has grown in functionality over time with recent "add job" functionality

**Suggestion**
A finer control over which users can view/edit/... particular jobs. 

It would be nice if there would be a function run per req that can determine " CRUD-like functionality per adapter be able to:
- view job
- create new jobs
- delete job
- retry job
- pause/play queues
- clear queue

**Reasoning**
We want to give access to bull-board (as an internal tool) depending on the roles of the user:
- admin developer (full access)
- developer (view permission and possible retry-job)
- PM (view jobs, to help aid in troubleshooting operation issues)

felixmosh:
Yeah, that can be really cool,
If you have some time, a PR will really help me to push this feature.

AjiMk:
Hi @renarsvilnis,

I noticed this issue and would like to work on it. Is anyone currently assigned to this or actively working on it? If not, I would be happy to take it on :).

Thanks!

felixmosh:
No one is assigned to it, it can really help use.
Pay attention for current behavior, try to support it by default (if possible)

AjiMk:
Hi @felixmosh , 

I would love to work on implementing the granular permission job system as described. Could you please assign this issue to me so I can get started?

felixmosh:
I've added queue visibilty guard feature... that can show/hide queues to users

riya-kaurav:
Hi @felixmosh I'd love to work on this if it's still open noticed it's been assigned to @AjiMk k since 2024 with no PR yet. Is it still being worked on or free for someone else to pick up? Also wanted to check are outside contributors welcome to take issues like this one?

felixmosh:
Sure, I guess that it is not active you can work on it.

All contributions are welcome 🤗.

What is your plan? How do you think to implement it?

riya-kaurav:
Thanks for the go-ahead! Here's my initial plan:

I'd add an optional `getPermissions` (or similar) config function to `BullBoardOptions` that runs per-request, alongside the existing `readOnlyMode`. It would take the request (and queue/job context) and return an object like:

```
{ view: true, create: false, delete: false, retry: true, pauseQueue: false, clearQueue: false }
```

[...]

PR #1382, opened 2026-08-24
feat(api): add before/after hooks for fine-grained request access control

riya-kaurav:
## Summary

Adds an optional `hooks` config to `createBullBoard()` for fine-grained, per-request access control — addresses #780.

Rather than a per-handler/per-queue permission system, this implements two generic hook points wired centrally in `createBullBoard` (not duplicated per server adapter), as discussed in #780:

- **`before(context)`** — runs before any API route handler. Can deny the request (`{ allow: false, status?, message? }`), short-circuiting with an error response before the handler runs.
- **`after(context, result)`** — runs after the handler, receives the result, and can modify the response body (e.g. filtering fields a given caller shouldn't see).

Both are fully optional and additive — omitting `hooks` preserves current behavior exactly.

## Example

```ts
createBullBoard({
  queues,
  serverAdapter,
  options: {
    hooks: {
      before: (ctx) => {
        if (isReadOnlyUser(ctx.request) && ctx.method !== 'get') {
          return { allow: false, message: 'Read-only access' };
        }
        return { allow: true };
      },
      after: (ctx, result) => {
        // e.g. strip sensitive fields for certain roles
        return result;
      },
    },
  },
});
```

## Implementation notes

- Wired once in `packages/api/src/index.ts` (wraps `apiRoutes` before `.setApiRoutes()`), so no server adapter package needs changes — works identically across Express/Fastify/Koa/etc.
- Denied requests use the existing `errorResponse()` / `ErrorTranslationKey` convention (added `ERRORS.FORBIDDEN`) rather than a raw string, per the repo's i18n approach. `message` (the caller-supplied reason) is passed through as the existing free-text `message` field.
- Ran `sync:locales` after adding the new key.

## Tests

Added `packages/api/tests/api/hooks.spec.ts` covering:
- before-hook deny (custom message, default message, custom status)
- before-hook allow
- after-hook modifying the response
- after-hook not running when before-hook denies
- backward compatibility with no hooks configured

**Note:** `hooks.spec.ts` occasionally shows `Connection is closed` when run as part of the full suite (isolated runs always pass). This looks like the same Redis-connection teardown flakiness reported in #1374 — unrelated to this feature's logic.

## Open questions for review

- Is `ERRORS.FORBIDDEN` the right key name/place, or should it be scoped differently?
- Happy to extend `after` to also run on the `errorHandler` path if that's useful, currently it only wraps the normal `apiRoutes`.

### DIFF
--- a/packages/api/src/hooks.ts
+++ b/packages/api/src/hooks.ts
@@ -18,11 +18,17 @@ export function wrapHandlerWithHooks(
     const context = { method, route: routePath, request: request as BullBoardRequest };
 
     if (hooks.before) {
-      const beforeResult = await hooks.before(context);
+      let beforeResult;
+      try {
+        beforeResult = await hooks.before(context);
+      } catch {
+        return errorResponse(500, 'ERRORS.INTERNAL_SERVER_ERROR');
+      }
+
       if (beforeResult && beforeResult.allow === false) {
         return errorResponse(
           beforeResult.status ?? 403,
-          'ERRORS.FORBIDDEN',
+          beforeResult.errorKey ?? 'ERRORS.FORBIDDEN',
           beforeResult.message ? { message: beforeResult.message } : {}
         );
       }
--- a/packages/api/src/hooks.ts
+++ b/packages/api/src/hooks.ts
@@ -1,0 +1,35 @@
+import {
+  AppControllerRoute,
+  BoardHooks,
+  BullBoardRequest,
+  ControllerHandlerReturnType,
+} from '../typings/app';
+import { errorResponse } from './errors';
+
+export function wrapHandlerWithHooks(
+  route: AppControllerRoute,
+  hooks: BoardHooks
+): AppControllerRoute['handler'] {
+  const originalHandler = route.handler;
+  const method = Array.isArray(route.method) ? route.method[0] : route.method;
+  const routePath = Array.isArray(route.route) ? route.route[0] : route.route;
+
+  return async (request?: BullBoardRequest): Promise<ControllerHandlerReturnType> => {
+    const context = { method, route: routePath, request: request as BullBoardRequest };
+
+    if (hooks.before) {
+      const beforeResult = await hooks.before(context);
+      if (beforeResult && beforeResult.allow === false) {
+        return errorResponse(
+          beforeResult.status ?? 403,
+          'ERRORS.FORBIDDEN',
+          beforeResult.message ? { message: beforeResult.message } : {}
+        );
+      }
+    }
+
+    const result = await originalHandler(request);
+
+    return hooks.after ? hooks.after(context, result) : result;
+  };
+}
--- a/packages/api/tests/api/hooks.spec.ts
+++ b/packages/api/tests/api/hooks.spec.ts
@@ -18,9 +18,16 @@ describe('hooks', () => {
   });
 
   afterEach(async () => {
-    for (const queue of queueList) {
-      await queue.close();
-    }
+    await Promise.allSettled(
+      queueList.map(async (queue) => {
+        try {
+          await queue.waitUntilReady();
+        } catch {
+          // ignore
+        }
+        return queue.close();
+      })
+    );
   });
 
   describe('before hook', () => {
--- a/packages/api/tests/api/hooks.spec.ts
+++ b/packages/api/tests/api/hooks.spec.ts
@@ -33,19 +33,16 @@ describe('hooks', () => {
       createBullBoard({
         queues: [new BullMQAdapter(paintQueue)],
         serverAdapter,
-        options: { hooks: { before } },
+        options: { handlerHooks: { before } },
       });
 
-      await request(serverAdapter.getRouter())
+      const res = await request(serverAdapter.getRouter())
         .get('/api/queues')
         .expect('Content-Type', /json/)
-        .expect(403)
-        .then((res) => {
-          const body = JSON.parse(res.text);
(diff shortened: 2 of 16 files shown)


packages/api/src/hooks.ts
```diff
@@ -0,0 +1,35 @@
...
+): AppControllerRoute['handler'] {
+  const originalHandler = route.handler;
+  const method = Array.isArray(route.method) ? route.method[0] : route.method;
+  const routePath = Array.isArray(route.route) ? route.route[0] : route.route;
+
+  return async (request?: BullBoardRequest): Promise<ControllerHandlerReturnType> => {
+    const context = { method, route: routePath, request: request as BullBoardRequest };
+
+    if (hooks.before) {
+      const beforeResult = await hooks.before(context);
```
felixmosh: what will happend if something goes wrong (will throw) in here?

packages/api/src/hooks.ts
```diff
@@ -0,0 +1,35 @@
...
+      if (beforeResult && beforeResult.allow === false) {
+        return errorResponse(
+          beforeResult.status ?? 403,
+          'ERRORS.FORBIDDEN',
```
felixmosh: allow to define it as well, fallback to 403 + 'ERRORS.FORBIDDEN'

packages/api/tests/api/hooks.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+        queues: [new BullMQAdapter(paintQueue)],
+        serverAdapter,
+        options: { hooks: { before: () => ({ allow: true }) } },
+      });
+
+      await request(serverAdapter.getRouter())
+        .get('/api/queues')
+        .expect('Content-Type', /json/)
+        .expect(200)
+        .then((res) => {
```
felixmosh: extract assertions from the Promise chain.

```
const res = await request(serverAdapter.getRouter())
        .get('/api/queues')
        .expect('Content-Type', /json/)
        .expect(200);

const queues = res.body.queues;

expect(queues).toHaveLength(1);          expect(queues[0].name).toBe(paintQueue.name);
```

packages/api/tests/api/hooks.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+          const queues = JSON.parse(res.text).queues;
```
felixmosh: `res.body` should have a parsed json :]

packages/api/typings/app.d.ts
```diff
@@ -514,6 +534,7 @@ export type BoardOptions = {
   uiBasePath?: string;
   uiConfig?: UIConfig;
   historyProvider?: MetricsHistoryProvider;
+  hooks?: BoardHooks;
```
felixmosh: Do you think that it is better to name it tp something with intent? 
like `handlerHooks`

felixmosh (changes requested):
Thank you for this PR, it looks good...
Just minor things to fix

packages/api/src/hooks.ts
riya-kaurav: Fixed - wrapped hooks.before() in a try/catch, returns a 500 (ERRORS.INTERNAL_SERVER_ERROR) if it throws.

packages/api/tests/api/hooks.spec.ts
riya-kaurav: Fixed - extracted assertions from the Promise chain across all tests, now using await + direct expect() calls.

packages/api/tests/api/hooks.spec.ts
riya-kaurav: Fixed - switched to res.body everywhere, removed the manual JSON.parse(res.text).

packages/api/typings/app.d.ts
riya-kaurav: Good call - renamed hooks to handlerHooks in BoardOptions and everywhere it's used.

packages/api/src/hooks.ts
riya-kaurav: Done - added an optional errorKey field to BeforeHookResult, falls back to 'ERRORS.FORBIDDEN' when not provided.

riya-kaurav:
@felixmosh addressed all the review comments pushed a new commit. Let me know if anything else needs changes!

felixmosh:
Can you check why tests are failing?

riya-kaurav:
Checked the CI logs both failures are Connection is closed from ioredis during teardown, exactly the same error I saw locally. When I run either failing test in isolation (yarn test hooks -t "test name"), it passes consistently it's only when run as part of the full suite that it intermittently fails, and a different one fails each time. This looks like the same Redis-connection teardown race you flagged separately in #1374, not something in this PR's hook logic. Happy to add extra cleanup/retry in afterEach if you'd prefer, just let me know your take.

felixmosh:
It is not related to #1374,

Currently ci is passing, there is something that changed due to this PR, can you review it?

riya-kaurav:
You're right, dug in and found it. It's not fully unrelated  hooks.spec.ts reuses the same queue name ('Paint') across 7 back-to-back tests, and unlike other spec files, these tests execute very fast (no real data ops), which increases the odds of hitting a race where the previous test's Redis connection is still closing when the next one opens under the same name. Fixing by giving each test a unique queue name, pushing shortly.

riya-kaurav:
"Found it the throwing test's queue never finished connecting before `close()` ran. Fixed by waiting on `queue.waitUntilReady()` in `afterEach`. Ran it several times locally passing consistently now. Pushed.

felixmosh (approved):

PR #1383, opened 2026-08-25
feat(api): widen the BullMQ peer range to ^5.56.0 and test the floor in CI

### LINKED ISSUE #1097: Suggestion: run CI against earlier/multiple bullmq version(s)
There have been a couple recent issues (#1096, #1074) caused by using features only available in the latest version of bullmq. It might be a good idea to lower the bullmq version being tested against. I would recommend running CI against a matrix of multiple versions spanning the full supported range of bullmq versions.

Alternate solutions:
- Reduce the version of bullmq in the `devDependencies` to the minimum supported version
- Raise the minimum supported version of bullmq (will require a major version bump)

Stosiu:
Closes #1097.

The peer range was never chosen. #1266 declared it as `^5.79.2` by copying whatever the devDependency happened to be that afternoon, so the lower bound is a claim CI has never tested, which is what this issue was opened about. It is also far narrower than the truth: `^5.79.2 || ^6.0.0` accepts 33.5% of BullMQ installs, so two thirds of the ecosystem gets a peer warning for a version that works fine.

This widens it to `^5.56.0 || ^6.0.0`, which accepts 88.3%, and makes CI prove the new bound on every commit.

## Finding the real floor

I pinned an npm alias to a series of releases and ran the whole `packages/api` suite, 224 tests, against each one, twice per version with Redis flushed in between.

| bullmq | failures | `GET /api/queues` |
|---|---|---|
| 5.60.0 and up | 0 | 200 |
| 5.58.0, 5.56.0 | 1 | 200 |
| 5.52.0 | 3 | 200 |
| 5.44.0 | 4 | 200 |
| 5.36.0 | 13 | 200 |
| 5.29.0 | 35 | 500 |
| 4.18.3 | 55 | 500 |

The cliff at 5.30 is `Queue#getJobSchedulersCount`, which the queue listing calls on every poll. Below it the board does not degrade, it stops: the endpoint the whole UI is built on 500s, and the same is true of every v4 release. Above 5.36 the picture is genuinely partial instead, individual features misbehaving on a working board.

## The one fix

The single failure at 5.56.0 and 5.58.0 was ours rather than BullMQ's. `cleanJob` refuses to delete the run a scheduler is waiting on, because removing it leaves the schedule registered and unable to fire again. `BullAdapter` answers that through `getArmedJobSchedulerId`. `BullMQAdapter` never overrode it, so the handler fell back to sniffing BullMQ's numeric error code `-8` off the thrown error, and that code does not exist before 5.59. Without it the request 500s instead of returning the 400 that names the scheduler.

The override works it out the way the schedulers view already does, from the ids BullMQ derives:

```ts
const scheduler = await this.queue.getJobScheduler(repeatJobKey).catch(() => null);

return scheduler?.next && this.schedulerRunId(repeatJobKey, scheduler.next) === id
  ? repeatJobKey
  : null;
```

The armed run is `repeat:<schedulerId>:<scheduler.next>`. Past runs of the same scheduler carry the same `repeatJobKey` and are ordinary jobs that should still delete, which the id comparison gets right and the error code never distinguished. That is what moves the floor from 5.60.0 to 5.56.0, and it makes the guard deterministic on every version rather than dependent on an undocumented number.

## Why it stops at 5.56 and not lower

The table above suggests 5.52 is three small fixes away and 5.44 is four, so I built them and measured, and the gain turned out not to be real.

Two of those failures are BullMQ storing a scheduler's `every` as a string and leaving the old `every` behind when a schedule is rewritten as a cron pattern, both of which the adapter can normalise away. The third is `FlowProducer#getFlow` ignoring a custom `prefix` and answering with a childless root between 5.44 and 5.48, which is upstream and fixed in 5.50.

The one that matters is the fourth, which is not in the table because it only shows up under load. Below 5.55, BullMQ's scheduler skips a slot whenever a worker is late, so `scheduler.next - every` names a run that never existed and `findLastRunId` cannot name the previous run. Measured over 20 runs of an interval schedule with a real worker:

| bullmq | runs with no job at `next - every` |
|---|---|
| 5.50.0 | 3/20 |
| 5.53.0 | 3/20 |
| 5.54.0 | 4/30 |
| 5.55.0 | 0/20 |
| 5.56.0 | 0/20 |
| 5.81.3 | 0/20 |

So the lowest version where the board actually works is 5.55.0, worth 88.4% against 5.56.0's 88.3%. One tenth of a point, for a fix that only papers over versions where the previous-run link is genuinely unreliable. I dropped it and left the floor at 5.56.0.

Below that, 5.36 is 91.0% and costs a `supportsGlobalConcurrency` capability plus the UI work to hide a control that currently no-ops silently, and an existence check so a `PATCH` to a mistyped scheduler id answers 404 instead of creating a schedule. Below 5.30 it is a second scheduler backend on the repeatable API. Both are written into the docs so nobody has to rediscover them.

## Keeping the floor honest

`jest.config.bullmq-floor.js` replays the whole default suite against `bullmq-v5-floor`, an alias pinned to exactly `5.56.0` with no caret so a dependency bump cannot quietly redefine what the floor means. It throws at load if that pin and the first term of the peer range disagree, so the declared bound cannot drift from the version that proves it. `tsconfig.bullmq-floor.json` joins `typecheck:bullmq` for the same reason one level up.

It runs as its own `jest` invocation from the `test` script rather than a fourth entry in `projects`. My first attempt made it a fourth project and the suite went red: it replays the default project's spec files, and two projects driving the same queue names against one Redis obliterate each other's fixtures. Sequential costs about 13 seconds.

## Docs

`website/docs/queue-adapters/bullmq.md` gains a support policy under Supported versions: the three versions CI runs, what breaks below the floor, that the floor moves down whenever a fix makes a lower version pass, and that raising it is a breaking change. That is the half of #1097 asking for a stated policy with real ranges rather than "latest". README and `CLAUDE.md` follow.

## Tests

The floor project is its own positive control. Remove the `getArmedJobSchedulerId` override and it fails with exactly the one scheduler failure it was added for, while the default, `bullmq@5` and `bullmq@6` projects stay green, which is the point: this is a version the suite could not previously see.

`yarn build`, `yarn lint:check`, `yarn format:check`, `typecheck:bullmq`, the full `yarn test` across every workspace and `yarn test:ui` are green locally against Redis and PostgreSQL.

`packages/metrics` gets no floor project. It declares no `bullmq` peer of its own and inherits the range through `@bull-board/api`, and its suite is serial, so a second full pass there costs more than it would prove.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -74,9 +74,9 @@ jobs:
         env:
           CI: true
 
-      # Peer majors break types before they break runtime, so the `bullmq` peer range is
-      # checked against both supported majors as well as exercised by the runtime matrix.
-      - name: Typecheck against every supported BullMQ major
+      # Peer versions break types before they break runtime, so the `bullmq` peer range is
+      # checked at both ends as well as exercised by the runtime matrix.
+      - name: Typecheck against the BullMQ peer floor and every supported major
         run: yarn workspace @bull-board/api typecheck:bullmq
 
       - name: Test
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -45,20 +45,34 @@ All adapter tests require Redis. `testTimeout` is set to 30 000 ms in each jest 
 
 ## BullMQ version matrix
 
-`@bull-board/api` declares `bullmq` as `^5.79.2 || ^6.0.0`, and both majors are tested on every
-run. `packages/api/jest.config.js` is a `projects` aggregate over three configs:
+`@bull-board/api` declares `bullmq` as `^5.56.0 || ^6.0.0`. Both majors and the exact lower bound
+are tested on every run. `yarn workspace @bull-board/api test` is two `jest` invocations:
+`jest.config.js`, a `projects` aggregate over three configs, followed by
+`jest.config.bullmq-floor.js` on its own.
 
-| Project | Specs | `bullmq` resolves to |
+| Config | Specs | `bullmq` resolves to |
 |---|---|---|
-| `@bull-board/api` | everything except `tests/bullmq-matrix/` | the plain `bullmq` devDependency |
-| `bullmq@5` | `tests/bullmq-matrix/` only | `bullmq-v5` (npm alias) |
-| `bullmq@6` | `tests/bullmq-matrix/` only | `bullmq-v6` (npm alias) |
+| `jest.config.default.js` | everything except `tests/bullmq-matrix/` | the plain `bullmq` devDependency |
+| `jest.config.bullmq-v5.js` | `tests/bullmq-matrix/` only | `bullmq-v5` (npm alias, latest 5.x) |
+| `jest.config.bullmq-v6.js` | `tests/bullmq-matrix/` only | `bullmq-v6` (npm alias, latest 6.x) |
+| `jest.config.bullmq-floor.js` | everything except `tests/bullmq-matrix/` | `bullmq-v5-floor` (npm alias, pinned to 5.56.0) |
 
-The two version projects remap the bare `bullmq` specifier with `moduleNameMapper`, the same
-trick the Express adapter uses for its express@4/express@5 matrix. Subpaths are remapped too, so
+The version configs remap the bare `bullmq` specifier with `moduleNameMapper`, the same trick the
+Express adapter uses for its express@4/express@5 matrix. Subpaths are remapped too, so
 `helpers.ts` can read the resolved major out of `bullmq/package.json`; `assertResolvedMajor()`
 fails the suite if a mapping ever stops applying, which is what stops the matrix from silently
 running one major twice.
+
+The floor config replays the default project's specs, so it cannot be a fourth entry in
+`projects`: two projects driving the same queue names against one Redis race each other. It also
+throws at load if `bullmq-v5-floor` and the first term of the peer range disagree, so the declared
+lower bound cannot drift away from the version that proves it.
+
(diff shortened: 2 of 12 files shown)


felixmosh (approved):
Thanx

Stosiu:
Ignore the churn on this one today, the branch is back to exactly the commit you approved and CI passed on.

What happened: I said in the description that going below 5.56 would cost three small fixes for two more percentage points, so I went and built them. They work, and the suite goes green down to 5.50. But when I ran the floor project in CI it failed one scheduler test that had passed locally, and chasing that turned up the thing the earlier measurements had missed.

Below 5.55, BullMQ's scheduler skips a slot whenever a worker is late. `scheduler.next - every` then names a run that never existed, so `findLastRunId` cannot name the previous run of an interval schedule. Over 20 runs with a real worker, 5.50 and 5.53 miss 3 times, 5.54 misses 4 in 30, and 5.55 and everything above it never miss. It only shows up under load, which is why two clean local passes at 5.50 said nothing.

So the real lowest working version is 5.55.0 at 88.4%, against 5.56.0 at 88.3%. One tenth of a point, and the fixes would have bought it by papering over versions where that link genuinely does not work. I dropped them.

Only the description changed, and only to record why the floor stops where it does rather than leaving 5.44 looking like cheap money for whoever reads this next.

felixmosh (approved):

felixmosh:
Github stuck on loadings it, and not allows me merge it, weird.

PR #1384, opened 2026-08-25
feat(api): lower the BullMQ peer floor to ^5.50.0

Stosiu:
Stacked on #1383, which you have already approved. Merge that one first and this diff collapses to the two commits it actually adds. I kept it separate rather than pushing over an approved branch.

#1383 stops at `^5.56.0` because that is where the suite went green with one fix. I said in it that reaching 5.44 would cost three changes and return two percentage points, and that I would do it in a follow-up if you wanted. Having now done two of the three, the answer came out better than I estimated: the floor lands at `5.50.0`, and the third change turned out not to be ours to make.

## The two fixes

Both are in `getJobSchedulers`, and both are corrections at any version rather than compatibility shims.

BullMQ before 5.50 stores a scheduler's `every` as a string. The board passed it straight through, so the interval column rendered `"60000"` instead of a number, and `findLastRunId` computed `scheduler.next - scheduler.every` on a string and got `NaN`, which silently cost you the previous-run link on every interval schedule.

Separately, `upsertJobScheduler` on those versions leaves the old `every` in the hash when a schedule is rewritten as a cron pattern, so the scheduler ends up storing both. A schedule is one or the other, and BullMQ resolves the ambiguity in favour of the pattern, so the board now does the same:

```ts
function schedulerInterval(scheduler: JobSchedulerJson): number | undefined {
  if (scheduler.pattern || scheduler.every === undefined || scheduler.every === null) {
    return undefined;
  }

  const every = Number(scheduler.every);

  return Number.isFinite(every) ? every : undefined;
}
```

One existing assertion moved. `swaps an interval for a pattern` read `every` back out of BullMQ's own storage, which is asserting BullMQ's behaviour rather than ours, and on an affected version it fails for a reason the board has already handled. It now checks the pattern in storage, which is right on every version, and checks the board's own listing for the interval, which is what a user actually sees.

## The third change, which is not ours

The remaining failure between 5.44 and 5.48 is the flow tab going empty for queues with a custom `prefix`. I chased it and it is BullMQ's, not ours. `FlowProducer#getFlow` ignores the prefix and answers with a childless root:

```
bullmq 5.44.0   getFlow(prefix)   children=none
bullmq 5.81.3   getFlow(prefix)   children=1
```

Same script, same Redis, same flow written by a correctly prefixed `FlowProducer`. Working around it would mean walking the dependency keys by hand instead of using `getFlow`, for the 1.0 pp between 5.50 and 5.44, so the floor stops at 5.50 where the upstream fix lands.

## Where that leaves the range

| peer range | accepts |
|---|---|
| `^5.79.2 \|\| ^6.0.0`, before #1383 | 33.5% |
| `^5.56.0 \|\| ^6.0.0`, #1383 | 88.3% |
| `^5.50.0 \|\| ^6.0.0`, this PR | 89.3% |

I would stop here. Below 5.50 the next stop worth anything is 5.36 at 91.0%, and it costs a `supportsGlobalConcurrency` capability plus UI work to hide a control that currently no-ops silently, and a real existence check so a `PATCH` to a mistyped scheduler id answers 404 instead of creating a schedule. Below 5.30 there is no job scheduler API at all, `GET /api/queues` returns 500 because the listing calls `getJobSchedulersCount`, and supporting that band means a second scheduler backend on the repeatable API. Both are in the docs so the next person does not have to rediscover them.

## Tests

The floor project is the control for both fixes. Put `every: scheduler.every` back and `bullmq@5.50.0 (peer floor)` fails on exactly the two scheduler tests the normalization was added for, while the default, `bullmq@5` and `bullmq@6` projects stay green, which is the point: these are versions the suite could not previously see. 5.50.0 passed three consecutive full runs with Redis flushed between them.

`yarn build`, `yarn lint:check`, `yarn format:check`, `typecheck:bullmq`, the full `yarn test` across every workspace and `yarn test:ui` are green locally against Redis and PostgreSQL.

Docs, README and `CLAUDE.md` follow the new floor, and the supported-versions section now names the two upstream bugs that set it and the hard cliff at 5.30, rather than just stating a number.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -74,9 +74,9 @@ jobs:
         env:
           CI: true
 
-      # Peer majors break types before they break runtime, so the `bullmq` peer range is
-      # checked against both supported majors as well as exercised by the runtime matrix.
-      - name: Typecheck against every supported BullMQ major
+      # Peer versions break types before they break runtime, so the `bullmq` peer range is
+      # checked at both ends as well as exercised by the runtime matrix.
+      - name: Typecheck against the BullMQ peer floor and every supported major
         run: yarn workspace @bull-board/api typecheck:bullmq
 
       - name: Test
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -45,20 +45,34 @@ All adapter tests require Redis. `testTimeout` is set to 30 000 ms in each jest 
 
 ## BullMQ version matrix
 
-`@bull-board/api` declares `bullmq` as `^5.79.2 || ^6.0.0`, and both majors are tested on every
-run. `packages/api/jest.config.js` is a `projects` aggregate over three configs:
+`@bull-board/api` declares `bullmq` as `^5.56.0 || ^6.0.0`. Both majors and the exact lower bound
+are tested on every run. `yarn workspace @bull-board/api test` is two `jest` invocations:
+`jest.config.js`, a `projects` aggregate over three configs, followed by
+`jest.config.bullmq-floor.js` on its own.
 
-| Project | Specs | `bullmq` resolves to |
+| Config | Specs | `bullmq` resolves to |
 |---|---|---|
-| `@bull-board/api` | everything except `tests/bullmq-matrix/` | the plain `bullmq` devDependency |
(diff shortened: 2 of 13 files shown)


Stosiu:
Folded into #1383 instead, no point making you read two diffs for one change.

PR #1386, opened 2026-08-25
feat: surface the job diagnostics BullMQ already reports

Stosiu:
BullMQ puts five facts on every job that the board already fetches and then throws away. `formatJob` builds `AppJob` out of `job.toJSON()`, and none of `priority`, `attemptsStarted`, `stalledCounter`, `deduplicationId` or `deferredFailure` survive the trip. Each one answers a version of the same support question: the job is sitting there and the board will not say why.

All five are class properties on BullMQ's `Job`, so `toJSON()` is already carrying them and this costs no extra round trip. I checked each against `bullmq@5.56.0` and against `bullmq@6`, so they hold across the whole declared peer range and on both majors. Bull reports none of them, which is why every field is optional and every render is conditional.

They render as pills beside the job name, in the slot the `group` pill already occupies. `priority N` on a board whose prioritized tab lists jobs today without saying what any of their priorities are, a value that was previously only reachable by opening the Options tab and reading the raw JSON:

`dedup <id>`, so a job that was deduplicated away is explicable rather than simply absent. The id truncates in the pill the way a group id does and sits in full in the title attribute:

`stalled N` in the failed colour for a job a worker picked up and lost, beside `N started`, the count of how many times it was picked up. Neither number is derivable from anything the board shows today, so a stalled job that recovers looks identical to one that ran cleanly:

And `will fail`, for a job carrying `deferredFailure`. That is a job BullMQ has already condemned: it renders as an ordinary waiting job today and fails the instant a worker touches it.

Its reason also fills the Error tab, which otherwise tells you "No errors recorded" about a job that is about to be nothing but an error.

That last screenshot needs a caveat. The deferred state is transient by design: `moveStalledJobsToWait` RPUSHes the condemned job back onto the wait list, which is the end workers take from, so a queue with a live worker consumes it almost immediately. To hold it still I paused the queue and set `defa` and `stc` on the job hash directly, which is the same write the Lua script makes. The rendering is real, the dwell time is not.

Nothing is emitted unless it is set, so a priority of 0, a stall count of 0 and an empty deferred failure are dropped rather than shipped as dead weight on every job in every 15 second poll. `attemptsStarted` needed more than a truthiness check: it is at least 1 on every job that has ever run, so emitting it whenever it is set would have put an extra field on every job in every response for no reason. It is sent only when it exceeds `attemptsMade`, which is exactly the case that means something, the job having been picked up more times than it recorded outcomes.

`example.ts` gains `Payments.Settlement`, a queue whose workers are killed mid-job on purpose. A worker with a three second lock is force-closed while holding a job, the lock expires, and the queue's other worker's stalled check moves the job back to wait with `stalledCounter` incremented. It settles into a steady trickle of `stalled 1` and `2 started` pills, which is the state this PR exists to make visible and which nothing in the example produced before. `seedQueue` also gains four delayed jobs carrying a `deduplication` id, so the dedup pill is reachable on any grouped queue rather than only on one you build yourself, and arrivals on the new queue are gated on the waiting count so the demo holds around a dozen jobs instead of growing all afternoon.

`guide/exploring-the-dashboard.md` gains a section on all five, placed after Queue info, making the point that none of them appear on a healthy job so the only thing the board spends room on is the case worth acting on, the same way the no-workers badge does.

Six new keys under `JOB.DIAGNOSTICS`, translated across all twelve locales rather than filled with English.

## Tests

`job-diagnostics.spec.ts` covers the round trip through `/api/queues` for a real BullMQ job carrying a priority and a deduplication id, the three fields a stalled or condemned job sets, and each omission rule including the `attemptsStarted` equality case. Reverting `formatJob` to its current shape fails the two presence tests and leaves the three omission tests green, which is the right split: they assert absence, and absence is what the old code produced.

`yarn format:check`, oxlint, the full `packages/api` suite against Redis and PostgreSQL, `yarn workspace @bull-board/ui test` including the locale sync check, and the Express contract suite are green locally.


felixmosh:
Can you fix merge issues?

Stosiu:
@felixmosh fixed

felixmosh (approved):

PR #1387, opened 2026-08-25
feat: queue rate limits, and clearing one a worker has tripped

Stosiu:
BullMQ has had a queue-wide rate limit since 5.60 and a way to clear one a worker has tripped since 5.27, and the board exposes neither. Global concurrency is already wired end to end and is the same shape of control, so this fills in the sibling.

Two separate things share the name and the PR does both, because either alone is half a feature. The configured limit is `Queue#getGlobalRateLimit` / `setGlobalRateLimit` / `removeGlobalRateLimit`: a setting, edited in a modal, living next to global concurrency. The active limit is what a worker writes when it calls `Queue#rateLimit`, readable through `getRateLimitTtl` and clearable through `removeRateLimitKey`. That one is a transient state, and it is the reason a queue with healthy workers and a full waiting list is doing nothing. There is nowhere else an operator can clear that key, which makes it the more valuable half.

A queue currently rate limited grows a badge beside the workers badge, counting down, and clicking it releases the limit:

`activeRateLimitTtl` rides along on the queue listing so the badge needs no request of its own. That is one more Redis call per queue per poll, and to more than pay for it the five per-queue reads in `getAppQueues` now run concurrently instead of sequentially; they were five awaits in a row before this and had no reason to be. `getRateLimitTtl` returns Redis PTTL semantics, so `-2` for a key that does not exist, and the adapter normalises anything non-positive to 0 while the listing swallows a failure to 0 rather than taking a queue's row down over a limiter read.

The configured limit gets a menu entry beside Set concurrency, shown only when the adapter can actually do it:

That value is fetched on open through `GET /api/queues/:queueName/rate-limit` rather than riding the 15 second poll, the way `useQueueWorkers` and `useQueueDefaultJobOptions` already do, since nothing outside the modal needs it. It also fills a row in the queue info panel:

`setGlobalRateLimit` and friends arrived in 5.60, 5.61 and 5.62. #1383 has since taken the declared peer floor to 5.56, below all three, so the adapter probes for the method rather than assuming it, and `supportsGlobalRateLimit` rides the queue listing so the UI hides the control instead of offering one that silently does nothing. The read path needs no probe: `getGlobalRateLimit?.()` follows the shape `getGlobalConcurrency` already uses. Bull has a `limiter` fixed at queue construction and no runtime setter, so `BullAdapter` keeps the base defaults, the menu entry never appears, and a `PUT` answers 400 with `ERRORS.RATE_LIMIT_NOT_SUPPORTED`. Nothing new is abstract on `BaseAdapter`: all six methods land as non-abstract defaults next to `getArmedJobSchedulerId`, so a third-party adapter keeps compiling and reports "no rate limit" rather than breaking.

Two things worth deciding before this merges. `.content` carries `padding: 1rem` and the children wrapper `.description` adds `1rem 0.5rem` on top, so every modal in the board has its body inset 8px further than its own title and footer. It is invisible until a modal holds something with an edge in it, which the two input fields here do. The fix is `margin: 0 -0.5rem` on `.description`, which cancels the inset for layout while keeping the room the `overflow: auto` scroll container wants for focus rings; it is one line and it touches every modal, so say the word and I will pull it into its own PR. The queue info screenshot above is that fix. Separately, concurrency and rate limit are siblings and could share one Queue limits modal. I kept them apart because it is a smaller diff and neither modal has to grow a mode, but I will merge them if you would rather have one.

`Notifications.Sms` gets a 5 per 30s limit in `example.ts` and a burst of arrivals that trips it, so the badge, the countdown and the release action are all reachable by running the example rather than only by contriving the state. `website/demo` gains the same on `notifications:push`, plus the two new `AppQueue` fields on `DemoQueue` and the six adapter methods on `MockAdapter`, without which it stopped typechecking. `recipes/rate-limits.md` is a new page covering both halves, registered in the recipes table and the rspress sidebar, and it says plainly that clearing a limit does not stop a worker setting it again: releasing one whose cause is still broken just sends the next job back into the same 429. Eleven new keys across all twelve locales, translated rather than filled with English.

## Tests

`queue-rate-limit.spec.ts` covers the configured limit round trip through the API, removal when the body carries no max, four rejected bodies, the ttl appearing on the listing and clearing after a release, the adapter reporting whether it can configure a limit at all, and Bull reporting no support and refusing the write. Making `supportsGlobalRateLimit` return false and `getActiveRateLimitTtl` return 0 on `BullMQAdapter` fails 7 of them and leaves only the Bull case green, which is the right split.

The floor project #1383 added earned its keep here: the six configured-limit cases assume an API that does not exist at 5.56 and went red on Node 24 the moment this branch was tested against current master. They are now gated on `Queue.prototype.setGlobalRateLimit` rather than on a version, so they skip at the floor and run everywhere else, while the active-limit cases, which only need 5.27, run at every version. The capability assertion checks the adapter agrees with the same probe, so the floor proves the probe reports false rather than merely proving the tests were skipped.

`yarn format:check`, oxlint, the full `packages/api` suite against Redis and PostgreSQL, `yarn workspace @bull-board/ui test` including the locale sync check, the Express contract suite, and `tsc --noEmit` on `website/demo` are green locally.


felixmosh (approved):
Great addition, thank u 🙏🏼

PR #1388, opened 2026-08-25
feat: show ignored and failed children in the flow view

Stosiu:
A flow parent that will never complete because a child was ignored renders exactly like one that is about to. `simplifyNode` returns id, name, progress, state, queueName and children, and BullMQ's own answer to "what happened to this parent's children" is never asked for.

Two calls fill that in, both available since 5.43 and 5.50 and so below any peer floor under discussion. `Job#getDependenciesCount({ processed, unprocessed, ignored, failed })` gives the four counts, which go on `FlowNode` as `dependencies` and only when at least one of them is non-zero, so a leaf carries nothing extra. `Job#getIgnoredChildrenFailures()` gives the reason each ignored child failed, keyed by child job key, and is only called on a node that actually has ignored children, so the common case costs one extra call per node rather than two.

The parent in those shots reads `1 pending  1 ignored`. Before this it read `Media.Video.Transcode` and a progress bar, and the fact that one of its two children had been failed and forgiven was not on the page anywhere. Hovering the ignored count shows the failure reasons those children carry. Failed and ignored take the failed and delayed colours; processed and unprocessed stay muted, since a parent working through its children normally is not news.

`simplifyNode` already awaited `getState` per node and recursed serially into that await. It now runs the children, the state and the dependency counts concurrently, so a tree of N nodes costs one more round trip per node rather than one more round trip per node in series. Both reads are guarded by a `typeof` check and a `catch`, so a Bull-backed node, an older BullMQ, or a datastore that refuses the call degrades to no counts instead of failing the whole tree. That matters here more than usual, because the flow endpoint answers for the whole tree and one unlucky node would otherwise cost the entire view.

`Job#removeUnprocessedChildren()` is the obvious action to hang off a parent that is stuck, and it is deliberately not here. It is destructive, it wants a confirm modal, and it belongs in a review of its own rather than riding along with a read-only change. Happy to follow up with it.

`seedFlows` in `example.ts` gains three `batch-report-partial` flows whose second child is named `unreliable-chunk` and carries `ignoreDependencyOnFailure`. The sim worker fails that job name every time, so the ignored state exists within seconds of booting the example rather than depending on the 18% random failure landing on the right child.

`recipes/job-logs-and-flows.md` gains a subsection under Job flows on reading the counts, including the part worth knowing: `ignoreDependencyOnFailure` is doing its job when it sets a child aside, and it also means a report can complete having quietly skipped half its inputs.

Four new keys under `JOB.FLOW` across all twelve locales. They interpolate `{{n}}` rather than `{{count}}` on purpose: `count` turns on i18next pluralisation, which would mean writing correct plural forms for twelve locales across three different plural rule sets in order to render a number in a table cell.

## Tests

`job-flow-dependencies.spec.ts` covers the counts on a parent still waiting on two children, an ignored child reported with the reason it was ignored (a real worker throwing `UnrecoverableError` against `ignoreDependencyOnFailure`), and a leaf carrying neither field. Making the dependency read return nothing fails 2 of the 3 and leaves the leaf case green, which is correct, since it asserts absence.

`yarn format:check`, oxlint, the full `packages/api` suite against Redis and PostgreSQL, `yarn workspace @bull-board/ui test` including the locale sync check, and the Express contract suite are green locally.


felixmosh (approved):
Thank you, great addition

PR #1389, opened 2026-08-25
feat: reschedule a delayed job and reprioritise a prioritized one

Stosiu:
A delayed job's card tells you exactly when it will run, and the only two things the board lets you do about that are promote it, which runs it now, or delete it. Neither helps when a nightly export needs to move two hours later because the upstream feed is late. BullMQ has had `Job#changeDelay` and `Job#changePriority` for years and neither has ever had a route.

Delayed jobs get a Reschedule action beside Promote, opening on the job's current run time, and saving calls `changeDelay` with the difference from now. Prioritized jobs get Change priority, which calls `changePriority`.

A run time in the past is not rejected, it clamps to a delay of 0, which is the same outcome as promoting the job and is what someone dragging a time backwards means. Priority is bounded at 2097151, which is BullMQ's own `PRIORITY_LIMIT` and the point past which it stopped preserving FIFO at the boundary in 5.80.1. Both endpoints are `PATCH`, matching `update-data`, which is the other route that edits a job in place rather than moving it. Rescheduling checks the state first and refuses anything that is not delayed, because `changeDelay` on a waiting job silently does nothing useful and the resulting "nothing happened" is worse than a message. Both check for the method rather than the library, so Bull answers 400 with `ERRORS.JOB_EDIT_NOT_SUPPORTED` and the buttons never render there.

One thing about `changePriority` cost me a bug and is worth knowing. It updates the job's `priority` field and its position in the prioritized set, and leaves `opts` alone. `opts` is the frozen blob the job was created with, so after a change `opts.priority` still reports the old number. I had the modal prefilling from it, which meant setting a job to 2 and reopening the dialog showed 5. `formatJob` now passes the live `priority` through onto `AppJob`, emitted only when non-zero, and the modal reads that instead. That two-line passthrough overlaps #1386, which surfaces `priority` along with four other job fields, so whichever lands second will conflict on those two lines and the resolution is to keep one copy. I matched #1386's semantics exactly, omitting when 0, so it does not matter which survives.

The second commit is the paging fix, and it is a `perf` rather than a feature, so it splits out cleanly if you would rather take it separately. `retryAll` called `queue.getJobs([status])` with no range, which is `0..-1`, so retrying every failed job on a queue with 50k of them deserialised all 50k into memory first. `cleanAll` and `promoteAll` already avoid this, `cleanAll` through `Queue#clean` and `promoteAll` through `promoteJobs`; retry was the one left doing it by hand.

I started out meaning to swap it for `Queue#retryJobs({ count, state })`, which does the whole thing in Lua, and backed out after reading `moveJobsToWait-8.lua`. It takes every id in the ZSET range and moves it, whether or not a job hash still exists behind it. The current implementation is deliberately more careful: `getJobs` drops ids that no longer resolve, which is what `BaseAdapter#getJobs` documents and what the existing "an id in the set has lost its data" test pins at `{ retried: 2, skipped: 1 }`. Switching would have pushed that orphan into the waiting list and reported it as retried. So the fetch is paged at 100 instead and the mechanism is unchanged, with each page read from the head of the set rather than walked forward, because retrying moves jobs out of it, and a set of already-retried ids stops a job that refuses to move from spinning the loop.

`guide/exploring-the-dashboard.md` gains a section on both actions, placed after Queue info, including the point that these are operational changes rather than configuration, the same caveat the schedulers section carries. `example.ts` needed nothing, since `seedQueue` already adds delayed and prioritized jobs and both actions are reachable by running it. Thirteen new keys across all twelve locales, translated rather than filled with English.

## Tests

`edit-job.spec.ts` covers a delayed job rescheduled to a requested time, a past time clamping to 0, three rejected run times, refusing a job that is not delayed, a priority change, four rejected priorities including the 2^21 ceiling, a 404, the stale-`opts` case above, and Bull reporting the edit unsupported.

`retry-all.spec.ts` gains a 150 job case that asserts more than one `getJobs` call and no page larger than 100. That assertion fails against the old single-fetch implementation while the other four cases stay green, which is the point: the behaviour is meant to be identical and only the reads change.

I also drove both actions through a real browser against the example, since the specs cover the server and not the wiring: the delay went from 1h to 3h and the priority from 5 to 2, with the prefill reading the live value both times.

`yarn format:check`, oxlint, the full `packages/api` suite across all three projects including the new 5.56 floor, `yarn workspace @bull-board/ui test`, and the Express contract suite are green locally. `changeDelay`, `changePriority` and `retryJobs` all exist at 5.56.0, so nothing here needs a capability probe.


felixmosh (approved):
LGTM, thank you

PR #1393, opened 2026-08-26
fix(metrics): give each jest worker its own redis database

Stosiu:
The metrics specs share one Redis and several of them assert on state that is global by design: the `__global__` rollup every recorder increments, and the namespace-wide SCAN that `MetricsHistoryAdmin` purges with, which `HistoryAdmin.spec` follows by asserting the namespace is empty. Parallel workers therefore delete each other's fixtures. It failed on every run here, three or four suites out of twelve.

Unique queue names cannot isolate either of those, and neither can an ioredis `keyPrefix`, which prefixes command keys but not the pattern arguments to KEYS and SCAN. The isolation has to be a separate keyspace, and a logical database is the cheapest one Redis offers.

`tests/connection.ts` now owns the connection for the package and derives the database from `JEST_WORKER_ID`, counting down from 15 so database 0 stays free for a developer's dev board. The eleven spec files import it instead of each declaring the same literal. `maxWorkers` caps the worker count at the number of databases available rather than at 1, and the helper throws if the two numbers ever drift apart.

`maxWorkers` still has to appear in both `jest.base.js` and `jest.config.js`, because Jest reads global options from the root config only and ignores them inside a `projects` entry. That is why `jest --config jest.config.default.js` was running parallel in the first place.

### DIFF
--- a/packages/metrics/jest.base.js
+++ b/packages/metrics/jest.base.js
@@ -8,4 +8,8 @@ module.exports = {
   },
   testPathIgnorePatterns: ['/node_modules/'],
   testTimeout: 30000,
+  // Every spec wipes the whole `bull-board:metrics:` namespace on one shared Redis, so two
+  // running at once delete each other's fixtures. `jest.config.js` repeats this because Jest
+  // reads `maxWorkers` from the root config only when running `projects`.
+  maxWorkers: 1,
 };
--- a/packages/metrics/jest.base.js
+++ b/packages/metrics/jest.base.js
@@ -8,8 +8,8 @@ module.exports = {
   },
   testPathIgnorePatterns: ['/node_modules/'],
   testTimeout: 30000,
-  // Every spec wipes the whole `bull-board:metrics:` namespace on one shared Redis, so two
-  // running at once delete each other's fixtures. `jest.config.js` repeats this because Jest
-  // reads `maxWorkers` from the root config only when running `projects`.
-  maxWorkers: 1,
+  // Bounds JEST_WORKER_ID to the logical databases `tests/connection.ts` maps it onto, which
+  // throws if the two ever disagree. `jest.config.js` repeats it because Jest reads global
+  // options from the root config only, ignoring them inside a `projects` entry.
+  maxWorkers: 15,
 };
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -99,6 +99,16 @@ runs everything outside `tests/bullmq-matrix/` against the plain `bullmq` devDep
 POSTGRES_URL=postgres://bullmq:bullmq@localhost:5432/bullmq yarn workspace @bull-board/metrics test
 ```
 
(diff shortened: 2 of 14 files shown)


packages/metrics/jest.base.js
```diff
@@ -8,4 +8,8 @@ module.exports = {
   },
   testPathIgnorePatterns: ['/node_modules/'],
   testTimeout: 30000,
+  // Every spec wipes the whole `bull-board:metrics:` namespace on one shared Redis, so two
+  // running at once delete each other's fixtures. `jest.config.js` repeats this because Jest
+  // reads `maxWorkers` from the root config only when running `projects`.
+  maxWorkers: 1,
```
felixmosh: instead of this you can use a Queue prefix as [jest worker id](https://jestjs.io/docs/environment-variables#jest_worker_id).

```
const queue = new Queue('my-queue', {
  prefix: `bull-test-${process.env.JEST_WORKER_ID}`,
});
```

This way, each worker works on it's "namespace" without interfearing to others.

felixmosh (changes requested):

packages/metrics/jest.base.js
Stosiu: The prefix isolates the BullMQ keys, but what collides here is this package's own namespace. `bull-board:metrics:` is a module constant in `src/keys.ts`, not something derived from the queue's prefix, so the recorder writes to the same keys whatever the queue is called.

Two things in particular survive a per-worker queue name. `MetricsHistoryAdmin` purges by SCANning the whole `bull-board:metrics:*` keyspace, and `HistoryAdmin.spec` asserts the namespace is empty afterwards, so that spec is global by definition. And `__global__` is a single rollup that every recorder increments by design, shared across all queues.

I did check ioredis `keyPrefix` before giving up on prefixing, since that would have been the cheap version. It prefixes command keys but not pattern arguments: a client with `keyPrefix: 'w3:'` writes `w3:bull-board:metrics:X`, then `KEYS bull-board:metrics:*` returns nothing. That breaks the admin's own SCAN, so it is not usable here.

What does work is giving each worker its own logical database, which is the same idea as your suggestion applied one level down. I have pushed that instead of the serial version. `tests/connection.ts` derives the database from `JEST_WORKER_ID`, counting down from 15 so database 0 stays free for a dev board, and every spec takes its connection from there rather than declaring its own. `maxWorkers` now caps the worker count at the number of databases instead of at 1, and the helper throws if the two ever disagree.

Before and after, same command, `jest --config jest.config.default.js --maxWorkers=8`: 3 suites and 14 tests failing, then 12 suites and 143 tests passing on three consecutive runs. Full `yarn test` with `POSTGRES_URL` passes 14 suites and 149 tests. It also makes a coverage run on this package possible, which the serial version does not.

If you would rather keep the change minimal, the one-line serial version still fixes the reported failure and I am happy to go back to it.

felixmosh (approved):

PR #1398, opened 2026-08-28
feat(docker): publish an official image for the standalone CLI

### LINKED ISSUE #1396: Feature: official Docker image for the standalone CLI
The CLI from #1351 already does everything an image would need: discovery by prefix, basic auth, read-only, and every flag also readable from an env var. The only missing piece is something to `docker pull`. #491 and #1010 both asked for it, and both got stale-botted before the CLI existed.

I didn't want to guess at demand, so I went and counted. People have been building this themselves for years:

| Image | Pulls | Last built |
|---|---|---|
| `venatum/bull-board` | 1,776,390 | 2026-08-27, on v9.5.0 |
| `deadly0/bull-board` | 1,455,481 | 2021-06-03 |
| `diluka/bull-board` | 44,304 | 2026-08-28 |
| ~20 other `*/bull-board` images | ~30k between them | all over the place |

bullstudio's official image is at 7,344 for comparison, and Workbench publishes to GHCR, which doesn't report pull counts at all.

The number that bothers me is `deadly0/bull-board`. 1.46M pulls, last built June 2021. Anyone pulling `latest` from it today gets bull-board as it was five years ago, and that's the version they form an opinion from. Meanwhile @Venatum keeps the most-pulled image current with every release, and also opened #1371 and #1063 when releases didn't make it to npm, so the person doing that work is already watching our release train from the outside.

Our docs have a smaller version of the same problem. The Compose example in the CLI guide runs `npx -y @bull-board/cli` inside `node:20-alpine`, so it re-resolves from npm on every container start, pins nothing, and doesn't work without egress to the registry. Air-gapped is exactly where a standalone dashboard earns its keep.

What I'd add is a `Dockerfile` at the repo root, a thin alpine Node layer over `@bull-board/cli` with the CLI as the entrypoint, so every flag and `BULL_BOARD_*` variable keeps working as documented. Then a `docker` job on the end of the release workflow: buildx for amd64 and arm64, pushed to `ghcr.io/felixmosh/bull-board`, tagged with the exact version, the major, and `latest`. The Compose example in the docs switches to the image and keeps the `npx` form underneath it as the no-image option.

None of this is a new npm package. The image has no configuration of its own, it only re-exposes the CLI's, so there's no version to keep in step with the other ten packages and no contract spec to write. I'd start on GHCR alone since it comes free with the repo. Docker Hub is where most people will look first, but it needs an account someone has to own, so I'd rather ask than assume.

Two things I'd like your call on before I start, @felixmosh: which registry, and whether you'd rather I talk to @Venatum first. An official image will pull traffic away from something they've maintained well for a long time, and I'd rather that didn't come as a surprise to them.

Happy to implement it if you're on board.

Stosiu:
Closes #1396.

A `node:22-alpine` layer over the published `@bull-board/cli`, with the CLI as the entrypoint, so flags, `BULL_BOARD_*` variables and a config file mounted at `/app/bull-board.config.js` all work as documented. The image itself configures nothing beyond the two defaults that make no sense in a container, `BULL_BOARD_HOST=0.0.0.0` and `BULL_BOARD_OPEN=false`, both still overridable. 63 MB, amd64 and arm64, runs as uid 1000, has a healthcheck.

The build is a reusable workflow called by the release and dispatchable by hand for any published CLI version. It resolves its own tags: the exact version always, the major only when that version is the newest in the major, `latest` only when it is npm's `latest`. So a hand-dispatched build of an old version cannot drag `9` or `latest` backwards. CI builds the image on every PR, so a broken Dockerfile fails here rather than at tag time.

Docs get a page of their own at `/guide/docker`, since "bull-board docker" is a thing people search for and a section inside the CLI guide would never rank for it. It covers the tags, Compose, mounting a config, running behind a reverse proxy, building the image yourself, and the old `npx` form for anyone who does not want to pull an image. The CLI guide keeps a short pointer to it.

Verified locally: discovery, basic auth, the mounted config, the healthcheck, SIGTERM, and both architectures.

The first push creates the GHCR package private, so it needs flipping to public once by hand.

I am not proposing a backfill of the earlier CLI versions. The CLI is ten days old, 613 downloads last week across seven versions, and no image tag has ever existed for any of them, so nothing is missing. The dispatch covers it in one run per version if you would rather the history be complete.

The heads-up to Venatum from the issue is still open, your call on before or after this merges.


felixmosh (approved):
Thank you @Stosiu

felixmosh:
Can you review
https://github.com/felixmosh/bull-board/actions/runs/33314027239/job/99264575696

probably related to the fact that NPM takes it's time to make a new published package public (sometimes it was 2 mins)

PR #1405, opened 2026-08-31
docs: refresh the landing page, correct four wrong statements, and reshoot the screenshots at v9.6.0

Stosiu:
A pass over the whole docs site: the landing page, four statements that were wrong, the gaps, and the screenshots.

Depends on #1404. Thirteen screenshots are reshot against `master` with it applied, because six of them carry the overview status row and three carry the queue info panel, and two paragraphs describe what it adds. Merging this first would ship images of UI that is not there yet.

## The landing page has a section nobody has ever seen

`pageType: home` discards the markdown body. The dashboard screenshot and the entire "Explore the docs" block at the bottom of `index.md` have never rendered, on any version of the site. I checked the published page rather than only a local build:

```
$ curl -s https://felixmosh.github.io/bull-board/ | grep -c "Explore the docs"
0
```

It is a component now, passed through the theme's existing `afterFeatures` slot, styled from the rspress variables so it follows both colour schemes. That is why this PR touches `website/theme/index.tsx`, which nothing else has needed to.

The rest of the page had drifted behind the README, which already opens with "Try it" and one command. The tagline still said "Plug it into your server" as though that were the only way in, and the six tiles predated the CLI, the Docker image, the schedulers view, `@bull-board/metrics` and BullMQ v6 on Postgres. The hero leads with both ways to run it now, the tiles describe what ships, and the body opens with the `npx` line. The introduction page gained a matching "Two ways to run it".

`guide/exploring-the-dashboard.md` had no sidebar entry. Two thousand words and fifteen screenshots, reachable only from three table cells and one bullet in a recipe. It is in the sidebar and linked from the landing page.

## Four things that were wrong

`recipes/change-polling-interval.md` documented `forceInterval` in milliseconds, `{ forceInterval: 2000 }, // ms`. It is seconds. `useSettings.ts` defaults it to 5 and every consumer multiplies by 1000, so following that example set a 33 minute poll and made the board look frozen. The same page then said not to go below 1 second, contradicting its own comment, and the UIConfig reference had it right the whole time.

`configuration/ui-config.md` typed the three `dateFormats` entries as `string` and called them date-fns format strings. They are `Intl.DateTimeFormatOptions` objects handed to `Intl.DateTimeFormat` in `formatDate.ts`. A format string does nothing. Corrected with an example.

`recipes/formatters.md` and `recipes/visibility-guard.md` both cited line numbers in `queueAdapters/base.ts`. Both had already drifted by 18 lines. Removed rather than corrected, since nothing keeps them true.

`recipes/global-concurrency.md` said to pick "Global concurrency" from the dropdown. The item is "Set concurrency"; "Global Concurrency" is the modal title. It also called the control a text input on the queue detail page when it is a dialog behind the three-dot menu.

## Gaps

`handlerHooks` from #1382 shipped with no documentation at all, so there is a new access control hooks recipe: the before and after contract, that a throw fails closed on a 500 rather than falling through, that `errorKey` has to come from the `ErrorTranslationKey` union, that the hooks wrap the JSON API and not the page or its assets, and where they sit relative to visibility guards and read-only mode.

Both adapter pages were missing `jobDataSchema` even though it is on the shared `QueueAdapterOptions` and the overview documents it at length. The read-only "what gets disabled" table predated several releases, so it now lists rate limits, reschedule and change priority, scheduler edit and remove, and remove-unprocessed-children; I checked each against `queueProvider` rather than assuming from the changelog. The Fastify page said nothing about the version lock, so someone on fastify@4 gets a `fastify-plugin` mismatch with no warning anywhere in the docs. And installation claimed Node 18+ across the board while the CLI and metrics packages both declare `engines.node >= 20`.

## Wording

Em-dashes are gone from prose in ten files. The ones left in the UIConfig table's Default column are "no default" markers, not prose.

`guide/cli.md` had a 350 word paragraph on `--browser` covering precedence, shell-free spawning, whitespace splitting and the macOS path-with-spaces failure in a single block. Same facts, five paragraphs. It also used a double hyphen as a dash in three places, and told the reader to pass `--no-retry` "to get the old behaviour back", which means nothing to someone reading the page for the first time.

Smaller ones: the `job.log()` caveat sat at the bottom of the Job flows section instead of Job logs, a heading in the dashboard tour was missing its blank line, the recipes index put the demo sentence above the sentence that introduces the page, and the Elysia page used `npm install` for a Bun runtime package.

## Screenshots

Thirteen reshot from the demo at v9.6.0, same dimensions and roughly the same file sizes. Six carry the status row and now show the All tab, three show the info panel, where `queue-info-modal` predated the rate limit row entirely and had never shown it. Four more were simply stale at 9.0.0. The old `grouped-overview` had also caught the sidebar logo mid-animation.

I have deliberately left the rest. The chart-heavy ones and the job-state ones draw on demo data that is generated relative to now, and it has drifted since #1360: my reshoot of `historical-metrics-page` came out as a flat line ending in a cliff, clearly worse than what is committed. Reproducing `job-will-fail`, `job-deduplicated`, `flow-ignored-children`, `add-job-schema` and the whitelabel pair means driving the demo into each state by hand, which I would rather do as its own PR than smuggle in here. They still read 9.0.0.

## Testing

`yarn docs:build` is clean with no dead links and the new recipe and the home component both render. I checked the rebuilt landing page in a browser in both colour schemes, and confirmed the asset and link URLs still carry the `/bull-board/` base. oxlint and `yarn format:check` are clean.

`yarn workspace @bull-board/website typecheck` reports one error on `rspress.config.ts`, a `defineConfig` overload complaint that reproduces on a clean master and is not something this PR touches.


felixmosh (approved):
Thank you @Stosiu

PR #1412, opened 2026-09-02
feat(ui): render job flows as a pannable graph and bound the flow read

### LINKED ISSUE #1406: Large Job Tree Visualization
Feature Description: 

I wrote a library that makes it really easy to make composable steps with queues. I occasionally have job graphs that are wide and deep.
The tree is really useful for 90% of my use cases, but with all the amazing improvements that have been made to the project, I thought it would be a good time to ask.

What if the job tree used A library like React Flow ? Out of the box that would give you zoom, pan, and I imagine with some work you could even make it so that when you click a card it loads the data above for that job. 

@Stosiu has implemented so many amazing features, I figured I would at least inquire for one more.

Stosiu:
Closes #1406.

The flow view was a nested `<ul>`. No pan, no zoom, and clicking a child navigated away and remounted the lot. It is a React Flow canvas now.

Job flow graph with the details panel beside it

## What the viewer does

- Pans and zooms. It opens fitted to the whole flow, with a floor on the zoom so a big graph shows its shape instead of a grey smear.
- Recentres on the job you came in from, via the crosshair under the zoom buttons. On a large flow that is the only sane way back.
- Goes fullscreen from the card header, Escape to leave. The canvas is not remounted, so your viewport and selection survive.
- Selecting a node fills the panel beside the canvas rather than navigating. Anything you expanded stays expanded.
- A node that is holding back children says so and loads them when asked.

## The read had to change too

`jobFlow.ts` passed neither `depth` nor `maxChildren` to `getFlow`, so it ran on BullMQ's defaults of ten levels and twenty children per dependency category. Whatever came back then got `getState()` and `getDependenciesCount()` called on every node, and the UI repeated the whole thing on the polling interval, per open tab.

I seeded a 9,331 node flow to see how bad it actually was:

| | nodes walked | Redis commands | time |
|---|---|---|---|
| before | 9,331 | 18,662 | 1,177 ms |
| after | 201 | ~402 | ~580 ms |

The endpoint now takes `depth`, `maxChildren` and `root`, and caps the response at 200 descendants whatever you ask for. The two window parameters default to what BullMQ itself uses, so a normal flow fetches exactly what it did before.

Most of the 580 ms that is left is `getFlow` itself, which I did not touch. Bounding that means walking level by level, and it can wait for a follow up.

A separate 2,001 node flow returns in about 130 ms with a 31 kB payload, and 201 nodes render at a median 8.4 ms frame.

Fan out gets its own treatment, since that is the shape @mwalkersigma described on the issue: one parent, one child per unit of work. A single level costs one hydration pass per child, so it is not subject to the flat cap, and one click on a parent with 600 children loads all 600 in about 80 ms. Past a thousand on one node the graph is the wrong tool, so the node says how many it is showing instead of offering a button that cannot deliver.

## The child counts were lying

`UNPROCESSED` was rendered as "N pending", but BullMQ files a child that has failed and may still retry under `unprocessed`, not `failed`. Its `getDependencyCounts` script reads four keys, and `unprocessed` is the parent's `dependencies` set, meaning everything it is still waiting on. So a permanently stuck flow would report "2 pending" while both children sat there failed. It says "unfinished" now, and selecting a failed node shows the reason and the attempt count, read from the job the panel already fetches.

## Two things that turned up

`getFlow` returns `undefined` entries in `children` for jobs that no longer exist, which is what a parent looks like after `removeOnComplete` has cleared its finished children. The old recursion swallowed those by accident and the rewrite did not, so the endpoint 500'd until I put the guard back.

Those ghost children still count in the parent's dependency totals, so a node would advertise children that expanding could never produce. They are counted separately now.

A flow parent with one unfinished and one ignored child, both of them failed

## Cost

`@xyflow/react` is the only new dependency and it sits behind `React.lazy` inside the `isFlowNode` check, so its chunks are fetched only when you open a job that is actually in a flow. Main bundle goes 49.0 kB to 49.4 kB. The layout is a plain tidy tree in about 25 lines rather than a second dependency.

Strings are translated across all twelve locales, and node text clears WCAG AA in both themes on every job state.


felixmosh (approved):
WOW! What an improvement, thank you @Stosiu!
This project is blooming with you around :]

PR #1413, opened 2026-09-02
feat(h3): support h3 2.x

Stosiu:
h3 publishes `2.0.1-rc.29` under its `latest` tag, and `@bull-board/h3` had never been run against 2.x. Two things break there, both fixed here:

- Routes were registered as `router.use(path, handler, method)`, which loses its params on 2.x, so every `:queueName` and `:jobId` route answered `ERRORS.QUEUE_NOT_FOUND` for queues that exist. They are now registered as `router[method](path, handler)`.
- 2.x does not infer `text/html` from the response body, so the dashboard was served as `text/plain` and browsers showed the raw source. The entry route now sets the content type explicitly.

Neither changes behaviour on 1.x. The declared range becomes `^1.15.11 || ^2.0.0`, and `yarn workspace @bull-board/h3 test` now runs the contract suite against both majors.

`@bull-board/test-utils` computes its fixture path in a CommonJS sibling now, so the kit also loads from the ESM jest project that h3 2.x requires.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -270,6 +270,35 @@ const base = require('./jest.base.js');
 module.exports = { ...base, displayName: 'express@4', moduleNameMapper: { '^express$': 'express-v4' } };
 ```
 
+### h3 version matrix
+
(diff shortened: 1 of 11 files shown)


felixmosh (approved):
Thank you @Stosiu 🙏🏼

PR #1427, opened 2026-09-03
feat(cli): historical metrics behind --history

### LINKED ISSUE #1414: Using `historyProvider` with cli/docker image
Hey there,
is it possible to configure the `historyProvider` [(ref)](https://felixmosh.github.io/bull-board/recipes/historical-metrics#register-the-provider) when using the docker image?
IIUC the docker image is just a wrapper around the cli and therefore does not contain the `@bull-board/metrics` package. Is this correct and if so are there any plans to enable historical metrics usage via the docker image?

Thanks!

Stosiu:
Closes #1414.

The Docker image is just the published `@bull-board/cli`, which didn't depend on `@bull-board/metrics` and had no seam to pass a `historyProvider` through, so historical metrics were unreachable from both the image and `npx @bull-board/cli`.

This adds the package as a CLI dependency (its only dependency is `@bull-board/api`, already in the tree, so the image grows by a few tens of KB) and one flag:

```sh
bull-board --redis redis://redis:6379 --history
```

That registers `RedisMetricsHistoryProvider` on the Redis client the CLI already holds, starts a `MetricsRecorder` in the same process, and turns on `uiConfig.showMetrics` so the range selector actually appears on queue pages. `--history-retention-days` sets the window; the rest (`record`, `retention`, `latency`, `snapshotIntervalMs`) goes in the config file under a `history` key. Both flags have the usual `BULL_BOARD_*` equivalents.

Recording is gated on `--read-only` rather than a second flag: with it, the provider is still registered and the charts still show whatever another process recorded. Several instances at once stay correct, since minute upserts are deltas and latency sampling takes a per-queue lease. The CLI warns at startup when no queue has anything in BullMQ's metrics buffer, and when it is serving history without recording it.

Two changes outside the CLI:

- `MetricsRecorder`'s `queues` now also accepts `() => BaseAdapter[]`, resolved per tick, because the CLI rediscovers queues every 10 seconds. Arrays behave exactly as before.
- `LatencySampler` cold-started its watermark at `now - tickMs` while ending each scan at `now - safetyMarginMs`, so any tick at or below the 5 second margin left the window empty on every tick and duration sampling never produced anything. It now starts at `now - tickMs - safetyMarginMs`. The 60 second default was never affected; the CLI's config file exposing `snapshotIntervalMs` is what surfaced it.

Tests, `packages/cli` 101 and `packages/metrics` 166 passing against a real Redis: config resolution, three runs of the real binary (with the flag, read-only, without), the per-tick queue resolution, and the sampler cold start. Docs updated in the CLI and Docker guides, the historical metrics recipe, and the three READMEs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,12 @@ BullMQ keeps only a short ring buffer of per-minute metrics, so the throughput c
 npm install @bull-board/metrics
 ```
 
+The CLI and the Docker image carry the package already, so `--history` turns the same thing on with nothing to install:
+
+```sh
+npx @bull-board/cli -r redis://localhost:6379 --history
+```
+
 See the [historical metrics recipe](https://felixmosh.github.io/bull-board/recipes/historical-metrics) for the recorder setup and storage sizing, or try it on the [live demo](https://felixmosh.github.io/bull-board/demo/).
 
 ## Packages
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -38,6 +38,9 @@ Options:
       --user <name>       Basic auth user (requires --password)
       --password <pass>   Basic auth password (requires --user)
       --board-title <s>   Dashboard title
+      --history           Record and serve long-retention metrics history
+      --history-retention-days <n>
+                          Days of history to keep            [90]
       --config <file>     Path to a config file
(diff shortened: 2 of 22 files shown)


felixmosh (approved):
LGTM

PR #1430, opened 2026-09-03
feat(ui): run a job scheduler on demand from the schedulers view

Stosiu:
The schedulers view can list a schedule, edit it and remove it, but there is no way to make one fire. Verifying a fix to a nightly job means waiting for the next cron window, or adding the job by hand from the add-job form and getting its name, data and options right from memory.

This adds a Run now action to each row. It reads the scheduler's stored template and adds one ordinary job from it, with the same name, the same data and the same options.

Schedulers view with the Run now action on every row

The schedule itself is not touched. Its pending delayed run stays where it was and still fires at its own time, and `iterationCount` does not move, because as far as BullMQ is concerned this is a job you added by hand rather than an iteration of the schedule. Nothing needs sanitising on the way in either: BullMQ stores a scheduler template as `Omit<JobsOptions, 'jobId' | 'repeat' | 'delay' | 'deduplication' | 'debounce'>`, so the stored options can go through as they are and the job lands as a plain waiting job.

Because this puts real work on a real queue, it asks first. It goes through the same confirm path the remove action uses, so the existing "Confirm queue actions" setting governs it rather than a second prompt policy of its own.

The confirmation before a scheduler runs

Read-only queues get a 405 from `queueProvider` as usual. Legacy Bull queues get a 405 with `ERRORS.JOB_SCHEDULER_RUN_NOT_SUPPORTED`, because Bull keeps no template for a repeatable job and there is nothing to build the run from. That is the same capability-flag shape `supportsJobSchedulerUpdate` already uses, so an adapter that cannot do this is never asked to.

The demo's `MockAdapter` gained the same method, and also gained `supportsJobSchedulerUpdate`, which it had never declared. The mock reports its type as `bullmq`, so the UI has always drawn the edit pencil on the live demo and clicking it has always come back 405.

Five cases in `job-schedulers.spec.ts` cover it: the template job with the schedule left intact, the pending run still delayed afterwards, 404 for an unknown scheduler, 405 on a read-only queue, and 405 on Bull. Removing the route and rebuilding `dist` fails exactly those five and nothing else.

`website/docs/public/screenshots/schedulers-page.png` is reshot at 9.7.0, and `guide/exploring-the-dashboard.md` describes the action alongside edit and remove.


felixmosh (approved):
👍🏼

dredav:
@felixmosh will 9.8.0 will also published as [docker-image](https://github.com/felixmosh/bull-board/pkgs/container/bull-board)?

felixmosh:
I've didn't saw that the Docker build failed.

re-run it, you have it now.
https://github.com/felixmosh/bull-board/pkgs/container/bull-board

PR #1431, opened 2026-09-03
feat(api): document the HTTP API with a generated OpenAPI spec

Stosiu:
bull-board ships 38 HTTP routes. The dashboard UI is a client of that API and nothing else, so anything the UI can do, a script can do. None of it is documented: to learn what exists you open `packages/api/src/routes.ts`, and to learn what any of it returns you open the handler and read to the bottom.

So people who want to automate around the board either don't, or they reach past it into Bull/BullMQ and rebuild what the board already does. This PR generates a spec from the route table so the API describes itself.

## What it adds

An interactive reference at `/api/`, a published `openapi.json` for generating clients or importing into Postman, and a plain-text version of the same reference that lands in `llms-full.txt`, the file the docs already tell people to feed their tooling.

API reference

Endpoint detail

## Why it can't go stale

Each route declares its response next to the route, and that same type parameter fixes the handler's body type, so a handler that stops returning what it advertises does not compile:

```ts
defineRoute({
  method: 'get',
  route: '/api/queues/:queueName/workers',
  spec: {
    summary: 'List the workers currently consuming one queue.',
    response: 'GetQueueWorkersResponse',
  },
  handler: queueWorkersHandler,
})
```

`yarn workspace @bull-board/api openapi` writes both artifacts, and CI regenerates them and fails on any difference.

That guarantee has already fired once. #1430 merged while this was open and added the on-demand scheduler route; on rebase the build broke, because the new route had not said what it returns. One `spec` block later it is documented like everything else.

## Behaviour changes

Typing the responses meant making `ControllerHandlerReturnType` generic, and that turned up five real bugs, all fixed here:

- `addJob` never awaited `job.getState()`, so `POST /api/queues/:queueName/add` has been answering `status: {}` instead of the job's state.
- `redisStats` returned a bare `'Forbidden'` string body when `hideRedisDetails` is set. Now the existing `ERRORS.FORBIDDEN` key.
- `pauseAll` and `resumeAll` returned English `{ message: 'All queues paused' }` bodies that nothing reads. Now `{}`.
- `GetJobResponse.status` was typed `Status`, but `getState()` can also return `stuck` and `unknown`. There is now a `JobState` type, narrowed at the UI boundary.
- `jobFlow` built one of its two responses with a possibly-undefined `nodeId`.

## Reviewing

Most of the diff is the two generated files. The hand-written parts are `packages/api/src/routes.ts`, `packages/api/scripts/openapi/`, and `scripts/openapi/api-overview.md`, which holds the prose a spec cannot express: that bull-board does not authenticate requests, that read-only mode, visibility guards and `handlerHooks` can refuse any call, and that error bodies carry translation keys.

Green on Node 20/22/24, Bun, the UI suite, lint, format and the docs build. Notes on how to run it locally, plus two pre-existing oddities I left alone, are in a comment below.


Stosiu:
Ready for review, and ready to check out and poke at.

## Seeing it locally

```bash
gh pr checkout 1431
yarn install
yarn build
yarn workspace @bull-board/demo build          # so the Demo link in the top bar resolves
yarn workspace @bull-board/website docs:dev
```

That serves on http://localhost:3000/bull-board/. Then `/api/` for the interactive reference and `/reference/http-api` for the plain-text one. `openapi.json` is served next to them. To regenerate after changing a route: `yarn workspace @bull-board/api openapi`.

## The six commits, in order

1. `refactor(api): give every route a typed response body`: the enabling change and most of the hand-written diff.
2. `feat(api): generate an OpenAPI spec and HTTP API reference from the route table`: the generator, the artifacts, the drift test.
3. `feat(website): browse the HTTP API through Scalar at /api/`
4. `feat(website): mirror the docs nav in the API reference top bar`
5. `feat(api): document the on-demand scheduler run route`: the #1430 route picked up on rebase.
6. `ci: fail when the generated OpenAPI artifacts are stale`

Reviewing 1 and 2 gets you the whole design; 3 and 4 are docs-site only.

## Two decisions worth your opinion

**`QueueMetrics.data` is typed `number[]` and the spec now repeats that.** BullMQ appears to return those values as strings from Redis without coercing them, which is why `packages/metrics/src/dataMapping.ts` wraps them in `Number()`. I did not change a published type on the side of this PR, but it does mean the spec states something that may be false. Happy to fix it here if you would rather it be right from the start.

**`yarn workspace @bull-board/website typecheck` fails** on `rspress.config.ts` with a `defineConfig` overload error. It is pre-existing on master, reproduces on the unmodified file, and is not wired into either workflow, so nothing here depends on it. Mentioning it because it is adjacent and easy to mistake for mine.

## Two implementation notes

`info.version` in the spec is a constant describing the HTTP contract, deliberately not the package version. If it tracked the package, every release bump would leave the committed artifacts stale and redden CI on release PRs.

The generator reads the route table from `dist/` rather than `src/`. It has to: `typings/app.d.ts` imports `BaseAdapter` from a root shim that re-exports `./dist`, while `src/` imports it from `src/`, so TypeScript treats them as two different classes. The package build only avoids this because `clean` deletes `dist` first, which makes that type resolve to `any` during compilation. Pre-existing and worth its own look someday; nothing here depends on changing it, and CI builds before it tests.

## Also in the box

The reference follows the site theme, and Scalar's sidebar will generate an MCP server straight from the spec, which answers the recurring "let my agent look at the queues" ask without this repo maintaining an MCP package.

Light mode

Plain text reference

packages/api/typings/requests.d.ts
```diff
@@ -0,0 +1,94 @@
+import {
+  JobStatus,
+  MetricsHistoryGranularity,
+  MetricsHistoryMetric,
+  MetricsLatencyGranularity,
+  MetricsLatencyMetric,
+  QueueRateLimit,
+} from './app';
+
+export interface GetQueuesQuery {
```
felixmosh, line 10: what keeps these types from getting wrong?

What do you think regarding adding Valibot (It consumes less memory than zod, I prefer it) schema validation of request + responses, which will help us the make sure that all types are derived from API input & output schemas.

No hand written types + we will gain input (request query, request params) + output (response types) validations.

What do you think regarding it?

felixmosh (changes requested):

packages/api/typings/requests.d.ts
Stosiu: Agreed on requests. They are decorative today: `req.query/params/body` are `Record<string, any>` and nothing binds `RequestSchemas` to a handler. Query params also cannot be typed honestly in TS alone (`page` says `number`, the wire gives a string), which `v.pipe(v.string(), v.transform(Number), ...)` fixes. I would do it as a follow-up though, since it is a behaviour change: malformed requests that pass today would start returning 400.

Responses I would leave. `defineRoute`'s type parameter already gates them at compile time, and `AppJob`'s fields are aliases of bullmq's `QueueJobJson`, so hand-written schemas would drift as bullmq moves. If you want proof the handlers actually match, `ajv` asserting real responses against `components.schemas` in the tests is cheaper. Say the word and I will add it here.

Pushed `c4812e15` meanwhile: `QueueMetrics.data` was genuinely wrong. bullmq 5.56, the peer floor, returns those points as strings. `BullMQAdapter` now coerces like `BullAdapter` already did.

felixmosh: Valibot schema will act as ajv 😅

This way all will be derived from it.

felixmosh (approved):
As you've suggested, we will add request validation layer, based on Valibot on a separate PR.

PR #1437, opened 2026-09-07
feat(cli): connect to a Redis Cluster

Stosiu:
Follows #1436, now merged, which made `@bull-board/metrics` work on a cluster. This does the same for the CLI and the Docker image, so the "cluster support is library-only" caveat in that PR goes away.

Rebased onto `master`, so the diff below is now this one commit and 17 files. #1438, which was stacked behind this branch, has been rebased onto `master` as well and no longer depends on it, so the two can be merged in either order.

## What changes

`--cluster` takes a comma-separated list of startup nodes, alongside `--redis` and `--sentinel` and mutually exclusive with both. `BULL_BOARD_CLUSTER_NODES` mirrors it. Node addresses are parsed by the same helper the sentinel list uses, including the IPv6 bracketing, with the port defaulting to 6379.

```sh
npx @bull-board/cli --cluster n1:7000,n2:7000,n3:7000 --prefix '{bull}'
```

Discovery scans every master rather than one, for the same reason the metrics admin had to: `SCAN` carries no key, so ioredis routes it to an arbitrary node.

`--redis-username` and `--redis-password` apply as they do in sentinel mode. `--redis-db` is rejected, since a cluster only has database 0 and passing it would otherwise be a silent no-op.

**Bull 3 queues are skipped, with a warning naming the queue.** Bull builds its keys without a hash tag and its Lua touches several at once, so on a cluster every command it issues fails with `CROSSSLOT`. Showing the queue and letting each action error is worse than not showing it. BullMQ queues on the same cluster are unaffected.

## Testing

Three cases in `packages/cli/tests/cli.spec.ts`, against the real three-master cluster #1436 added to `docker-compose.redis.yml`, skipped without `REDIS_CLUSTER_NODES`: a BullMQ queue discovered and served, `--history` recording into the cluster and reporting usage back, and a Bull 3 queue skipped rather than served. Plus config-level cases for parsing, credentials, IPv6, the rejected `--redis-db` and the mode conflicts.

Verified by hand as well: the CLI serves a BullMQ queue off the cluster, and `--history` writes `{bull-board:metrics}:...` keys and answers `/api/metrics/history/usage` with a footprint gathered from all three masters.

CLI suite: 142 pass with a cluster, 139 pass and 3 skip without one.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -86,7 +86,7 @@ jobs:
             || { echo "::error::OpenAPI artifacts are stale. Run 'yarn workspace @bull-board/api openapi' and commit the result."; exit 1; }
       # A cluster cannot be a `services:` entry: it needs three servers brought up and then
       # joined, which that block has no way to express. Without REDIS_CLUSTER_NODES the
-      # metrics cluster spec skips itself.
+      # metrics and CLI cluster specs skip themselves.
       - name: Start Redis Cluster
         run: |
           docker compose -f docker-compose.redis.yml up -d redis-cluster
--- a/packages/cli/src/config/connection.ts
+++ b/packages/cli/src/config/connection.ts
@@ -2,19 +2,26 @@ import type { RedisOptions } from 'ioredis';
 import type { FlagValues } from './flags';
 import type { FileConfig } from './types';
 
+export interface Address {
+  host: string;
+  port: number;
+}
+
 export type ConnectionConfig =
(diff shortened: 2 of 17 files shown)


felixmosh (approved):

felixmosh:
can you fix conflicts?

Stosiu:
Rebased onto master now that #1436 is in. This is a single commit and 17 files, no stack preamble needed. #1438 is off master too, so the two are independent and can go in either order.

PR #1438, opened 2026-09-07
fix(api): report the whole cluster in redis stats, not one arbitrary node

Stosiu:
Last of the Redis Cluster set, and the only one that affects the board itself rather than the metrics package or the CLI.

#1436 and #1437 are both merged, so the stack is gone and this is rebased onto current `master`: one commit, 6 files. The three-master cluster its tests need is already in `docker-compose.redis.yml` and already wired into CI.

`INFO` carries no key, so a cluster client has no slot to route it by and ioredis answers it from an arbitrary node. `BullMQAdapter.getRedisInfo()` feeds the board's Redis stats panel straight from that, so on a cluster the panel shows one node's memory, client count, uptime and port as if they were the whole cluster's, and switches to a different node's on the next poll. Eight consecutive calls against a three-master cluster came back from two different nodes:

```
getRedisInfo x8: answered by 2 distinct node(s): 7101/56372345 7102/957d1234
```

This is the same bug class as the `SCAN` one in #1436, and it predates that PR: it affects anyone pointing bull-board at a cluster-backed queue, with or without `@bull-board/metrics`.

## What changes

`getRedisInfo()` now merges `INFO` from every master when the client is a cluster. The additive fields (`used_memory`, `used_memory_peak`, `maxmemory`, `total_system_memory`, `connected_clients`, `blocked_clients`) are summed, uptime is the youngest node's, since the cluster is only as old as that, and everything the nodes agree on (version, mode, os) comes from the first. The result is still an `INFO` payload, so `redisStats.ts` and everything downstream keep parsing it unchanged, and `cluster_known_nodes` is set so the master count is available.

Standalone connections are untouched: they take the same `client.info()` path as before.

## Testing

`tests/queueAdapters/clusterInfo.spec.ts` covers the merge as a pure function, so it runs everywhere rather than only where a cluster exists: summing, the uptime minimum, agreed fields, a field no node reports, and an empty master list.

`tests/api/cluster.spec.ts` drives a real board over a real three-master cluster, skipped without `REDIS_CLUSTER_NODES` like the other cluster specs: serving a queue with its counts and jobs, adding, retrying and pausing through the API, and the stats panel reporting a total larger than any single node's and stable across repeated polls.

Reverting the merge turns that last case red (`Expected: > 3863280, Received: 3853496`, the arbitrary node's own memory) and leaves the other two green.

### DIFF
--- a/packages/api/src/queueAdapters/bullMQ.ts
+++ b/packages/api/src/queueAdapters/bullMQ.ts
@@ -21,6 +21,7 @@ import {
 import { DATASTORES } from '../constants/datastores';
 import { STATUSES } from '../constants/statuses';
 import { BaseAdapter } from './base';
+import { clusterInfo, isCluster } from './clusterInfo';
 
 /** The `:w:<name>` suffix BullMQ appends to the connection name of a named worker. */
 const WORKER_NAME_SEPARATOR = ':w:';
@@ -76,7 +77,11 @@ export class BullMQAdapter extends BaseAdapter {
 
   public async getRedisInfo(): Promise<string | null> {
     const client = await this.resolveRedisClient();
-    return client ? client.info() : null;
+    if (!client) {
+      return null;
+    }
+
+    return isCluster(client) ? clusterInfo(client) : client.info();
   }
 
   private async resolveRedisClient(): Promise<RedisClient | null> {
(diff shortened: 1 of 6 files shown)


felixmosh:
can you fix conflicts?

Stosiu:
Rebased straight onto master, off the stack. #1436 already put the three-master cluster in `docker-compose.redis.yml` and in CI, which is all this needed from the other two, so it is now a single commit and 5 files and does not wait on #1437.

felixmosh (approved):

PR #1443, opened 2026-09-09
feat(api): derive request and response types from valibot schemas and validate requests

Stosiu:
Follow-up to #1431, answering the thread on `typings/requests.d.ts` where you asked what keeps those types from getting wrong and proposed deriving everything from Valibot.

Everything derives from one schema now. `packages/api/src/schemas/` holds the shared payload shapes, each route's query and body, and each route's response. The TypeScript types are `v.InferOutput` of those schemas, the OpenAPI components are `@valibot/to-json-schema` of the same objects, and requests are parsed against them at runtime. `ts-json-schema-generator` is gone.

## Requests

Every documented query string and body is checked before its route runs, and anything that does not match is refused with 400. The check sits inside the hook wrapper and runs after `handlerHooks.before`, so a visibility guard still answers first and a malformed request cannot be used to find out that a hidden route exists.

Query parameters are typed honestly now: the wire carries `page=2`, the schema coerces it, the spec documents the string, and the handler receives `2`. The six handlers that hand-rolled numeric predicates no longer parse anything.

Existing 400s answer the keys they answered before, since a `key()` helper types a Valibot message as an `ErrorTranslationKey` and carries interpolation options through it. Only the routes that validated nothing before need new keys, `ERRORS.INVALID_QUERY_PARAM` and `ERRORS.INVALID_REQUEST_BODY`, translated across all eleven locales.

## Responses

The schema fixes the type, which fixes the handler, and generates the components in the spec as well. `defineRoute` declares `handler` as a property rather than a method so TypeScript checks it contravariantly, which is what makes a handler that reads an undeclared query fail the build.

Runtime response checking is in, behind `options.validateResponses`, off by default. Valibot's `v.object` drops unknown keys, so parsing a response would silently strip whatever an adapter or an `after` hook added, and a response carries forwarded job data where a false rejection turns a working dashboard into a 500. The production checklist says to use it while developing an adapter or a hook.

Switching it on found two published types that were lying. `AppJob.failedReason` was `string`, but the field is absent on any job that has not failed. `AppQueue.counts` was `Record<Status, number>`, claiming a `latest` count the queue never reports. Both optional now.

## Why the type declarations moved

`typings/app.d.ts` is `packages/api/src/types.ts` now, and the three published entry points are one-line re-exports of `dist/`. A `.d.ts` under `typings/` that imports from `../dist/...` resolves to `any` during a clean build rather than failing, because dist does not exist yet at that point. Deriving the types through that path would have turned `ResponseSchemas` into `any` inside our own build and taken #1431's compile-time gate with it, green the whole way. `tests/types/published-types-are-derived.ts` fails if a re-export path breaks. Published paths and export names are unchanged.

## Breaking changes

- Malformed requests that were accepted before now answer 400.
- A malformed body sent to a read-only queue answers 400 rather than 405, since validation runs ahead of the read-only check.
- `AppQueue.counts` entries and `AppJob.failedReason` are optional. Both were wrong before.
- `valibot` becomes a runtime dependency of `@bull-board/api`. It stays out of the browser bundle.

Requests are validated unconditionally. The response check stays behind the flag, which is what we settled on below.


felixmosh:
I'm down with you, let's put response validation behind a flag.

Stosiu:
That is what is in here. `options.validateResponses` is off by default, checks each body against the schema its route declares and answers 500 on a mismatch, and the production checklist points at it for developing a custom adapter, provider or `after` hook. Requests are validated unconditionally.

Switching the flag on during this work caught two published types that were wrong, both fixed here: `AppJob.failedReason` was `string` when the field is absent on any job that has not failed, and `AppQueue.counts` claimed a `latest` count the queue never reports.

Ready for review, CI green.

packages/api/src/schemas/requests.ts
```diff
@@ -0,0 +1,253 @@
...
+  metricsHistoryMetricSchema,
+  metricsLatencyGranularitySchema,
+  metricsLatencyMetricSchema,
+  statusSchema,
+} from './domain';
+import { key } from './support';
+
+const DAY_PATTERN = /^\d{4}-\d{2}-\d{2}$/;
+
+const wireNumber = v.union([v.string(), v.number()]);
```
felixmosh: it should be 
`v.pipe(v.string(), v.toNumber())`

packages/api/src/schemas/requests.ts
```diff
@@ -0,0 +1,253 @@
...
+
+function queryInteger(message: string, minimum: number) {
+  return v.pipe(
+    wireNumber,
+    v.transform(Number),
```
felixmosh: the `toNumber`, removes the need for transform.

packages/api/src/schemas/requests.ts
```diff
@@ -0,0 +1,253 @@
...
+    v.check((value) => Number.isInteger(value) && value >= minimum, message)
```
felixmosh:
```js
v.pipe(wireNumber, v.integer(), v.minValue(2))
```

packages/api/src/schemas/requests.ts
```diff
@@ -0,0 +1,253 @@
...
+      const parsed = typeof raw === 'string' || typeof raw === 'number' ? Number(raw) : Number.NaN;
+      if (raw === '' || raw === null || raw === undefined || !Number.isInteger(parsed)) {
+        return fallback;
+      }
+      return Math.min(Math.max(parsed, minimum), maximum);
+    })
+  );
+}
+
+export const getQueuesQuerySchema = v.object({
```
felixmosh, line 44: all fields are optional, so 

```js
v.partial(v.object({}))
```

packages/api/package.json
felixmosh, line 29: this should point to the new types, no?

felixmosh (changes requested):

packages/api/src/schemas/requests.ts
Stosiu: Done. `wireNumber` is gone from the query helpers, which are `v.pipe(v.string(msg), v.toNumber(msg), ...)` now. It also makes the spec honest: `page` and `jobsPerPage` document as `"type": "string"` instead of `anyOf[string, number]`, and openapi.json is 84 lines smaller.

The union survives in one place, `bodyNumber`, for the scheduler's `every` and `endDate`. Those come from a JSON body where a number really is a number, and the UI sends them that way.

packages/api/src/schemas/requests.ts
Stosiu: Right, `v.transform(Number)` is gone from both query helpers. `queryFinite` is a single line now:

```ts
v.pipe(v.string(message), v.toNumber(message), v.finite(message))
```

`toNumber` rejects NaN on its own, so the explicit emptiness and finiteness checks went with it.

packages/api/src/schemas/requests.ts
Stosiu: Done, with the message threaded through so the translation key survives:

```ts
v.pipe(v.string(message), v.toNumber(message), v.integer(message), v.minValue(minimum, message))
```

Same treatment elsewhere: the scheduler interval is `v.gtValue(0, ...)` rather than a hand-rolled `> 0`, and the only `v.check` left in the file is `endDate > Date.now()`, which has no built-in because it reads the clock.

packages/api/src/schemas/requests.ts
Stosiu: Applied where it works: `GetJobSchedulersQuery`, `ObliterateQueueBody`, `PurgeMetricsHistoryBody` and `UpdateJobSchedulerBody` are `v.partial(v.object({ ... }))` now, which drops five `v.optional(` wrappers from the scheduler body alone.

Not on this one, though. `page` and `jobsPerPage` are `v.optional(schema, default)`, and `v.partial` replaces each entry with a plain `optional()`, so the defaults go:

```
v.partial  =>  {}
v.optional =>  { page: 1, jobsPerPage: 10 }
```

`getPagination` needs both to be numbers, and having them in the schema is what puts `default` in the spec instead of leaving `+query.page || 1` in the handler. `GetJobFlowQuery` is the same case, its window defaults to 10 and 20.

packages/api/package.json
Stosiu: It already does, one hop further along. `typings/app.d.ts` is a single line now:

```ts
export * from '../dist/types';
```

so the map stays `"./typings/*": "./typings/*.d.ts"` and `@bull-board/api/typings/app` keeps resolving for every consumer, plus the UI and the adapters in this repo. Pointing the map at `dist/` directly would break the subpath shape, since `typings/app` would have to find `dist/app`, and the file is `dist/types.d.ts`.

The reason the declarations moved into `src/` at all: a `.d.ts` under `typings/` that imports from `../dist/...` resolves to `any` during a clean build rather than failing, because dist does not exist yet at that point. `tests/types/published-types-are-derived.ts` asserts the published aliases are not `any` so a broken path fails loudly.

felixmosh (approved):

PR #1445, opened 2026-09-10
chore: remove dead files, correct stale docs, and bring the examples to v9

Stosiu:
Housekeeping from a pass over the repo. Four independent commits, so any one can be dropped.

**Files that shouldn't be tracked.** An editor tool's cache in `packages/ui/src`, nine Yarn `install-state.gz` blobs under `examples/`, a pnpm lockfile in `packages/h3` that pins bull-board 5.17 and affects nothing, and two empty CSS modules. `.gitignore` gained rules so they don't come back.

**Docs that no longer match the code.** The Fastify build error that CLAUDE.md and CONTRIBUTING.md warn about is gone, adding an API error no longer touches `typings/app.d.ts`, the contract battery has 12 cases rather than 8, and the README's minimal Express example referenced three variables it never declared. Corrected those and a handful of similar ones.

**Examples.** Every project under `examples/` was on bull-board 5, 6 or 8; they now install 9.10.0 with lockfiles regenerated. Five were already broken before the bump and are fixed too: `with-nestjs-fastify-auth` never started at all, `with-nestjs` didn't compile and pointed at the wrong Redis port, `with-bun` had no `package.json`, and `with-hono` and `with-elysia` printed URLs that 404.

Worth a look on its own: `@bull-board/api@9.10.0` pulls in valibot, whose types need TypeScript 5. `with-nestjs` (4.7) and `with-nestjs-module` (4.3) stopped compiling with ~6000 errors out of `valibot/dist/index.d.mts` until I moved them to 5.9. Real consumers on TypeScript 4 will hit the same wall, so it may deserve a note in the release or the docs.

All twenty examples were started against a local Redis and checked to serve the dashboard, or to answer 302/401 where they sit behind auth. Both Next.js examples build and still pass `scripts/check-nextjs-trace.mjs`.


felixmosh (approved):

felixmosh:
Is it ready for review?

Stosiu:
@felixmosh yes, i just tested on bun as well, it's good to go

PR #1446, opened 2026-09-11
fix(api): name flow nodes as the board registered them

### LINKED ISSUE #1444: usage of `prefix` breaks job flow view
Hi there,
I stumbled across a subtle bug when using the bull-board `prefix` feature together with Flows.
It appears the link to jobs within the flow is not generated correctly. Give the following settings:

```txt
prefix: `Category.`
queueName: `my_queue`
```

expected: `queue/Category.my_queue/2fd73f00-2b9a-419c-8d17-a5289b7bbadc`

actual: `queue/my_queue/2fd73f00-2b9a-419c-8d17-a5289b7bbadc` (prefix is omitted)

---

This also results in the job beeing displayed as not available anymore despite it still being available

Image

Stosiu:
Fixes #1444.

A queue is registered on the board under `adapter.getName()`, which is the adapter's `prefix` followed by the queue name, and that is what every route param and every link in the UI is keyed by. BullMQ knows nothing about that prefix: a job reports `queueName` as the bare name, and `opts.parent.queue` as the qualified `redisPrefix:name`. The flow provider built its adapter lookup keyed by the board name and then matched BullMQ's names against it, so as soon as an adapter had a `prefix`, nothing matched.

Two failures followed. Walking a child job up to its flow root could not resolve the parent's adapter, so the endpoint answered `flowRoot: null` and the job looked like it had no flow at all. Where the tree did render, which is opening the root job itself since that needs no parent lookup, every node carried the bare queue name, so the details panel fetched `/api/queues/my_queue/<id>`, got a 404 and reported the job as gone, and "Open this job" pointed at a queue that does not exist on the board. `root=node` broke the other way round, passing the prefixed board name straight into `producer.getFlow()`, which finds nothing stored under that name.

The lookup is now keyed by the queue's qualified name, which is the exact form `opts.parent.queue` carries. That also settles the multi-tenant setup the NestJS docs describe, where one queue name runs under two Redis prefixes and two board prefixes, and the old bare-name match would have picked whichever adapter was registered first. `getFlowTree` takes a board name and translates it back to the raw name for BullMQ, `findFlowRoot` returns a board name, and every node in the response goes through `buildBoardQueueNameResolver`. The UI needed no change.

Both screenshots are the same board, registered as `new BullMQAdapter(queue, { prefix: 'Category.' })`, opened on the flow's root job with a child node selected.

Before, the nodes are labelled `my_queue` and `my_chunks`, and the selected child reports "This job is no longer in its queue":

Flow view before the fix

After, the nodes carry the names the board registered, and the child resolves:

Flow view after the fix

Three cases cover it in `tests/api/job-flow.spec.ts`: a prefixed board resolving the root from a child job and naming every node as the board does, `root=node` under a prefix, and two tenants sharing one queue name under different Redis prefixes. One more sits in `tests/bullmq-matrix/flow.spec.ts` so both BullMQ majors exercise it, and one in `tests/bullmq-matrix/postgres.spec.ts`, because a v6 PostgreSQL backend's qualified name is just the bare queue name rather than `prefix:name`. Each of them was confirmed failing against the current code before the fix landed. The full workspace suite passes, including the BullMQ v5 floor and v6 projects with `POSTGRES_URL` set.

### DIFF
--- a/packages/api/src/handlers/jobFlow.ts
+++ b/packages/api/src/handlers/jobFlow.ts
@@ -58,13 +58,13 @@ function countDependencies(dependencies: FlowNode['dependencies']): number {
   );
 }
 
-function toFlowNode(node: JobNode): FlowNode {
+function toFlowNode(node: JobNode, resolveQueueName: (job: Job) => string): FlowNode {
   return {
     id: node.job.id as string,
     name: node.job.name,
     progress: node.job.progress,
     state: 'unknown',
-    queueName: node.job.queueName,
+    queueName: resolveQueueName(node.job),
     children: [],
   };
 }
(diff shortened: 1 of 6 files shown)


felixmosh (approved):
LGTM

